PR #12, opened 2020-03-25
Improve checks for config and feed selection

### LINKED ISSUE #11: Abort on missing or wrong configuration
Currently, if no feed gets selected, the application still exits with return code 0 and the user doesn't see what's wrong. You will see in stdout that only the init action is visible, but no error is thrown. Especially for new users, this is confusing.
I see 2 situations where we should abort the execution and throw an error:

1. No configuration is loaded. Can happen if simply no configuration is present in the classpath or the given location or if there's a typo in the specified location.
1. A configuration is loaded, but the provided feed selector regexp doesn't select anything. This can happen if there's a general problem in the configuration or if there's a typo in the feed names.

pgruetter:
Abort if config could not be loaded, doesn't contain actions or dataObjects.
Abort if no feeds were selected.

Resolves #11

### DIFF
--- a/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
+++ b/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
@@ -21,6 +21,7 @@ package io.smartdatalake.app
 import java.io.File
 import java.time.LocalDateTime
 
+import com.typesafe.config.Config
 import configs.syntax._
 import io.smartdatalake.config.{ConfigLoader, ConfigParser, InstanceRegistry}
 import io.smartdatalake.util.hdfs.PartitionValues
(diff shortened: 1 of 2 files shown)


src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -168,15 +169,18 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
     logger.info(s"Deploy-Mode: ${appConfig.deployMode}")
 
     // load config
-    val config = appConfig.configuration match {
+    val config:Config = appConfig.configuration match {
```
danielspicar:
```suggestion
    val config: Config = appConfig.configuration match {
```

src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -168,15 +169,18 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
...
       case None => ConfigLoader.loadConfigFromClasspath
     }
+    require(config.hasPath("actions"), s"No configuration parsed or it does not have a section called actions")
+    require(config.hasPath("dataObjects"), s"No configuration parsed or it does not have a section called dataObjects")
     val globalConfig = config.get[Option[GlobalConfig]]("global").value.getOrElse(GlobalConfig())
 
     // parse config objects and search actions to execute by feedSel
     implicit val registry: InstanceRegistry = ConfigParser.parse(config)
     val actions = registry.getActions.filter(_.metadata.flatMap(_.feed).exists( _.matches(appConfig.feedSel)))
+    require(actions.nonEmpty, s"No action was selected with the given criteria (${appConfig.feedSel}). At least one action needs to be selected.")
```
danielspicar:
```suggestion
    require(actions.nonEmpty, s"No action matched the given feed selector: ${appConfig.feedSel}. At least one action needs to be selected.")
```

PR #15, opened 2020-03-25
Fix JDBC connections and dataObjects:

pgruetter:
- Load JDBC driver correctly with Class.forName
- Fix check for existing db and table as some databases (MySQL) do not necessarily store these in upper case


pgruetter:
> Oracle can have case sensitive table names... Can we put a configuration property into Environment class to enableJdbcCaseSensitivity?

Good idea, I added it in https://github.com/smart-data-lake/smart-data-lake/pull/15/commits/0c4a205d8c0210f3c6c1068b7b536e8a2acf306c. 
Default is false, so names are NOT case sensitive by default.

PR #18, opened 2020-03-29
allow multiple config files/directories as command line arg

zzeekk:
small non-breaking changes to command line:
- allow multiple config files/directories, separated by comma
- make name optional and use feed-sel as name if not given

### DIFF
--- a/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
+++ b/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
@@ -75,7 +75,7 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
 
   // read version from package manifest (not defined if project is executed in IntellJ)
   val appVersion: String = Option(getClass.getPackage.getImplementationVersion).getOrElse("develop")
-  val appType: String = getClass.getSimpleName.replaceAll("\\$$","") // remove $ from object name and use it as appName
+  val appType: String = getClass.getSimpleName.replaceAll("\\$$","") // remove $ from object name and use it as appType
 
   /**
    * Create a new SDL configuration and initialize it with environment variables if they are set.
--- a/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
+++ b/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
@@ -46,7 +46,7 @@ import scopt.OptionParser
  * @param keytabPath      Path to Kerberos keytab file for local mode.
  */
 case class SmartDataLakeBuilderConfig(feedSel: String = null,
-                                      applicationName: String = null,
+                                      applicationName: Option[String] = None,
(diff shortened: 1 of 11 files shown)


src/main/scala/io/smartdatalake/config/ConfigLoader.scala
```diff
@@ -148,12 +149,12 @@ object ConfigLoader extends SmartDataLakeLogger {
    *
    * This is an internal method to create a utility data structure.
    *
-   * @param rootPath      root [[Path]] pointing to a configuration file or a directory from where the traversal starts.
+   * @param rootPaths     root [[Path]]'s pointing to a configuration file or a directory from where the traversal starts.
```
pgruetter:
```suggestion
   * @param rootPaths     root [[Path]]s pointing to a configuration file or a directory from where the traversal starts.
```

src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -75,7 +75,7 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
 
   // read version from package manifest (not defined if project is executed in IntellJ)
   val appVersion: String = Option(getClass.getPackage.getImplementationVersion).getOrElse("develop")
-  val appName: String = getClass.getSimpleName.replaceAll("\\$$","") // remove $ from object name and use it as appName
+  val appType: String = getClass.getSimpleName.replaceAll("\\$$","") // remove $ from object name and use it as appName
```
pgruetter: Can you give the reasoning behind the variable name change appName to appType ? I see sometimes it's still appName.
Didn't check everything, but I guess a few comments / references like here would also need to be changed.

src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -167,10 +166,11 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
     logger.info(s"Application: ${appConfig.applicationName}")
     logger.info(s"Master: ${appConfig.master}")
     logger.info(s"Deploy-Mode: ${appConfig.deployMode}")
+    val appName = appConfig.applicationName.getOrElse(appConfig.feedSel)
```
pgruetter: Did you try with different regular expressions if everything still works? i.e when using ".*" as feed selector

src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
zzeekk: The applicationName is actionally given as parameter on the command line. It's confusing if we use it also here, so i was looking for a different name.
I'll check the other references.

zzeekk: No additional wrong references found for appName with text search.

PR #23, opened 2020-04-02
implement skip dag execution if there is no data to process

zzeekk:
if initExecutionMode finds no data to process, skip the following actions in the dag.

### DIFF
--- a/src/main/scala/io/smartdatalake/workflow/ActionDAG.scala
+++ b/src/main/scala/io/smartdatalake/workflow/ActionDAG.scala
@@ -61,7 +61,7 @@ private[smartdatalake] case class ActionDAGRun(dag: DAG[Action], runId: String, 
 
   private def createScheduler(parallelism: Int = 1) = Scheduler.fixedPool(s"dag-$runId", parallelism)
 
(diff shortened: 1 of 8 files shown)


src/main/scala/io/smartdatalake/workflow/ActionDAG.scala
```diff
@@ -72,9 +72,23 @@ private[smartdatalake] case class ActionDAGRun(dag: DAG[Action], runId: String,
       // wait for result
       val result = Await.result(futureResult, Duration.Inf)
       scheduler.shutdown
-      logger.info(s"""$op result: ${result.map(_.getClass.getSimpleName).mkString(",")}""")
 
-      // extract subfeeds, throwing exceptions if there are errors
+      // collect all root exceptions
+      val dagExceptions = result.filter(_.isFailure).map(_.failed.get).flatMap {
+        case ex: DAGException => ex.getDAGRootExceptions
+        case ex => throw ex // this should not happen
```
kaelte, line 79: It might be helpfull to write a logger message explaining that a DAGException is expected.

zzeekk: Well, all exceptions that can happen when executing an Action are catched and included in a DAGException. So if we have something other in my opinion we can just rethrow, SDLB will stop here and the exception is show in the console. I think a hard stop is good here and the exception with stacktrace in the console is even better than a log message.

PR #25, opened 2020-04-03
fix exception nbOfPartitions must be greater than 0

zzeekk:


src/main/scala/io/smartdatalake/workflow/action/CustomFileAction.scala
```diff
@@ -78,7 +78,7 @@ case class CustomFileAction(override val id: ActionObjectId,
     val tgtDO = output // avoid serialization of whole action by assigning output to local variable
     tgtDO.filesystem // init filesystem to prepare hadoop conf serialization
     val transformerVal = transformer // avoid serialization of whole action by assigning transformer to local variable
-    val nbOfPartitions = filePathPairs.size/filesPerPartition
+    val nbOfPartitions = math.max(filePathPairs.size/filesPerPartition,1)
```
kaelte, line 81:
```suggestion
    assert(0<filesPerPartition, s"filesPerPartition must be greater than 0. Current value: $filesPerPartition")
    val nbOfPartitions = math.max(filePathPairs.size/filesPerPartition,1)
```

zzeekk: Good hint, i added assertion to initialization

Issue #30, opened 2020-04-05
Control number of files created and sorting for exporting files


zzeekk:
HadoopFileDataObject creates as many files as spark partitions exists. When creating data exports this is not suitable. The number of files created (typically 1) and it's internal sorting (columns to sort a csv-file before writing) should be configurable.

PR #37, opened 2020-04-12
Config number of files created by SparkFileDataObject (#30)

zzeekk:


src/main/scala/io/smartdatalake/util/hdfs/SparkRepartitionDef.scala
```diff
@@ -0,0 +1,68 @@
...
+   * @return
+   */
+  def prepareDataFrame(df: DataFrame, partitions: Seq[String], nbOfPartitionValues: Int, dataObjectId: DataObjectId): DataFrame = {
+    // repartition spark DataFrame
+    val dfRepartitioned = if (partitions.nonEmpty && nbOfPartitionValues > 1) {
+      if (numberOfTasksPerPartition > 1 && keyCols.isEmpty ) logger.warn(s"($dataObjectId) SparkRepartitionDef: cannot distribute records over Spark tasks within Hadoop partitions with no keyCols defined. Define keyCols!")
+      // to distribute records across tasks within partitions, we need calculate a task number from keyCols
+      val taskNbCol = pmod(hash(keyCols.map(col):_*),lit(numberOfTasksPerPartition))
+      df.repartition(numberOfTasksPerPartition * nbOfPartitionValues, partitions.map(col) :+ taskNbCol :_*)
+    } else {
```
kaelte, line 59: I am not sure about the intended behaviour in case of partitions.Empty. As far as I understand prepareDataFrame is doing the following:

> When no partition columns are given, i.e. partitions.Empty, or the number fo PartitionsValues is less than 2 (negative values allowed) then the dataframe is repartitioned into numberOfTasksPerPartition many partitions ignoring partition columns but considering keyCols.

If this is correct I suggest to add the documentation.

src/main/scala/io/smartdatalake/workflow/dataobject/CsvFileDataObject.scala
```diff
@@ -82,9 +85,9 @@ case class CsvFileDataObject( override val id: DataObjectId,
   override val fileName: String = "*.csv*"
 
   private val formatOptionsDefault = Map(
-    "header" -> "false",
+    "header" -> "true",
     "inferSchema" -> "false",
-    "delimiter" -> "|",
+    "delimiter" -> ",",
```
kaelte, line 90: I rather prefer `|` as delimiter because it occurs less often in string fields.

zzeekk: hi @kaelte, but that would then be a Psv (pipe separated values) file... i suggest to use the pseudo standard for csv: delimiter = comma, header = true and override that in the config if we have other needs (like psv :-). Could you live with that?

src/main/scala/io/smartdatalake/util/hdfs/SparkRepartitionDef.scala
zzeekk: Hi @kaelte, thanks for the remark, i extended scala doc of SparkRepartitionDef,

src/main/scala/io/smartdatalake/workflow/dataobject/CsvFileDataObject.scala
kaelte: Yes I can live with comma as delimiter.
However a pipe delimited file is still a CSV: _Character-separated values_ .

PR #41, opened 2020-04-15
Custom Transformations by SQL

pgruetter:
This should work fine as a first version. We can still discuss putting SQL code in separate files but I don't think it needs to be quite as complex as with Scala code. Maybe it's enough to organize SQL code in separate configuration files.

### DIFF
--- a/src/main/scala/io/smartdatalake/workflow/action/Action.scala
+++ b/src/main/scala/io/smartdatalake/workflow/action/Action.scala
@@ -74,9 +74,8 @@ private[smartdatalake] trait Action extends SdlConfigObject with ParsableFromCon
 
     // Make sure that data object names are still unique when replacing special characters with underscore
     // Requirement from SQL transformations because temp view names can not contain special characters
-    val invalidTableNameCharacters = "[^a-zA-Z0-9_]".r
     val duplicateNames = context.instanceRegistry.getDataObjects.map {
-      dataObj => invalidTableNameCharacters.replaceAllIn(dataObj.id.id, "_")
+      dataObj => ActionHelper.replaceSpecialCharactersWithUnderscore(dataObj.id.id)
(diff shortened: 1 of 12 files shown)


pgruetter:
> Hi @pgruetter , reading Map[DataObjectId,String] works with an additional implicit converter.
> If you like it, you can pull branch SDL-112-Custom-transformations-by-SQL-zku.

Thanks a lot, liked and merged it ;-)

src/main/scala/io/smartdatalake/workflow/action/Action.scala
```diff
@@ -72,6 +71,16 @@ private[smartdatalake] trait Action extends SdlConfigObject with ParsableFromCon
   def prepare(implicit session: SparkSession, context: ActionPipelineContext): Unit = {
     inputs.foreach(_.prepare)
     outputs.foreach(_.prepare)
+
+    // Make sure that data object names are still unique when replacing special characters with underscore
+    // Requirement from SQL transformations because temp view names can not contain special characters
+    val invalidTableNameCharacters = "[^a-zA-Z0-9_]".r
+    val duplicateNames = context.instanceRegistry.getDataObjects.map {
+      dataObj => invalidTableNameCharacters.replaceAllIn(dataObj.id.id, "_")
```
kaelte, line 79: How can we ensure that the replacement which is tested is the same which will be used?
https://github.com/smart-data-lake/smart-data-lake/blob/98bbcc3fa7337ad880f31340f52f4122ae05fea1/src/main/scala/io/smartdatalake/workflow/action/customlogic/CustomDfTransformer.scala#L76-L77
I would prefer to have the replacement of invalid table name characters to be implemented once only.

pgruetter: Done in https://github.com/smart-data-lake/smart-data-lake/pull/41/commits/ebd8b215e2ec606ffd790f5f1b557f373b61d81f

PR #57, opened 2020-04-16
Replace mvn versions with sed command

### LINKED ISSUE #56: mvn versions:set fails
Hi @zzeekk 
The release pipeline failed because maven gets a NPE executing
`mvn versions:set -DnewVersion=1.0.0`

Found similar cases, but in our case it's because we have an expression in the artifactId:
`smartdatalake_${scala.minor.version}`

As a quickfix we could hard code the minor version 2.11 again, then the build process should continue. But we need to find a solution for this.

pgruetter:
Fix #56

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,7 +99,7 @@ jobs:
         current-version: ${{ env.MVN_RELEASE_VERSION }}
         version-fragment: bug
     - name: Develop Set Next Snapshot version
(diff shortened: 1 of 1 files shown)


zzeekk:
There is another mvn versions:set in line 102. It probably has the same problem, doesn't it?

pgruetter:
> There is another mvn versions:set in line 102. It probably has the same problem, doesn't it?

Well spotted, replaced it in https://github.com/smart-data-lake/smart-data-lake/commit/5b18d9e0e3d690344dca609cdac7f8767d602f80

PR #64, opened 2020-04-22
Prepare for maven central and fix actions

pgruetter:
- Generate javadoc jar when deploying
- Use personal access token (PAT) in order to activate branch restrictions again
- Add url and developer information in pom.xml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,7 @@ jobs:
       with:
         fetch-depth: 0
         ref: master
+        persist-credentials: false
     - name: JDK 8 Set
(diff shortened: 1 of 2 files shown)


pgruetter:
@zzeekk Looks like the build is a teeny bit too slow for this test case:
` 4.002 was not less than 4 Calculation took longer than 4 seconds, is this really executing in parallel? (DAGTest.scala:189)`
Do you see any way to make it more robust?
Maybe we can execute a parallel test and a serial test and measure the time for both, then see if parallel is significantly faster. Still not clean, but would probably handle slow executions better.

PR #80, opened 2020-05-10
implement configurable memory usage logging

zzeekk:
This is code i can contribute from another project to periodically log memory usage and trace shutdown cause.

I moved GlobalConfig in a separate file.

Starting a timer task on executors to log memory usage needs an ExecutorPlugin. The implementation is ready but not yet tested on a cluster.


zzeekk:
@pgruetter: can you review again last 3 commits? Thanks.

docs/YARN.md
```diff
@@ -0,0 +1,46 @@
...
+          feed = ab-csv
+        }
+      }
+    }
+    ```
+
+1.  Submit application to YARN cluster with spark-submit (make sure spark-submit version is 2.x and scala minor version is 2.11).
+    Dont forget to replace the SmartDataLake version (2x).
+    ```bash
+    spark-submit --master yarn --deploy-mode client --jars target/smartdatalake_2.11-1.0.3-jar-with-dependencies.jar --class io.smartdatalake.app.DefaultSmartDataLakeBuilder target/smartdatalake_2.11-1.0.3-jar-with-dependencies.jar --feed-sel ab-csv -c file://`pwd`/application.conf
```
pgruetter, line 45: Can we stick to the notation $(pwd) instead of `pwd` in scripts / descriptions? 
Also, this won't work in Windows, maybe make a note about that (replace pwd with the location of your application.conf)

docs/YARN.md
```diff
@@ -0,0 +1,46 @@
...
+        outputId = ab-csv2
+        metadata {
+          feed = ab-csv
+        }
+      }
+    }
+    ```
+
+1.  Submit application to YARN cluster with spark-submit (make sure spark-submit version is 2.x and scala minor version is 2.11).
+    Dont forget to replace the SmartDataLake version (2x).
```
pgruetter, line 43:
```suggestion
    Don't forget to replace the SmartDataLake version (2x).
```

pom.xml
```diff
@@ -271,7 +271,7 @@
 					<archive>
 						<manifest>
 							<addClasspath>true</addClasspath>
-							<mainClass>io.smartdatalake.app.DefaultSmartDataLakeBuilder</mainClass>
+							<mainClass>io.smartdatalake.app.LocalSmartDataLakeBuilder</mainClass>
```
pgruetter, line 274: Does the name "Default" SmartDataLakeBuilder still make sense if it's not the default anymore?

swissrussdev: If we're going to change it, now would seem to be a good time! What would you recommend? ClusterSDLB? I have no Mesos experience, but think we could use the 'Default' on any cluster manager. Would also fit as a contrast to Local. Interested in your thoughts.

swissrussdev: ...but doesn't cover standalone remote master. Default might not be so bad after all - Local is really the exception, it jist happens to be the usual entry point.

zzeekk: I also don't have a better name in mind for now. Let's leave it like that for the moment.

docs/YARN.md
zzeekk: Done

PR #81, opened 2020-05-13
fix starting with spark-submit

zzeekk:


swissrussdev:
Checks w. spark-submit on kerberized cluster ok

swissrussdev:
Spark submit to kerberized cluster also successful

PR #82, opened 2020-05-13
Feature/74 splunk token authentication

swissrussdev:
- Adds Token connection for Splunk
- Additionally (discussed with @zzeekk) introduces AuthMode as a means of consistently handling a single set of credentials for each connection/DO. Connection-specific subtraits enable a stepwise rollout of supported credential types for each connection type with early validation.

### DIFF
--- a/src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
+++ b/src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
@@ -18,6 +18,8 @@
  */
 package io.smartdatalake.definitions
 
+import io.smartdatalake.util.misc.CredentialsUtil
+
 /**
  * Authentication modes define how an application authenticates itself
  * to a given data object/connection
@@ -28,11 +30,23 @@ sealed trait AuthMode
 /**
  * Derive options for various connection types to connect by basic authentication
  */
-case class BasicAuthMode(userVariable: String, passwordVariable: String) extends AuthMode
+case class BasicAuthMode(userVariable: String, passwordVariable: String) extends AuthMode {
+  val user: String = CredentialsUtil.getCredentials(userVariable)
+  val password: String = CredentialsUtil.getCredentials(passwordVariable)
+}
 
 /**
  * Derive options for various connection types to connect by token
   */
-case class TokenAuthMode(tokenVariable: String) extends AuthMode
+case class TokenAuthMode(tokenVariable: String) extends AuthMode {
+  val token: String = CredentialsUtil.getCredentials(tokenVariable)
+}
 
+/**
+ * Validate by user and private/public key
+ * Private key is read from .ssh
+ */
+case class PublicKeyAuthMode(userVariable: String) extends AuthMode {
+  val user: String = CredentialsUtil.getCredentials(userVariable)
+}
 
--- a/src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
+++ b/src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
@@ -1,0 +1,31 @@
+/*
+ * Smart Data Lake - Build your data lake the smart way.
+ *
+ * Copyright © 2019-2020 ELCA Informatique SA (<https://www.elca.ch>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package io.smartdatalake.definitions
+
+/**
+ * Authentication modes define how an application authenticates itself
+ * to a given data object/connection
+ */
+sealed trait AuthenticationMode
+
+/**
+ *
+  */
+case class TokenAuthMode(token: String) extends AuthenticationMode
+case class UserPassAuthMode(user: String, password: String) extends AuthenticationMode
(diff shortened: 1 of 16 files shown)


swissrussdev:
Sorry - somehow missed the change to the config parser test

swissrussdev:
Comment on last commit incorrect  - apologies. Simply adapting the test to support BasicAuthMode.

src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
```diff
@@ -0,0 +1,77 @@
...
+sealed trait KafkaAuthMode extends AuthenticationMode { def kafkaOpts: Map[String, String]}
+sealed trait HttpAuthMode extends AuthenticationMode { def headerOpts: Map[String, String]}
+
+/**
+ * Derive options for various connection types to connect by basic authentication
+ */
+case class BasicAuthMode(user: String, password: String)
+  extends SplunkAuthMode
+{
+  val simpleOpts: Map[String, String] = Map(
```
zzeekk, line 60: I find creating a Map of the parameters quite complicated, given that in SplunkConnection you just take it appart again...
Couldn't we just create two variables with decrypted credentials and use them directly in Splunk Connection?
  val user: String = CredentialsUtil.getCredentials(userVariable)
  val password: String = CredentialsUtil.getCredentials(passwordVariable)
If different Connection classes use this parameters differently, this can be implemented in the corresponding Connection class. But if a Connections class needs other parameters, it's probably a different AuthMode, isn't it?

src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
```diff
@@ -0,0 +1,77 @@
...
+ * and prevent passing unsupported credentials into a connection/DO
+ */
+sealed trait SplunkAuthMode extends AuthenticationMode { def splunkOpts: Map[String, String]}
+sealed trait KafkaAuthMode extends AuthenticationMode { def kafkaOpts: Map[String, String]}
+sealed trait HttpAuthMode extends AuthenticationMode { def headerOpts: Map[String, String]}
+
+/**
+ * Derive options for various connection types to connect by basic authentication
+ */
+case class BasicAuthMode(user: String, password: String)
```
zzeekk, line 57: Convention for parameters to be "decrypted" is appending "Variable" -> userVariable, passwordVariable (see SftpFileRefConnection and JdbcTableConnection). Weather this is a good convention or not can be discussed, but we need some convention to mark parameters which will be passed through CredentialUtils...

src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
```diff
@@ -0,0 +1,77 @@
...
+    * @return : Map of decrypted credentials
+   */
+  def simpleOpts: Map[String, String]
+}
+
+/**
+ * The individual traits indicate which modes are supported by each connection type
+ * and prevent passing unsupported credentials into a connection/DO
+ */
+sealed trait SplunkAuthMode extends AuthenticationMode { def splunkOpts: Map[String, String]}
```
zzeekk, line 50: I like the marker traits! But they prevent modularization (putting SplunkDataObject in a different artifact because of complicated dependencies), because all sealed traits for a given base trait need to be located in the same scala file...
Maybe it would be better to do this without and validate if it's a supported AuthenticationMode in a require-statement in the beginning of SplunkConnection?

src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
swissrussdev: Correct  - cleaned up

src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
swissrussdev: Implemented - I can see the point about modularization, but I'm not completely happy with my implementation, feels a little clunky. However, don't see an alternative which doesn't binds the authmodes to the full set of Connections/Data Objects and thus prevent modularization.

src/main/scala/io/smartdatalake/definitions/AuthenticationMode.scala
swissrussdev: Agreed - done. I've moved the decryption back into the connection, but a generic "decrypt everything" shorthand method would be nice and save scattered calls to CredentialsUtil all over the place.

src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
```diff
@@ -32,24 +33,21 @@ import io.smartdatalake.util.misc.{CredentialsUtil, SmartDataLakeLogger}
  * @param id unique id of this connection
  * @param url jdbc connection url
  * @param driver class name of jdbc driver
- * @param userVariable variable used to get the user
- * @param passwordVariable variable used to get the password
+ * @param authMode optional authentication information: for now BasicAuthMode and PublicKeyAuthMode are supported.
```
pgruetter, line 36:
```suggestion
 * @param authMode optional authentication information: for now BasicAuthMode are supported.
```

src/main/scala/io/smartdatalake/workflow/dataobject/JmsDataObject.scala
```diff
@@ -37,18 +37,16 @@ import scala.concurrent.duration.Duration
  *
  * @param jndiContextFactory JNDI Context Factory
  * @param jndiProviderUrl JNDI Provider URL
- * @param userVariable Variable containing the JNDI user
- * @param passwordVariable Variable containing the JNDI password
+ * @param authMode authentication information: for now BasicAuthMode and PublicKeyAuthMode are supported.
```
pgruetter, line 40:
```suggestion
 * @param authMode authentication information: for now BasicAuthMode are supported.
```

src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
swissrussdev: Thanks - corrected and pushed

swissrussdev:
@zzeekk / @pgruetter : should I merge or do you want to?

zzeekk:
Hi @swissrussdev, normally the one who created the PR should merge when he got the approval. So i let that to you :-)

PR #89, opened 2020-05-20
fix checking for HADOOP_HOME environment variable on windows only

zzeekk:


src/main/scala/io/smartdatalake/app/LocalSmartDataLakeBuilder.scala
```diff
@@ -70,7 +71,7 @@ object LocalSmartDataLakeBuilder extends SmartDataLakeBuilder {
       case Some(config) =>
 
         // checking environment variables for local mode
-        require( System.getenv("HADOOP_HOME")!=null, "Env variable HADOOP_HOME needs to be set in local mode!" )
+        require( !EnvironmentUtil.isWindowsOS || System.getenv("HADOOP_HOME")!=null, "Env variable HADOOP_HOME needs to be set in local mode on Windows!" )
```
pgruetter, line 74:
```suggestion
        require( !EnvironmentUtil.isWindowsOS || System.getenv("HADOOP_HOME")!=null, "Env variable HADOOP_HOME needs to be set in local mode in Windows!" )
```

Issue #61, opened 2020-04-20
Implement recovery of failed runs


zzeekk:
As an Operator i want to recover failed SmartDataLakeBuilder runs by executing the missing actions after a potential problem is fixed.

Implementation hints:
- SmartDataLakeBuilder must write a summary file which includes the parameters for a specific run and the resulting state for each action of the dag
- SmartDataLakeBuilder needs a command line parameter to resume a previous run from it's summary file. It must be able to skip succeeded Actions in the DAG execution.

zzeekk:
The summary file must be written after execution of every action. The state of the action must include the subfeeds (partition values and fileRefs, but not DataFrame). On recovery the succeeded actions can be removed before building the DAG, but start actions of the remaining DAG must be initialized with the correct subfeed state.

PR #101, opened 2020-05-29
Feature 61 recovery

zzeekk:
# What changes are included in the pull request?
Implements Recovery and persisting execution state according to #61 
The SDLB command line has now two subcommands start & recover. You need to choose one of them.
In order to persist the execution state the new optional parameter statePath must be specified. It can be any hadoop path. After every action executed a file "<appName>_<runId>_<attemptId>.json" is written (overwritten) to this path. The same path can be shared between multiple applications.
The json files contains all parameters given to start SDLB, and for every action the resulting subfeeds (except the DataFrame) and main metrics.
To recover a failed run SDLB must be started with subcommand recover and parameters appName and statePath. The recovery run has the same runId but attemptId is increased by one. Succeeded actions are filtered before starting the execution.
If a run succeeds the next start of SDLB for this appName has runId increased by one.

# Why are the changes needed
see #61


pgruetter:
Great work! Actually a bit difficult to test because early validation is so good and catches most of the errors too early. A UDF worked.

Got a few things to discuss:

1. From an operations point of view, I think the distinction between "start" and "recover" is a bit unfortunate. Jobs are usually scheduled somewhere with set parameters. If a job fails (no matter what the reason), the problem gets fixed and I want to restart the same job again with the same parameters. I wouldn't want to start a recovery job manually by defining all the parameters for the recovery job.
I would therefore propose:
If a job is started with a new set of parameters, the job is started normally. If you have a failed job and start it again with the exact same parameters, it automatically tries to do a recovery. If you use the same parameters but the last job was successful, the whole job is started again.

1. The state information is never deleted. After a successful run, the state information should be removed. 
This would also be in line with what I've proposed in 1. --> If a run is successful (state data gets deleted) and you start it again with the same parameters, the whole pipeline is started.

1. If you do a recovery, only the failed actions are visible in the ASCII representation of the DAG. It would be very nice to see all actions again, but with a status of SKIPPED. Makes the recovery process much clearer.

1. Something I don't understand: To try a successful recovery, I fixed my class in IntelliJ and started the recovery process. Here, all the actions were started again, not only the failed one. Why is that? From my understanding, re-compiling everything should not change that behavior as long as the state files are still there.

zzeekk:
Hi @pgruetter, thanks for the review
1) i agree, we should remove the subcommands (start, recover). But i would fail if a run is started with different parameters (e.g. partition values...) and there is a failed run. I suggest that the user needs to cleanup the failed run (e.g. delete the state file) if he wants to start a run with different parameters.
2) the state information includes runtime statistics which could be used for reporting and visualization. I wouldn't delete it. But we could archive it, e.g. move it in a subfolder "succeeded" and have the current and failed runs in a subfolder "current"?
3) there is a log message before the ASCII-DAG listing the skipped actions. It's much harder to keep them in the DAG and feed them with the correct initialization subfeed. I would leave that as improvement for later.
4) i'll check myself

@swissrussdev: whats your 10cents on this PR?

pgruetter:
> 1. i agree, we should remove the subcommands (start, recover). But i would fail if a run is started with different parameters (e.g. partition values...) and there is a failed run. I suggest that the user needs to cleanup the failed run (e.g. delete the state file) if he wants to start a run with different parameters.

OK. Maybe it's warranted to create a command line option for this. As a special case you can just call Smart Data Lake Builder with --cleanup and all unfinished runs are discarded.

> 2. the state information includes runtime statistics which could be used for reporting and visualization. I wouldn't delete it. But we could archive it, e.g. move it in a subfolder "succeeded" and have the current and failed runs in a subfolder "current"?

OK. Since you need to provide the state info location anyway, it's up to the user to handle cleanup of this folder, agreed.

> 3. there is a log message before the ASCII-DAG listing the skipped actions. It's much harder to keep them in the DAG and feed them with the correct initialization subfeed. I would leave that as improvement for later.

👍

src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -55,11 +57,20 @@ case class SmartDataLakeBuilderConfig(feedSel: String = null,
...
                                       multiPartitionValues: Option[Seq[PartitionValues]] = None,
                                       parallelism: Int = 1,
+                                      statePath: Option[String] = None,
+                                      runId: Option[Int] = None,
                                       overrideJars: Option[Seq[String]] = None
                                 ) {
   def validate(): Unit = {
-    assert(!master.contains("yarn") || deployMode.nonEmpty, "spark deploy-mode must be set if spark master=yarn")
-    assert(partitionValues.isEmpty || multiPartitionValues.isEmpty, "partitionValues and multiPartitionValues cannot be defined at the same time")
+    assert(!applicationName.exists(_.contains("_")), s"Application name must not contain character '_' ($applicationName)")
```
ursin, line 65: Why are application names with `_` not allowed?

ursin:
@zzeekk: Again great work and a nice feature!

1. A dedicated `recover`-subcommand would make it more clear to users, that a job has to execute in a special mode and maybe skip some calculations. But I agree with @pgruetter: A single command will be easier for ops for recovering a failed job, especially when the jobs are triggered by other tools (airflow, azkaban, ...). This way, the ops don't have to mess around with the job parameters. Nevertheless, the states of the jobs have to be persisted somewhere.

2. I also prefer keeping the states somewhere for debugging purposes or logging. The disadvantage is, that sooner or later, some housekeeping has to be done. Will this be done by SDL with `--cleanup`?

3. +1 for state representation (skipped, failed, ...) in ASCII DAG, but this can be added later.

src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
zzeekk: We use "_" as delimiter in the state filename to separate appName_runId_attemptId.

zzeekk:
> 2. I also prefer keeping the states somewhere for debugging purposes or logging. The disadvantage is, that sooner or later, some housekeeping has to be done. Will this be done by SDL with `--cleanup`?

You're right, the functionality to keep succeeded runs in a separate subdirectory was missing. I added it now. For cleanup of succeeded subdirectory we could later add a global config like "stateMaxRunsToKeep".

PR #108, opened 2020-06-04
fix check if command line arguments have been parsed successfully

zzeekk:
Check if command line arguments have been parsed successfully in DatabricksSmartDataLakeBuilder and throw exception otherwise. It's already implemented the same in DefaultSmartDataLakeBuilder.

DatabricksSmartDataLakeBuilder doesn't stop when parsing command line arguments fail


pgruetter:
OK, was able to test it now after the merge. When specifying "-m yarn" it now correctly displays
Error: Unknown option -m
Error: Unknown argument 'yarn'
instead of throwing an error about override-jars.

Issue #105, opened 2020-06-04
Historization with neither whitelist or blacklist for columns: Unchanged row detection error


MarkusRothSBB:
**Describe the bug**
During historization, new rows are not mapped to existing rows properly when neither whitelist nor blacklist for columns are set.

**To Reproduce**
This unit test in HistorizationTest.scala
```
 test("When an existing historized record remains unchanged, history should not be updated") {
    val columns = List((123, "Egon", 23, "healthy"), (124, "Erna", 27, "healthy"))
    val dfOldHist = toHistorizedDf(session, columns, HistorizationPhase.NewlyAdded)
    println("History at beginning:")
    dfOldHist.collect.foreach(println)

    val dfNewFeed = toNewFeedDf(session, columns)
    println("New feed:")
    dfNewFeed.collect.foreach(println)

    val dfHistorized = Historization.getHistorized(dfOldHist, dfNewFeed, primaryKeyColumns, referenceTimestampNew, None, None)

    println("Historization result:")
    dfHistorized.collect.foreach(println)

    assert(TestUtil.isDataFrameDataEqual(sortResults(dfOldHist), sortResults(dfHistorized)))
  }
```
fails and outputs this:
```
History at beginning:
[123,Egon,23,healthy,2020-06-04 11:18:37.663,9999-12-31 00:00:00.0]
[124,Erna,27,healthy,2020-06-04 11:18:37.663,9999-12-31 00:00:00.0]
New feed:
[123,Egon,23,healthy]
[124,Erna,27,healthy]
Historization result:
[123,Egon,23,healthy,2020-06-04 11:18:37.663,9999-12-31 00:00:00.0]
[124,Erna,27,healthy,2020-06-04 11:18:37.663,9999-12-31 00:00:00.0]
[123,Egon,23,healthy,2020-06-04 11:18:37.663,2020-06-04 11:18:37.662]
[124,Erna,27,healthy,2020-06-04 11:18:37.663,2020-06-04 11:18:37.662]
```
**Expected behavior**
```
History at beginning:
[123,Egon,23,healthy,2020-06-04 11:11:31.601,9999-12-31 00:00:00.0]
[124,Erna,27,healthy,2020-06-04 11:11:31.601,9999-12-31 00:00:00.0]
New feed:
[123,Egon,23,healthy]
[124,Erna,27,healthy]
Historization result:
[123,Egon,23,healthy,2020-06-04 11:11:31.601,9999-12-31 00:00:00.0]
[...]

PR #109, opened 2020-06-04
Historization Tests reactivated, Bugfixing Hash calculation, test cle…

swissrussdev:
Historization Tests were inactive and there were errors in the handling of non-black/whitelisted dataframes and also in the calculation of hash values.

Link to issue: https://github.com/smart-data-lake/smart-data-lake/issues/105

Historisation results are incorrect without these changes


zzeekk:
Hi @swissrussdev, if my last commit is ok for you, let's merge and go for 1.0.4 :-)

swissrussdev:
Thanks - wanted to clean that up too, but ran out of time

Issue #110, opened 2020-06-05
Historisation fails when null values are interchanged


swissrussdev:
**Describe the bug**
The hashing mechanism operates on a string generated from the non-null Row values. A change in a record from "A", null, "B" to "A","B",null will therefore not be updated correctly.

**To Reproduce**
Unit test to be provided in PR.

**Expected behavior**
Changes in position of null values should cause a new historized record to be generated.

**Additional context**
Adding Schema Information to the Hash would make it more robust.

zzeekk:
Hello @swissrussdev, i also wondered about exluding null columns for hash calculation. In another project we use the following udf for calculating a hash for a Row:
`val udf_hash:UserDefinedFunction = udf((row: Row) => row.hashCode)`

swissrussdev:
Hi @zzeekk - thanks. Since the hash has to be calculated over a subset of columns, we don't always have a Row to work with. However, I've pushed a proposed solution, which incorporates the schema in the hash, allows for hashing on column subsets and also eliminates the need for a udf. Let me know what you think

PR #111, opened 2020-06-05
Correct historisation when null values swapped

swissrussdev:
* Historization mechanism based on JSON string
* Generalized function to hash a list of columns
* Additional Test for Null-Swaps
* Correction to isDataFrameData equal which was incorrectly failing identical dataframes

Link to Issue: #110 

See issue.

### DIFF
--- a/src/main/scala/io/smartdatalake/util/historization/Historization.scala
+++ b/src/main/scala/io/smartdatalake/util/historization/Historization.scala
@@ -38,10 +38,6 @@ private[smartdatalake] object Historization extends SmartDataLakeLogger {
 
   private val ts1: java.time.LocalDateTime => Column = t => lit(t.toString).cast(TimestampType)
 
-  private val fooSeed = MurmurHash3.stringHash("mySpecificSeed")
-  private val hashRow = (colsJson: String) => MurmurHash3.stringHash(colsJson, fooSeed)
-  private val hashFunc = udf(hashRow)
-
   def historizedForEmptyLoad(history: DataFrame, referenceTimestamp: LocalDateTime, doPersist: Boolean,
                              filterClause: Option[String])(implicit session: SparkSession): DataFrame = {
     session.sparkContext.setJobDescription(s"${getClass.getSimpleName}.historizedForEmptyLoad")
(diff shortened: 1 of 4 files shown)


src/main/scala/io/smartdatalake/util/historization/Historization.scala
```diff
@@ -145,21 +137,25 @@ private[smartdatalake] object Historization extends SmartDataLakeLogger {
       case (None, None) => newFeedDf.columns.sorted.toSeq
       case (Some(_), Some(_)) => throw new ConfigurationException("historizeWhitelist and historizeBlacklist mustn't be used at the same time.")
     }
-    val newFeedHashed = newFeedDf.withColumn("hash", hashFunc(struct(historizeHashCols.map(col): _*)))
+
+    // Generic column expression to generate a JSON string of the hash fields
+    def colHashExpr(cols: Seq[String]): Column = hash(to_json(struct(cols.map(col): _*)))
```
zzeekk, line 142: What is the benefit of using to_json hier? The names of the columns can't change inside a DataFrame. In my opinion this just eats performance... but maybe you have a good reason?
I suggest to used directly "hash(cols.map(col): _*)"

swissrussdev: Hi @zzeekk ,

See the example in the issue. The structure of the dataframe doesn't change, but which columns are null can do and therefore the solution using values only can miss version changes. Agreed we increase the payload for the hash function but I think that's a price worth paying for accuracy.

swissrussdev: ...maybe the hash is redundant?

swissrussdev:
Removed hashing and reverted to direct comparison of json strings. The logic is that time for hashing and string comparison are both proportional to string length but comparison can shortcut on first difference. Since we only compare rows once, there is no incremental benefit from reusing the hash code. I think the performance issues should be minimised with this approach. Direct struct comparison would have been nice, but doesn't seem possible using different data frames.

ursin:
Hi @swissrussdev and @zzeekk, I agree that the hashing of the json string is an overkill and the direct comparison of the strings should work fine. In the future such changes should be accompanied with some benchmark runs. Without that, we don't really know how big changes  impact performance.

PR #126, opened 2020-06-24
#19 Path Optional For Hive Tables

### LINKED ISSUE #19: Make path optional in HiveTableDataObject
Currently, the path is mandatory in HiveTableDataObject and TickTockHiveTableDataObject. 
This only makes sense when writing to a DataObject. 
When reading from it, you usually don't care where the underlying files are stored, you just access the external Hive table.

pgruetter:
Make path optional if you're only reading from a HiveTableDataObject or if you're writing to an existing table.

Closes #19


pgruetter:
High duplication is already documented as FIXME in TickTockHiveTableDataObject.

pgruetter:
> Sensible change in the code - what are all the pom.xml changes for? Can't see where geospark fits in here...

That was a mistake, accidentally pushed some local test settings. Have reverted them, thanks for catching it.

src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
```diff
@@ -216,6 +216,7 @@ private[smartdatalake] object HiveUtil extends SmartDataLakeLogger {
                     hiveDb: String, partitions: Seq[String], saveMode: SaveMode,
                     hdfsOutputType: OutputType = OutputType.Parquet, numInitialHdfsPartitions: Int = -1): Unit = {
     implicit val sss = session
+    require(!outputDir.equals(""), s"Can not write HiveTable ${hiveDb}.${hiveTable} because path was not set.")
```
zzeekk, line 219: Can it really happen that we pass an empty string? For me it's unusal to test for empty string, as that should normally be an Option[String] with value None...

src/main/scala/io/smartdatalake/workflow/dataobject/TickTockHiveTableDataObject.scala
```diff
@@ -55,7 +55,7 @@ case class TickTockHiveTableDataObject(override val id: DataObjectId,
   private val connection = connectionId.map(c => getConnection[HiveTableConnection](c))
 
   // prepare final path and table
-  @transient private[workflow] lazy val hadoopPath = HdfsUtil.prefixHadoopPath(path, connection.map(_.pathPrefix))
+  @transient private[workflow] lazy val hadoopPath = HdfsUtil.prefixHadoopPath(path.getOrElse("."), connection.map(_.pathPrefix))
```
zzeekk, line 58: It would be better if hadoopPath is an Option as well, because the default of "." is actionally not really what we want...

src/main/scala/io/smartdatalake/workflow/dataobject/HiveTableDataObject.scala
```diff
@@ -41,6 +41,8 @@ import scala.collection.JavaConverters._
  * @param table hive table to be written by this output
  * @param path hadoop directory for this table. If it doesn't contain scheme and authority, the connections pathPrefix is applied.
  *             If pathPrefix is not defined or doesn't define scheme and authority, default schema and authority is applied.
+ *             If DataObject is only used for reading or if the HiveTable already exist, the path can be omitted.
+ *             If the HiveTable already exists but with a different path, execution is stopped (unexpected behavior)
```
zzeekk, line 45: this is only a warning now, isn't it?

src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
pgruetter: You're right, removed the check completely. It didn't make much sense, the string can not be empty.

src/main/scala/io/smartdatalake/workflow/dataobject/TickTockHiveTableDataObject.scala
pgruetter: Please see new commit: https://github.com/smart-data-lake/smart-data-lake/pull/126/commits/c2cc545d39dd48874a514ddead4d15969613a4ca
Replaced hadoopPath with a placeholder as discussed.

src/main/scala/io/smartdatalake/workflow/dataobject/HiveTableDataObject.scala
pgruetter: Right, changed it in https://github.com/smart-data-lake/smart-data-lake/pull/126/commits/c2cc545d39dd48874a514ddead4d15969613a4ca

src/main/scala/io/smartdatalake/workflow/dataobject/HiveTableDataObject.scala
```diff
@@ -42,7 +42,7 @@ import scala.collection.JavaConverters._
  * @param path hadoop directory for this table. If it doesn't contain scheme and authority, the connections pathPrefix is applied.
  *             If pathPrefix is not defined or doesn't define scheme and authority, default schema and authority is applied.
  *             If DataObject is only used for reading or if the HiveTable already exist, the path can be omitted.
- *             If the HiveTable already exists but with a different path, execution is stopped (unexpected behavior)
+ *             If the HiveTable already exists but with a different path, a warning issued
```
zzeekk, line 45:
```suggestion
 *             If the HiveTable already exists but with a different path, a warning is issued
```

src/main/scala/io/smartdatalake/workflow/dataobject/TickTockHiveTableDataObject.scala
```diff
@@ -55,7 +55,40 @@ case class TickTockHiveTableDataObject(override val id: DataObjectId,
...
+    if (hadoopPathHolder == null) {
+      hadoopPathHolder = {
+        if (isTableExisting) new Path(HiveUtil.existingTableLocation(table))
+        else HdfsUtil.prefixHadoopPath(path.get, connection.map(_.pathPrefix))
+      }
+
+      // For existing tables, check to see if we write to the same directory. If not, issue a warning.
+      if(isTableExisting && path.isDefined) {
+        // Normalize both paths before comparing them (remove tick / tock folder and trailing slash)
+        val hadoopPathNormalized = hadoopPathHolder.toString
```
zzeekk, line 76: Could we create a function HdfsUtils.normalizePath to avoid this code dupliation?

pgruetter: That of course, makes a lot of sense: https://github.com/smart-data-lake/smart-data-lake/pull/126/commits/09f39d7ed596f34fd1ba6091043eeba698aa7055

pgruetter:
@zzeekk Want to take another look at https://github.com/smart-data-lake/smart-data-lake/pull/126/commits/09f39d7ed596f34fd1ba6091043eeba698aa7055 or can I merge?

zzeekk:
Nice, let's merge!

Issue #76, opened 2020-04-29
Allow dataframes from file sources to optionally include the filename


swissrussdev:
Often filenames contain additional metadata regarding the payload (for example snapshot date or region). It would be useful to optionally include this so that it can be parsed using custom logic. A considerable increase in parallelization could also be achieved. 

Implementation using:
https://spark.apache.org/docs/1.6.2/api/java/org/apache/spark/sql/functions.html#input_file_name()

zzeekk:
+1

swissrussdev:
Additional idea: regex pattern parameter to extract the partition name directly from the filename? Would eliminate custom code requirement in most instances.

zzeekk:
We already have something similar for SFtpFileRefDataObject. With partitionLayout you can extract parts of the path as partition columns, and replace them with columns values on writing.
For HadoopFileDataObjects this is not possible on writing, as filename is set by Spark. partitionLayout is set statically in HadoopFileDataObject, but is used only for directories.

So typically filename is parsed on ingestion with FileRefSubFeeds. Files are then written to HadoopFS and partition columns mapped as directories. How are your files ingested to HadoopFS?

To support this we could introduce an additional filenameLayout property on HadoopFileDataObjects, but if it is defined, these DataObjects are read-only for Spark.

swissrussdev:
I'll need to check that - I wasn't proposing anything quite that radical! My proposal was read only - I'll try to get some details posted soon.

swissrussdev:
Had a chance to reread this - 100% agreed. This has to be read only due to HadoopFS naming, but any other source (local FS, plus HDFS files created via HDFS put etc.) can be treated  exactly as you describe. One for 1.0.4.

PR #132, opened 2020-06-30
Optionally append filename as additional column to file DOs on reading

swissrussdev:
File Data Objects can now optionally include an additional column name which will contain the filename (full path)
on reading. 

Issue #76 
Enables Custom Transformers to easily parse filenames and create partitioning columns or derive additional logic if the filename has some semantic meaning.


swissrussdev:
Thanks @zzeekk - like the withOptionalColumn feature, didn't know that one!

Issue #44, opened 2020-04-15
Implement Delta Execution Mode


pgruetter:
ActionPipelines should allow delta mechanisms (only load new data). This would also greatly help recovery procedures.

swissrussdev:
Hi @pgruetter,
I reckon the work @zzeekk and I have started on structured streaming integration  could make this fairly straightforward. It would require a few custom microbatch readers to define "what is new data", but the underlying checkpointing mechanism is precisely for that purpose.

zzeekk:
Hi @swissrussdev , you're right, structured streaming execution mode with triggerOnce works well for all SparkFileDataObjects. We can also write to all the other DataObjects (Hive, Jdbc, ...). But we cannot read delta's from Hive or Jdbc, as there is no source for them (https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#input-sources).
We therefore plan for another execution mode which checks compares a "comparable" column in source and target DataObject for the delta, which then serves as parameter for spark batch.

swissrussdev:
Hi @zzeekk, I know there aren't existing DataSources for Jdbc/Hive, but it appears to be a case of implementing MicrobatchReader. Given that you get the whole checkpointing infrastructure for free, isn't that an alternative worth considering?

swissrussdev:
Following up on this : a big bonus with the structured streaming approach is that it works even when the target is heavily transformed - in particular if the reference column is aggregated (e.g. timestamp to date) or denormalized.

swissrussdev:
This is the sort of thing I mean.

https://github.com/sutugin/spark-streaming-jdbc-source

 Hive would be pretty much identical. Splunk/Web are trickier - but the direct comparison with target columns is if anything more difficult to implement.

zzeekk:
Nice example. I like the approach. It has at-least-once guaranteed, but not exactly-once, isn't it?

swissrussdev:
True - but I think that's an issue which we'll have anyway if we want to support arbitrary sinks using foreachbatch, multiple targets etc. Transactional targets e.g. JDBC/TickTock will be ok - needs a health warning for writing to non-tx targets, but we can support a LOT of use cases this way.

zzeekk:
Name to be discussed: SparkIncrementalMode or SparkDeltaMode? With SparkDeltaMode we have some conflicts with DeltaLake, e.g. DeltaLakeTableDataObject.

swissrussdev:
[...]

PR #143, opened 2020-07-07
Feature/44 incremental mode

zzeekk:
Implementation of #44 
New SparkIncrementalMode implemented, which allows an Action to select the difference between input and output by comparing a the max value of a given column.

SparkIncrementalMode is a simpler implementation of a delta load mechanism, compared to SparkStreamingOnceMode. It can be used with any DataObject which can create a DataFrame and has a column which is sortable and defines the ordering of records as they are inserted into source.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,21 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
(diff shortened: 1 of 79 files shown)


zzeekk:
Hi @MarkusRothSBB, can you please check this implementation if it fits your need?

MarkusRothSBB:
I am currently testing this branch for my use case, and have the following issue: During an initial load, the output hive table does not exist, leading to an AnalysisException when creating the DataFrame for it. I want to automatically create the hive table in this case.

I see two options:

1. Change HiveTableDataObject to create the hive table during prepare (similar to how JdbcTableDataObject works)

2. Add a createTable method to the TableDataObject trait, and do the following in applyExecutionMode: ` sparkOutput match {
              case table : TableDataObject => if (!table.isTableExisting) table.createTable(dfInput)
            }`

Which do you prefer? I think option 1 would lead to more consistent behavior, but maybe I will break something I can't see right now.

PR #145, opened 2020-07-08
various small fixes and improvments

zzeekk:
fixes
- avoid path not found exception by using dummy DataFrame if user defined schema exists
- fix using connection pathPrefix for FileSubFeedAction / HadoopFileDataObject
- fix deleting existing partitions for FileSubFeedAction
improvements
- add options to CustomFileTransfomer (as already existing for CustomDfTransformer)

### DIFF
--- a/src/test/scala/io/smartdatalake/config/TestFileTransformer.scala
+++ b/src/test/scala/io/smartdatalake/config/TestFileTransformer.scala
@@ -25,11 +25,12 @@ class TestFileTransformer extends CustomFileTransformer {
   /**
    * This method has to be implemented by the custom transformer to define the transformation.
    *
+   * @paran options
    * @param input  Hadoop Input Stream of the file to be read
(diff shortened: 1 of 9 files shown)


src/test/scala/io/smartdatalake/config/TestFileTransformer.scala
```diff
@@ -25,11 +25,12 @@ class TestFileTransformer extends CustomFileTransformer {
   /**
    * This method has to be implemented by the custom transformer to define the transformation.
    *
+   * @paran options
```
pgruetter, line 28:
```suggestion
   * @param options
```

pgruetter: Didn't fully understand what kind of options you can provide either. Maybe this is the correct place to add some comments about that?

PR #147, opened 2020-07-25
Feature/refactor executionmode needing main

zzeekk:
- configuration of mainInput/Output of CustomSparkAction is moved from execution modes to CustomSparkAction/SparkSubFeedsAction.
- refactoring of ExecutionMode code - it get's much simpler if it's implemented in methods of the corresponding ExecutionMode instead of method in ActionHelper.
- extension of PartitionDiffMode with an alternativeOutputId. It allows to build a partition diff with an arbitrary DataObject later in the DAG.

- Simpler, cleaner implementation.
- Selection of mainInput/Output is only needed in CustomSparkAction/SparkSubFeedsAction.
- alternativeOutputId can be used to ensure processing all partitions over multiple actions in case of errors.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -56,11 +56,11 @@ Data objects are structured in a hierarchy as many attributes are shared between
 Here is an overview of all data objects:
 ![data object hierarchy](images/dataobject_hierarchy.png)
 
-
 ## Actions
 For a list of all available actions, please consult the [API docs](site/scaladocs/io/smartdatalake/workflow/action/package.html) directly.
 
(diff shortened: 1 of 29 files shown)


zzeekk:
As this PR changes part of the existing interface (PartitionDiffMode.mainInput/OutputId) maybe we should increase version to 1.1.0-SNAPSHOT.

pgruetter:
> maybe we should increase version to 1.1.0-SNAPSHOT.

Yes, let's do that. Just change it directly before you merge. Action pipeline will release 1.1.0 automatically then.

PR #155, opened 2020-08-06
#153 Create empty dataframe automatically

### LINKED ISSUE #153: Create empty DataFrame according to schemaMin if table does not exist
When you use recursive inputs (see #150) you can build complex delta logics. On the first run, the output DataObject will not exist as the action has never ran before. In these cases, we would want to start with an empty DataObject that already has the correct schema. 

If a HiveTable does not exist yet but you provided a schemaMin in the configuration, an empty DataFrame should automatically be created for you.

**Additional context**
Extend getDataFrame in HiveTableDataObject

pgruetter:
If TickTockHiveTable does not exist when getting dataframe, use SchemaMin to create an empty DataFrame. 

Closes #153 

See #153


pgruetter:
@zzeekk We haven't implemented the usual methods on DeltaLakeTableDataObject like isTableExisting because delta lake table doesn't use the Hive Catalog. That's why I haven't implemented this feature for DeltaLake tables yet.

PR #160, opened 2020-08-11
Explode input and output ids...

pgruetter:
... so you get one line for each input and output dependency

Insert explode function in action exporter

In the exporter, CustomSparkActions with multiple inputs and / or multiple outputs are in the exported files only once. 
To build the lineage graph, we probably want one line per input/output dependency, not one line per action.


pgruetter:
@zzeekk Didn't test the output in visualization. Do you think this change makes sense? You already started with another lineage graph, do you think this would help?

pgruetter:
OK, fair enough. I also see your first version of the visualization in PR 39 doesn't have a problem with multiple input / outputs.

PR #164, opened 2020-08-12
Bugfix/fix check partitions existing

zzeekk:
Bugfixes:
- only check partitions existing in exec phase or for start SubFeeds of the DAG.
- fix clearDAGStart for FileSubFeedAction's
- validate FileSubFeeds partition values
- reduce code duplication

SDL run might fail on first run because partitions of intermediate results dont exist yet in init-Phase with Exception java.lang.AssertionError: assertion failed: (Action~z) partitions Vector(x=y) missing for DataObject~a


pgruetter:
> Great! Thanks for this fix.
> I stumbled on this bug just yesterday and thought there's a problem in my configuration. Tested it with this branch and it works correctly now.

Was too quick with that answer. Got furhter, but still get a similar error. Will analyze it and open another bug if necessary.

PR #174, opened 2020-08-17
Bugfix/173 fix unmatched exception delete data after read

### LINKED ISSUE #173: deleteDataAfterRead fails with unmatched case exception
**Describe the bug**
Setting deleteDataAfterRead=true causes an unmatched case exception.

e.g. `Unmatched case (CsvFileDataObject(DataObject~gugus-stg,/stage/gugus/raw,Map(quote -> ", header -> true, escape -> \, delimiter -> ;),List(),None,None,date,Overwrite,None,Some(AclDef(-rwxrwx---,Vector())),Some(Connection~hdfs),Some(sourcefile),None),InitSubFeed(DataObject~gugus-stg,List()))
`
**To Reproduce**
* Copy a csv file to a temporary directory
* Configure a CopyAction to copy the file to a target of your choice, set deleteDataAfterRead=true
* Run the Action

**Expected behavior**
Action should run und delete the input on success

**Additional context**
The Problem arises because CopyAction postExecSubFeed only works with SparkSubFeeds and not InitSubFeeds. I don't understand the messy details to know where the difference lies.

zzeekk:
fix #173

### DIFF
--- a/src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
+++ b/src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
@@ -26,6 +26,7 @@ import io.smartdatalake.util.evolution.SchemaEvolution
 import io.smartdatalake.util.hdfs.{HdfsUtil, PartitionLayout, PartitionValues}
 import io.smartdatalake.util.misc.SmartDataLakeLogger
 import io.smartdatalake.workflow.dataobject.Table
+import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.functions.{array, col}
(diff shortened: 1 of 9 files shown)


swissrussdev:
Thanks for the fix : unfortunately in my scenario this doesn't work as I'm working with non-partitioned input, so the case matches, but the condition doesn't. Can we extend the fix to deal with the following cases:
1) Non-Partitionable data source (not my case, but why not?)
2) CanHandlePartitions, but isn't partitioned (just extend the existing logic to deal with the case where partition values are empty)
?

zzeekk:
@swissrussdev: Yes we can implement all data if input is not partitioned or subfeed has not partition values (at least for FileRefDataObjects). Please check my suggestion.

Issue #83, opened 2020-05-13
Included spark / hadoop dependencies


zzeekk:
Spark / Hadoop dependencies have "compile" scope for now. This means these dependencies are normally included with their version (spark 2.4.4/hadoop 2.6.5) in projects which use the release build. This is good and simple for playing around, but in production most projects have specific spark/hadoop versions they want/need to use (e.g. hadoop 2.6.0-cdh5.16.1). Also when using spark-submit these dependencies are provided... if not excluded in the submitted jar they override the versions which are correct/needed on the cluster.

@pgruetter, @swissrussdev : What's you're opinion on this?

Here an example of the exclusions for spark-submit:
```
		<dependency>
			<groupId>io.smartdatalake</groupId>
			<artifactId>smartdatalake_${scala.minor.version}</artifactId>
			<version>1.0.3-SNAPSHOT</version>
			<scope>compile</scope>
			<exclusions>
				<exclusion>
					<groupId>org.apache.hadoop</groupId>
					<artifactId>hadoop-common_${scala.minor.version}</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.apache.spark</groupId>
					<artifactId>spark-core_${scala.minor.version}</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.apache.spark</groupId>
					<artifactId>spark-sql_${scala.minor.version}</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.apache.spark</groupId>
					<artifactId>spark-catalyst_${scala.minor.version}</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.apache.spark</groupId>
					<artifactId>spark-avro_${scala.minor.version}</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.scala-lang</groupId>
					<artifactId>scala-library</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.scala-lang</groupId>
					<artifactId>scala-reflect</artifactId>
				</exclusion>
				<exclusion>
					<groupId>org.scala-lang</groupId>
					<artifactId>scala-compiler</artifactId>
				</exclusion>
			</exclusions>
		</dependency>
```

swissrussdev:
Good catch - docs are fairly clear on this, they should be listed as provided. 

https://spark.apache.org/docs/latest/submitting-applications.html

zzeekk:
I discussed further maven best practices with Dominik. 

[...]

Issue #28, opened 2020-04-05
Modularisation / reduce dependencies


zzeekk:
In order to reduce unused dependencies we should think about modularisation of smartdatalake and separate specific components into their own artifacts. Candidates are:
- JMS data object
- Splunk data object
- Databricks App and Secrets
- Kafka data object (future)
- ...

PR #193, opened 2020-09-15
Refactor maven dependencies & modularize

zzeekk:
#83 refactor pom dependencies - copyied dependencyManagement and version properties from spark-parent
#28 modularize - create separate modules for core, kafka, splunk, jms and a parent module with dependencyManagement

See sdl-examples for an example pom using sdl-parent as parent and sdl-core as dependency:
https://github.com/smart-data-lake/sdl-examples/blob/feature/83-improve-pom/pom.xml

see issues

Note that we should merge this PR before developing further features, as merging must be done manually otherwise because of the files moved.


pgruetter (approved):
Thanks a lot. Tested sdl_core and it works correctly for me, also on Databricks @ Azure.
The resulting fat-jar is still a bit large (~140MB), maybe we could try to identify additional libraries to omit. I see com.ibm.icu is one of the larger ones, but it's needed by delta-core. 
Will also change one of the projects to use the new sdl_core to see how it integrates, but I don't see a reason not to merge already.

zzeekk:
Merged after discussing with swissrussdev and ursin that we can make optimization later.

PR #197, opened 2020-09-21
Update libraries due to security vulnerabilities

pgruetter:
Library updates:
com.google.guava:guava:14.0.1 -> 24.1.1-jre (CVE-2018-10237)
org.jpmml:pmml-model:1.2.15 -> 1.4.3 (WS-2019-0065)
org.apache.commons:commons-compress:1.4.1 -> 1.18 (CVE-2018-11771)

Github Dependabot reported 3 vulnerabilities in our dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -620,13 +620,13 @@
 			<dependency>
 				<groupId>com.google.guava</groupId>
 				<artifactId>guava</artifactId>
-				<version>14.0.1</version>
+				<version>24.1.1-jre</version>
 				<scope>${spark.deps.scope}</scope>
 			</dependency>
 			<dependency>
 				<groupId>org.jpmml</groupId>
 				<artifactId>pmml-model</artifactId>
-				<version>1.2.15</version>
+				<version>1.4.3</version>
 				<scope>provided</scope>
 				<exclusions>
 					<exclusion>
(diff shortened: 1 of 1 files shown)


zzeekk:
Hi @, i'm not sure that it is a good idea to update these dependencies without building a custom spark version, as we actually use the versions referenced in spark-parent. There is also issues about that: https://issues.apache.org/jira/plugins/servlet/mobile#issue/SPARK-25762
https://issues.apache.org/jira/plugins/servlet/mobile#issue/SPARK-23897

If we upgrade these versions we shoud at least run a spark build to test if it still compiles. Otherwise we might have classNotFound or MethodNotFound exceptions on using some functionality. If the build runs it should probably also work with zhe official releases.

@pgruetter : are you motivated to try a spark build? If im not wrong @raproth has experience with it...

pgruetter:
Hi @zzeekk 
Thanks a lot for your feedback. It looks like you're right, the impact is much larger than I thought. I didn't see that these libraries are only included now (after the pom refactoring) but are actually coming from the Spark dependencies. 
There are many dependencies and we won't be able to simply upgrade the libraries without impact. I propose to close the PR again without merging. I will open a issue so we don't forget about it and we will check again in the future how Spark handles them.

Issue #158, opened 2020-08-07
Case sensitive Jdbc schema names not supported


pgruetter:
**Describe the bug**
Tried with writing to postgresql: 
If the schema has a case sensitive name like "XyZ" you can't write into it properly.

**To Reproduce**
Create a connection:

```
postgresJdbc {
  type = JdbcTableConnection
  url = "jdbc:postgresql://host:port/postgres"
  authMode  = {
    type = BasicAuthMode
    userVariable = "CLEAR#username"
    passwordVariable = "CLEAR#password"
  }
  driver = "org.postgresql.Driver"
  db = XyZ
}
```
You get an error that the schema does not exist. 
If you escape it
`db = "\"XyZ\""`
the schema is found, but writing still doesn't work, you get an error that the table doesn't exist yet.

**Expected behavior**
Best solution would accept "db = XyZ" and handle the case correctly. 
If that doesn't work, at least the escaped sequence "\"XyZ\"" should be used everywhere so inserts work correctly.

pgruetter:
Maybe the same cause, having a schema name with hyphen and escaping it with \" will result in the same error:

```
ERROR ActionDAGRun$ActionEventListener: Action~some_table[CopyAction]: Prepare failed with AssertionError: assertion failed:
(DataObject~SOME_TABLE) Table "SCHEMA-WITH-HYPHEN".SOME_TABLE doesn't exist. Define createSQL to create table automatically.
```
It looks like only the Assertion is the problem (JDBCTableDataObject:154), not the actual interaction with the table.

PR #219, opened 2020-11-14
Use JdbcDialects' implementation of getTableExistsQuery

pgruetter:
Special characters, reserved words or case sensitive schema names in JDBC configurations could not be handled correctly. 
See #158 for details. Problems were seen in Postgres and MSSQL databases.

The following schemas don't work without this fix:

- Postgres schema called XyZ
- Postgres called schema-with-dash
- MSSQL schema called "AS" (reserved words)

### Additional information
In the fix, I ignore the Environment variable enableJdbcCaseSensitivity because I don't see an easy way to use JdbcDialects' implementation AND check for case sensitivity.
I would even argue, that we don't need the Environment variable. If a database like MSSQL is NOT case sensitive, it doesn't matter what you configure. For a schema called "AB", both "ab" and "AB" will work no matter what you use for enableJdbcCaseSensitivity.
In environment with case sensitivity like MySQL on Linux, I think the variable enableJdbcCaseSensitivity either doesn't work correctly or is very confusing. If you have two tables called "sample" and "Sample" and you insert data into "SAMPLE", where should your data go?

I propose to get rid of the variable and leave it to the JDBC driver and database if your config has to be case sensitive or not.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
@@ -26,6 +26,7 @@ import io.smartdatalake.config.{FromConfigFactory, InstanceRegistry}
 import io.smartdatalake.definitions.{AuthMode, BasicAuthMode, Environment}
 import io.smartdatalake.util.misc.SmartDataLakeLogger
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.jdbc.JdbcDialects
 
 /**
  * Connection information for jdbc tables.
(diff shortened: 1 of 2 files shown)


zzeekk (approved):
Thanks @pgruetter,

I made a commit to reuse catalog.isTableExisting function, as Option table.db is always defined, see assert statement in JdbcTableDataObject.

I also support to get rid of "enableJdbcCaseSensitivity" variable, but we would need to modify catalog queries for this.

pgruetter:
> I also support to get rid of "enableJdbcCaseSensitivity" variable, but we would need to modify catalog queries for this.

I opened #221 for this so I can merge this PR today.

PR #220, opened 2020-11-15
#215 and #182: multiple fail condition / move partition columns last

### LINKED ISSUE #182: Partition columns are only moved last during write
**Describe the bug**
There seems to be a discrepancy between the schemas of a DataFrame and the schema that is actually written to a HiveTable.

**To Reproduce**
Difficult to reproduce with a simple example. What I have is basically this:

Two actions between the 3 DataObjects:
`A --> <action1> --> B --> <action2> --> C`
If you execute action2 on it's own, it runs fine. If you run action1 and action2 together, then only action1 will run without error. action2 aborts with the error, that a schema evolution was detected. 
Reason seems to be that action1 puts the partiton columns at the end of the HiveTable **when writing the data**. action2 gets the schema from action1 and doesn't expect the partition columns at the end.

**Expected behavior**
If partition columns are moved to the end of the column list, it should be done uniformely in the DataFrame and in the written data.

**Additional context**
We should move the partition columns to the end in the transformation already, not only in writeDfToHive.

### LINKED ISSUE #215: Extend execution mode fail condition to allow multiple conditions with description
One can define a failCondition for Actions execution mode to fail depending on given execution context. For complex expressions it is hard to comprehend why a fail condition triggered.

Allow to configure multiple fail conditions with separate descriptions which are logged in error message.

zzeekk:
fixes #215
fixes #182 

Additional improvements:
- SparkIncrementalMode: throw NoDataToProcess if input DataObject is empty
- Better evaluation of spark expressions in configuration properties through updated spark-extension artifact
- additional unit tests for ExecutionMode
- reduced warn logs in unit tests


sdl-core/src/main/scala/io/smartdatalake/definitions/Condition.scala
```diff
@@ -0,0 +1,43 @@
...
+package io.smartdatalake.definitions
+
+import io.smartdatalake.config.ConfigurationException
+import io.smartdatalake.config.SdlConfigObject.ConfigObjectId
+import io.smartdatalake.util.misc.SparkExpressionUtil
+
+import scala.reflect.runtime.universe.TypeTag
+
+/**
+ * Definition of a condition.
```
pgruetter, line 29: Please expand description a bit. Condition of what? Used when?

sdl-core/src/main/scala/io/smartdatalake/util/misc/DataFrameUtil.scala
```diff
@@ -239,8 +238,9 @@ private[smartdatalake] object DataFrameUtil {
     def isEqual(df2: DataFrame): Boolean = {
       // As a set-theoretic function symmetricDifference ignores multiple occurences of the same row.
       // Thus we need also to compare the cardinalities and the schemata of the two data frames.
+      // For the schema the order of columns doesnt need to much.
```
pgruetter, line 241:
```suggestion
      // For the schema, the order of columns doesn't need to match.
```

pgruetter (approved):
Thanks, looks good. Just two minor comments in Scaladoc. 
And one general remark: if possible, please submit multiple PRs for that many different changes as it makes the reviews a bit confusing otherwise.

sdl-core/src/main/scala/io/smartdatalake/util/misc/DataFrameUtil.scala
zzeekk: wow, thats too match in one sentence ;-)

zzeekk:
> And one general remark: if possible, please submit multiple PRs for that many different changes as it makes the reviews a bit confusing otherwise.

You're right, i could have at least merge the changes of develop in a seperate PR.

PR #222, opened 2020-11-16
Fix schema migration with DeduplicateAction

MarkusRothSBB:
-Refactor DeduplicateAction for better testability
-Fix bug regarding schema migration and DeduplicateAction
-Introduce new test helpers for easier testing

I am still working on the new test helpers as discussed. In the end, I think they should not live in test but in main, so that applications using SDL can use the test helpers when writing tests for their applications.
However, I added them here because I needed to write tests to fix a bug in the DeduplicateAction regarding schema evolution, and I think they allow for much easier to follow tests than the ones present.

I need the bugfix soon, so if we want to postpone adding the testhelpers until they are more refined, that's fine with me. In that case, I'll modify the pull request to just include the bugfix with no tests, and we will re-add them when the test helpers are complete.

### DIFF
--- a/src/main/scala/io/smartdatalake/workflow/action/DeduplicateAction.scala
+++ b/src/main/scala/io/smartdatalake/workflow/action/DeduplicateAction.scala
@@ -126,8 +126,20 @@ object DeduplicateAction extends FromConfigFactory[Action] {
     if (existingDf.isDefined) {
       // apply schema evolution
       val (baseDf, newDf) = SchemaEvolution.process(existingDf.get, enhancedDf, ignoreOldDeletedColumns = ignoreOldDeletedColumns, ignoreOldDeletedNestedColumns = ignoreOldDeletedNestedColumns)
-      deduplicate(baseDf, newDf, pks)
+      // Schema evolution puts new columns at the end of the column list, so we need to move the technical captured column back to the end
+      deduplicate(moveCapturedColumnToEnd(baseDf), moveCapturedColumnToEnd(newDf), pks)
     } else enhancedDf
+  }
+
+  /**
+   * The techincal "captured" column needs to be the last column in the DataFrame so that the union in the deduplicate method words
+   * In Spark, unions do not work by column name but by column index, so the column order of the dataframes to be unioned needs to be equal
+   * @param df A DataFrame
+   * @return A DataFrame where the techinical "captured" column is the last column
+   */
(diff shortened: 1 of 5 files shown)


src/test/scala/io/smartdatalake/testutils/DataFrameTestHelper.scala
```diff
@@ -0,0 +1,274 @@
...
+  implicit def valueToTypedValue[T](value: T): TypedValue = value match {
+    case (value: Any, dataType: DataType) => TypedValue(value, dataType)
+    case (null, dataType: DataType) => TypedValue(null, dataType)
+    case string: String => TypedValue(string, StringType)
+    case int: Int => TypedValue(int, IntegerType)
+    case bool: Boolean => TypedValue(bool, BooleanType)
+    case _ => throw new Exception("Unable to convert to TypedValue")
+  }
+
+  def assertSchemasEqual(expected: StructType, actual: StructType): Unit = {
```
zzeekk, line 170: In assertDataFramesEqual we have another implementation of schema comparision. Could we merge them?
I would prefere the one from assertDataFramesEqual.

src/main/scala/io/smartdatalake/workflow/action/DeduplicateAction.scala
```diff
@@ -93,26 +93,55 @@ case class DeduplicateAction(override val id: ActionObjectId,
...
     } else enhancedDf
   }
 
+  /**
+   * The techincal "captured" column needs to be the last column in the DataFrame so that the union in the deduplicate method words
+   * In Spark, unions do not work by column name but by column index, so the column order of the dataframes to be unioned needs to be equal
+   * @param df A DataFrame
+   * @return A DataFrame where the techinical "captured" column is the last column
+   */
+  def moveCapturedColumnToEnd(df: DataFrame) : DataFrame = {
```
zzeekk, line 140: I also came accross this problem when working on #182. There is a better solution by using df.unionByName in deduplicate(...). Shouldn't we go directly for that?

zzeekk (commented):
Thanks @MarkusRothSBB,

I have made some comments, and we would also need to consolidate DataFrameTestHelper and TestUtils. But I suggest to do this later in another PR.

MarkusRothSBB:
I can't push updates to this PR because I chose the wrong branch name. Closing this to re-open with a better branch name.

PR #223, opened 2020-11-17
fix release action

zzeekk:
set-env got deprecated, see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


pgruetter (approved):
Looks like a lot of work, thanks. Hope that makes the builds stable for a while again.

PR #225, opened 2020-11-18
DataFrameTestHelpers and unionByName in DeduplicateAction

MarkusRothSBB:
**What changes are included in the pull request?**
-Refactor DeduplicateAction for better testability
-Fix bug regarding schema migration and DeduplicateAction
-Introduce new test helpers for easier testing

**Why are the changes needed?**
I am still working on the new test helpers as discussed. In the end, I think they should not live in test but in main, so that applications using SDL can use the test helpers when writing tests for their applications.
However, I added them here because I needed to write tests to fix a bug in the DeduplicateAction regarding schema evolution, and I think they allow for much easier to follow tests than the ones present.

I need the bugfix soon, so if we want to postpone adding the testhelpers until they are more refined, that's fine with me. In that case, I'll modify the pull request to just include the bugfix with no tests, and we will re-add them when the test helpers are complete.


MarkusRothSBB:
This PR is a re-issue of https://github.com/smart-data-lake/smart-data-lake/pull/222 with a better branch name. I changed the bugfix to use .unionByName, which is much better than my solution. I also merged the two schema comparison methods into one and changed the DataFrameTestHelper to take an implicit parameter with the SparkSession.

zzeekk (approved):
Lgtm

PR #230, opened 2020-11-20
Replace maven-assembly-plugin with maven-shade-plugin

MarkusRothSBB:
In the POM, replace the maven-assembly-plugin with the maven-shade-plugin which features the option of appending resource files rather than replacing them in the resulting fat jar

If I create an SDL project and set the parent POM to the POM of sdl-core, I can use the fat jar build profile to build a fat jar containing my application as well as sdl and all dependencies.
However, the application does not work due to this issue: https://stackoverflow.com/questions/17265002/hadoop-no-filesystem-for-scheme-file/27532248#27532248
By using the shade-plugin, the resource files can be merged, and the fat jar works

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -400,29 +400,42 @@
 				<!-- Builds executable fat-jar that includes all dependencies -->
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
-					<artifactId>maven-assembly-plugin</artifactId>
-					<version>3.1.1</version>
-					<configuration>
-						<descriptorRefs>
-							<descriptorRef>jar-with-dependencies</descriptorRef>
-						</descriptorRefs>
-						<archive>
-							<manifest>
-								<mainClass>io.smartdatalake.app.LocalSmartDataLakeBuilder</mainClass>
-							</manifest>
-							<manifestEntries>
-								<Implementation-Version>${project.version}</Implementation-Version>
-							</manifestEntries>
-						</archive>
-						<skipAssembly>${skip.assembly}</skipAssembly>
-					</configuration>
+					<artifactId>maven-shade-plugin</artifactId>
+					<version>1.5</version>
 					<executions>
 						<execution>
-							<id>make-assembly</id>
 							<phase>package</phase>
 							<goals>
-								<goal>single</goal>
+								<goal>shade</goal>
 							</goals>
+							<configuration>
+								<filters>
+									<filter>
+										<artifact>*:*</artifact>
+										<excludes>
+											<exclude>META-INF/*.SF</exclude>
+											<exclude>META-INF/*.DSA</exclude>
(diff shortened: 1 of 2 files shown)


zzeekk:
Hi @MarkusRothSBB, do you use -Pfat-jar or -Pfat-jar-with-spark?

I read something similar here: https://docs.cloudera.com/HDPDocuments/HDP2/HDP-2.3.2/bk_storm-user-guide/content/ch_storm-pkg-topologies.html
It seems mainly to affect hadoop, but actionally hadoop should not be included in fat-jar, only in fat-jar-with-spark...
The bug is that it's not possible to override properties from a profile in the parent pom. This means that scala.deps.scope and spark.deps.scope are not set to "provided" as this profile is defined in the parent pom...
Interestingly this seemed to work for scala-2.11/2.12...

@pgruetter: Seems we need to move profile definition to the modules and sdl-example. And what do you think about maven-assembly vs maven-shade plugin?

pgruetter:
> And what do you think about maven-assembly vs maven-shade plugin?

Really don't know much about the two plugins. But quickly reading up on it (i.e. [here](https://medium.com/@randilfernando/when-to-use-maven-jar-maven-assembly-or-maven-shade-ffc3f76ba7a6), using shade seems like a plausible suggestion.

zzeekk:
I improved some things in the application of shade plugin. It's better than using assembly plugin as it tries to mergeresources correctly (if you add the corresponding transformer). 

For Java Modules (>=JDK9) assembly and shade just merge everything togehter, which actionally is not the idea of modularization. But when googling they say that it currently the best practice as many libraries dont support yet modularization, and so doesnt maven fat-jars...

MarkusRothSBB:
I use fat-jar-with-spark. My goal is to create a standalone jar that needs to local spark installation and no further additions to the classpath when calling it. This is for running in standalone mode in a docker container with the ability to access the Hadoop cluster without relying on a spark installation in the image. The goal is to build a docker image in Jenkins from an SDL application that can be deployed to OpenShift. I would prefer to have the spark dependency added as a maven dependency rather than installing it during creation of the docker image.

Thanks for the improvements on the use of the shade plugin! I have tested them in this use case, they work fine.

zzeekk:
I also fixed the problem in sdl-examples, that fat-jar still included all spark dependencies, see https://github.com/smart-data-lake/sdl-examples/commit/32e1bf956a261bb830dce88792b7c0df1a6fe47c.

zzeekk:
@MarkusRothSBB: Interesting. Note that we are working on a Dockerfile including Spark installation as part of sdl-examples: https://github.com/smart-data-lake/sdl-examples/issues/12

pgruetter (approved):
Looks good, thanks for the suggestion :+1:

PR #231, opened 2020-11-24
use information catalog to check if table exists if noop dialect

zzeekk:
1) sdl-example doesnt work anymore because isTableExisting throws exception for hsqldb if table doesnt exist. Reason: The noop jdbc dialect returns a query that throws an exception if the table doesn't exists.
-> use information_schema if noop dialect.
2) fix lost changes in OracleSQLCatalog to check for tables and views in isTableExisting


pgruetter (approved):
Wonder if we ever get all dialects working at the same time 8-)

PR #234, opened 2020-11-25
Enable CustomDfDataObject to deliver a custom schema during init

### LINKED ISSUE #232: CustomDfCreator.exec called during init
**Describe the bug**
During the init phase, DataFrames are created for Spark feeds. To create the DataFrame of a CustomDfCreator, exec is called on it. Then exec is called again during apply. With real Spark DataFrames this is not a problem, because DataFrame creation is lazy. If a CustomDfCreator calls a Webservice to manually populate a DataFrame, there is no lazyness involved, and the Webservice is called twice.

**To Reproduce**
```
dataObjects {
  webservice {
    type = CustomDfDataObject
    creator {
      class-name = <myDfCreator>
    }
  }

  datasink {
   ...
  }
}
actions {
  ingest-from-webservice {
    type = CopyAction
    input-id = webservice
    output-id = datasink
    metadata {
      feed = myFeed
    }
  }
}
```

**Expected behavior**
Exec on a CustomDfCreator must only be called once.

The problem is in SparkAction.scala:109. If the input DataFrame is empty, and no UserDefinedSchema is provided, the DataFrame is created to extract the schema. As far as I can tell, there is no way to provide a user-defined schema to a CustomDfDataObject. The only option where you can do that is SparkFileDataObject.

MarkusRothSBB:
-Enable CustomDfDataObject to deliver an empty DataFrame with a custom schema during init phase
-To make this possible, add ActionPipelineContext as an implicit parameter to CanCreateDataFrame.getDataFrame

Without this change, CustomDfDataObject.creator.exec is called twice, once during init and once during exec. That might be expensive, e.g. when calling a web service.

closes #232


zzeekk (approved):
Great work!

PR #237, opened 2020-11-27
Support for SAP Hana schema catalog

MarkusRothSBB:
SapHanaSQLCatalog in JdbcTableConnection

To support reading from SAP Hana databases


MarkusRothSBB:
This does not work yet if the SAP Hana table has special characters like periods in it. This can be worked around by adding a custom query to the DataObject for now, though.

zzeekk (approved):
lgtm

PR #249, opened 2020-11-30
Release version 1.2.1

zzeekk:
Seems version 1.2.0 is bricked on bintray from all my experiments. I propose to retry with version 1.2.1. There will be no release with version 1.2.0 in consequence...

Error on bintray:
Error:  Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project sdl-core_2.11: Failed to deploy artifacts: Could not transfer artifact io.smartdatalake:sdl-core_2.11:pom:1.2.0 from/to bintray-smart-data-lake-smart-data-lake (https://api.bintray.com/maven/smart-data-lake/smart-data-lake/smart-data-lake/;publish=1): Transfer failed for https://api.bintray.com/maven/smart-data-lake/smart-data-lake/smart-data-lake/;publish=1/io/smartdatalake/sdl-core_2.11/1.2.0/sdl-core_2.11-1.2.0.pom 409 Conflict

### DIFF
--- a/.github/workflows/release-spark3.yml
+++ b/.github/workflows/release-spark3.yml
@@ -104,7 +104,9 @@ jobs:
         current-version: ${{ env.MVN_RELEASE_VERSION }}
         version-fragment: bug
     - name: Develop-spark3 Set Next Snapshot version
-      run: sed -i "/${MVN_RELEASE_VERSION}/{s//${{ steps.bump_version.outputs.next-version }}-SNAPSHOT/;:p;n;bp}" pom.xml
+      run: |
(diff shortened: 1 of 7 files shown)


zzeekk:
@pgruetter: Shall we go with 1.2.1, or try to involve bintray support? What do you think?

pgruetter:
> @pgruetter: Shall we go with 1.2.1, or try to involve bintray support? What do you think?

On Bintray I can't see anything related to 1.2.0 so the error seems to be wrong or misleading. 

It depends on how quickly we would need the release. If we can wait another day or two, I would try to involve Bintray. Last time their response time was pretty good.

PR #255, opened 2020-12-02
Enable private auth headers in WebserviceFileDataObject

MarkusRothSBB:
Similar to the other credentials used in WebserviceFileDataObject, also pass authHeader through the CredentialsUtil, so that the ENV#MY_SECRET_AUTH_HEADER format works for auth headers.

I have a hardcoded auth header for a webservice that I don't want to hardcode into a config file and check in to source control.


zzeekk (approved):

pgruetter (approved):
As discussed, sorry it took a bit longer.

Issue #229, opened 2020-11-19
Add "virtual" partition columns for reading from JdbcTableDataObject


zzeekk:
**Is your feature request related to a problem? Please describe.**
In order to use PartitionDiffMode with JdbcTableDataObject's, the DataObject should implement CanHandlePartitions.
It is not so easy to get db partition information over jdbc, but SDL partitions and Db partitions also dont necessarily need to be linked with each other.

**Describe the solution you'd like**
Implement CanHandlePartitions for JdbcTableDataObjects, but dont link it with Db partitions. It is up to the user if he specifies the db partition columns as partition columns in SDL.
For listPartitions a query "select distinct partitionCols from tableX" is executed.
deletePartitions can be implemented by "delete from table where partitionCols=partitionValues"

In the future this could be refined, so that we can issue customized queries for list/deletePartitions depending on the db.

PR #257, opened 2020-12-03
Feature/229 jdbc partition columns

zzeekk:
see #229

Note that i implemented special handling for case sensitive partition column names. We have a unit test where the DB table is created from a Spark DataFrame. Here all DataFrame columns with lowercase letters become case sensitiv column names in the DB table!


pgruetter (approved):
LGTM

PR #258, opened 2020-12-03
fix rewriting pom on release-sonatype

zzeekk:
Properties ${scala.minor.version} are not replaced in pom on maven central for release 1.2.0/2.0.0.
Reason is a conflict between scala-cross-maven-plugin and scala-doc-nolinkwarnings profile.
This PR fixes it.

Background:
scala-cross-maven-plugin searchs for an active profile starting with "scala-" to replace properties in the pom. Unfortunately during deploy scala-doc-nolinkwarnings was enabled and choosen. As a result no expressions ${scala.minor.version} were replaced in the pom.
This fix removes scala-doc-nolinkwarnings completely and configures execution for creation of scala-doc directly in the release-sonatype profile, including "-no-link-warnings" argument.


pgruetter (approved):
Nice detective work!

PR #261, opened 2020-12-07
Release 1.2.1 - increase nexus staging timeout

zzeekk:
Default is 5min, we got errors "Warning: TIMEOUT after 303.8 s" while deployment...


pgruetter (approved):
Let's try it

Issue #268, opened 2020-12-08
Hook for custom partition execution mode


zzeekk:
**Is your feature request related to a problem? Please describe.**
Sometimes existing ExecutionMode implementations are not sufficient to determine what data is available to process, for example if there is a specific method of the source system to be called to query the data that is available.

**Describe the solution you'd like**
It should be possible to implement custom execution mode logic in Scala. As ExecutionMode is a sealed trait, it's not possible to create a custom implementation by extending ExecutionMode. We should create a **CustomPartitionExecutionMode** which extends ExecutionMode and takes a className of a class implementing method "def apply(session: SparkSession, options: Map[String,String], mainInput: DataObject, mainOutput: DataObject, givenPartitionValues: Seq[Map[String,String]], context: ActionPipelineContext): Option[Seq[Map[String,String]]]".
Returning None means no execution mode is applied, otherwise Seq[Map[String,String]] is used as partition values of subfeed. If there is no data to process throw NoDataToProcessWarning.

PR #270, opened 2020-12-09
Custom partition execution mode

zzeekk:
see #268


zzeekk:
I made some classes available outside of package io.smartdatalake, because they are use for such custom implementations. I tagged them with spark DeveloperApi annotation.

pgruetter (approved):
LGTM, sorry for the delay

zzeekk:
also merged to develop.

Issue #190, opened 2020-09-08
Validate column names when writing Jdbc table


zzeekk:
**Is your feature request related to a problem? Please describe.**
When action1 is writing a Jdbc table, Spark doesnt validate column names but just their datatypes according to column order. This works the same way as an sql insert statement.
The problem is that in the lineage the DataFrame is propagated with the original column names to the next action2. But if we start action2 separately column names are read from the Jdbc table and might be different.

**Describe the solution you'd like**
Before writing a DataFrame to a Jdbc table, columns should be selected according to the schema of the Jdbc table. Like this we are more tolerant regarding the order, and more restrictive regarding the column names. But we need to pay attention regarding case sensitive names.

PR #271, opened 2020-12-11
Jdbc validate column names / save mode

zzeekk:
see #190 and #229


pgruetter (approved):
This seems to be included in https://github.com/smart-data-lake/smart-data-lake/pull/272 already? Same comment about deleteTable applies then.

zzeekk:
Also merged to develop

PR #273, opened 2020-12-15
Add CustomFileDataObject

MarkusRothSBB:
Add new CustomFileDataObject. It works like a CustomDfDataObject, except that it creates a file rather than a DataFrame

If you want to read from a web service that requires some custom logic during querying, and want to save the raw data received from the service as a file before transforming it into Dataframes, you can use this new CustomFileDataObject. I am sure there are other uses as well.

The CustomFileDataObject does not have a path, and is not writable. It is just a custom action that creates a byte stream.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/customlogic/CustomFileCreator.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/customlogic/CustomFileCreator.scala
@@ -19,7 +19,7 @@
 
 package io.smartdatalake.workflow.action.customlogic
 
-import java.io.ByteArrayInputStream
+import java.io.{ByteArrayInputStream, InputStream}
 
 import io.smartdatalake.util.hdfs.HdfsUtil
 import io.smartdatalake.util.misc.CustomCodeUtil
@@ -28,13 +28,13 @@ import org.apache.spark.sql.SparkSession
 trait CustomFileCreator extends Serializable {
 
   /**
-   * This function creates a [[ByteArrayInputStream]] based on custom code.
+   * This function creates a [[InputStream]] based on custom code.
    *
    * @param session the Spark Session
    * @param config  input config of the action
    * @return a stream containing the custom file data as a sequence of bytes
    */
-  def exec(session: SparkSession, config: Map[String, String]): ByteArrayInputStream
+  def exec(session: SparkSession, config: Map[String, String]): InputStream
 }
 
 case class CustomFileCreatorConfig(className: Option[String] = None,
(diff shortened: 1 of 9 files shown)


pgruetter:
Actually looks good to me and your test cases run fine on my machine, so I fail to see the problem of GitHub Actions:
`CustomFileDataObjectTest.scala:43: not found: value testSession`
@zzeekk : Can you reproduce the error?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/CustomFileDataObject.scala
```diff
@@ -0,0 +1,76 @@
...
+  override def createInputStream(path: String)(implicit session: SparkSession): InputStream = {
+    creator.exec
+  }
+
+  override def factory: FromConfigFactory[DataObject] = CustomFileDataObject
+
+  override def partitionLayout(): Option[String] = None
+
+  override def getFileRefs(partitionValues: Seq[PartitionValues])(implicit session: SparkSession): Seq[FileRef] = {
+    Seq(FileRef("", "result", PartitionValues(Map())))
```
zzeekk, line 47: Can we name the default File as follows? It's easier to debug.
```suggestion
    Seq(FileRef("custom", "custom", PartitionValues(Map())))
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/customlogic/CustomFileCreator.scala
```diff
@@ -0,0 +1,79 @@
...
+trait CustomFileCreator extends Serializable {
+
+  /**
+   * This function creates a [[ByteArrayInputStream]] based on custom code.
+   *
+   * @param session the Spark Session
+   * @param config  input config of the action
+   * @return a stream containing the custom file data as a sequence of bytes
+   */
+  def exec(session: SparkSession, config: Map[String, String]): ByteArrayInputStream
```
zzeekk, line 37: I suggest to change the interface to return an InputStream and not a ByteArrayInputStream. This is more general and doesn't force the implementation to materialize the data as array.
```suggestion
  def exec(session: SparkSession, config: Map[String, String]): InputStream
```

zzeekk (changes requested):
Nice new DataObject. In the future we could also think about make it writable, e.g. use as custom file output...

sdl-core/src/main/scala/io/smartdatalake/workflow/action/customlogic/CustomFileCreator.scala
MarkusRothSBB: All good points. I changed the signature to use InputStream. Regarding being able to write to a CustomFileDataObject: Would be nice, but there are some questions that need to be addressed before doing so.

zzeekk (approved):

zzeekk:
Also merged to develop-spark3

Issue #104, opened 2020-06-04
Warn when config contains multiple actions with the same name


Geheiner:
When two actions have the same name in the config, only the last one gets taken into account and all the previous ones get discarded silently.

As most actions look quite similar, you may want to copy-paste to create them. When you do this and forget to make the action-name unique, it can be difficult to find out why some actions get ignored.

It would be nice if the smart datalake would produce a warning when two actions have the same name.

It is probably not so easy to implement this issue, since overwriting earlier occurences of config-objects by the last one is a feature of the HOCON-Language.

swissrussdev:
I agree this would be extremely difficult to implement, as the elimination of duplicate entries is done by Typesafe. In fact, objects with duplicate keys could be distributed across multiple confs on the classpath, so a validation at file level also seems impractical.

Currently I think the best way of working around this is by generous use of reuse blocks for standard elements. That way the action blocks remain compact and it is easier to spot duplicates. Not ideal - but the best I can suggest without rewriting Typesafe!

pgruetter:
I'm not sure if we should implement that. As noted, this is a feature of the Typesafe library:
https://github.com/lightbend/config/blob/master/HOCON.md#duplicate-keys-and-object-merging
You can explicitly merge two objects. Overwriting could also make sense in a large project where you have a hierarchy of configuration files. A warning could be unwarranted in this situation.

In any case, it's not impossible to implement. The right place is probably in the ConfigLoader, see here:
https://github.com/smart-data-lake/smart-data-lake/blob/d8f21f087e448502dfbe144a1e055e87fb48850d/src/main/scala/io/smartdatalake/config/ConfigLoader.scala#L104
At this location, you have a list of all configuration files that will be used by the ConfigParser. You could to some low level operations on all these files and check for duplicates.

swissrussdev:
Definitely doable (default classpath loader needs some extra logic to get the files). Also needs consistent treatment of HOCON, JSON and props. 
The issue of overwriting I don't see as critical : implicit overrides of DO/Action configs are just an accident waiting to happen and are better dealt with using references to template objects. I personally see no issue with issuing a warning in this case. If someone wants to implement it, why not?

zzeekk:
Solution implemented: 
As all files are parsed as Hocon config object before their are merged and resolved, we can check for duplicate entries for Connections, DataObjects and Actions before merging the config objects.
[...]

PR #279, opened 2020-12-27
Feature/104 fail on configuration objects defined in multiple locations

zzeekk:
see #104: if a configuration objects are defined in multiple locations, a ConfigurationException is thrown.

To be discussed if an exception is the desired behaviour, or if we just want to log a Warning/Error. 
In my opinion we can be strict here, as it's really confusing to define the same configuration object in different locations and have it merged by Hocon.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/config/ConfigLoader.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/config/ConfigLoader.scala
@@ -22,6 +22,7 @@ import java.io.InputStreamReader
 
 import com.typesafe.config.{Config, ConfigFactory}
 import io.smartdatalake.config.SdlConfigObject.{ActionObjectId, ConnectionId, DataObjectId}
+import io.smartdatalake.definitions.Environment
 import io.smartdatalake.util.hdfs.HdfsUtil
 import io.smartdatalake.util.misc.{EnvironmentUtil, SmartDataLakeLogger}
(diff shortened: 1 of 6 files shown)


pgruetter:
> To be discussed if an exception is the desired behaviour, or if we just want to log a Warning/Error. 
> In my opinion we can be strict here, as it's really confusing to define the same configuration object in different locations and have it merged by Hocon.

I can see complex setups with nested config files in which you define a default object in a parent config and overwrite them in a different config file. 
In these special cases, you would need at least the option to allow overwrites, don't you think?

zzeekk:
> > To be discussed if an exception is the desired behaviour, or if we just want to log a Warning/Error.
> > In my opinion we can be strict here, as it's really confusing to define the same configuration object in different locations and have it merged by Hocon.
> 
> I can see complex setups with nested config files in which you define a default object in a parent config and overwrite them in a different config file.
> In these special cases, you would need at least the option to allow overwrites, don't you think?

Well, the default object already needs to have the correct id. You can't create a default object with a generic id... to create templates we can use hocon inheritance.
But i would not override an action defined in file A in another file... in my opinion it's not a good pattern.
Nevertheless we could create an option to allow duplicates. I think its too much of a detail to add it as command line option, but we could read a system property?

pgruetter (approved):
Thanks for the addition, I think the environment variable is a good solution.

zzeekk:
Also merged to branch develop

Issue #112, opened 2020-06-05
Automatically determine if persist should be set to true for Actions


Geheiner:
The output of most Actions in the Smart Datalake Builder have to be read again afterwards.
It would make sense if for those Actions, the results  automatically get persisted so that the underlying dataframes don't have to be recomputed everytime they are accessed.
In the current implementation, persist is set to false, which means that I have to set it to true in all my Actions that are not in the final output.

zzeekk:
The implementation counts DataFrames reused in subsequent actions in init-phase, so that we can "persist/cache" them before the write operation in exec-phase.
It also tracks when postExec function when no Action needs the cached DataFrame anymore, and releases it automatically.
In my opinion the "persist" property of Actions to manually control caching isn't needed anymore, but it is left for the moment for backward compatibility as it is no problem if the same DataFrame is cached multiple times.
The feature can be disabled by setting java system property "sdl.enableAutomaticDataFrameCaching=false".

PR #281, opened 2021-01-04
implement automatic caching/release of DataFrames that are used multiple times

zzeekk:
See #112:
The implementation counts DataFrames reused in subsequent actions in init-phase, so that we can "persist/cache" them before the write operation in exec-phase.
It also tracks when postExec function when no Action needs the cached DataFrame anymore, and releases it automatically.
In my opinion the "persist" property of Actions to manually control caching isn't needed anymore, but we can leave it for the moment for backward as it is no problem if the same DataFrame is cached multiple times.

### DIFF
--- a/docs/Features.md
+++ b/docs/Features.md
@@ -79,3 +79,7 @@ The following is a list of implemented and planned (Future) features of Smart Da
   * Spark data pipeline simulation (acceptance tests)
 * Support for Deployment
   * Dry-run
+
+##### Spark Performance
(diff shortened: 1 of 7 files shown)


zzeekk:
@MarkusRothSBB: I put you on the PR mainly to inform you about this feature. If pgruetter makes a review that would be sufficient, but you are welcome to make review as well.

MarkusRothSBB (approved):
LGTM.

pgruetter (approved):
👍

zzeekk:
Also merged to branch develop.

Issue #269, opened 2020-12-09
Fat-jars of sdl-jms/kafka/splunk should not include sdl-core


zzeekk:
**Describe the bug**
To build&deploy SDL with jms/kafka/splunk you need a fat-jar of sdl-core and sdl-jms/kafka/splunk. For now fat-jars of sdl-jms/kafka/splunk include also sdl-core, which makes each of them about 100Mb big. For sure it's possible to take sdl-core and sdl-jms/kafka/splunk in the classpath, but 400Mbs because of duplication of sdl-core are not optimal.

**Expected behavior**
Scope of sdl-core should be set to provided for fat-jar of sdl-jms/kafka/splunk

zzeekk:
I fixed this by accident directly on develop-spark3, but at least it's fixed ;-)

Also merged to develop.

PR #290, opened 2021-01-18
fix compilation of module fat-jars with scala-2.11 (#269)

zzeekk:
artifact spark-tags needs to have compile scope for compilation of fat-jar for modules with scala-2.11

Through profile fat-jar scope of sdl-core is set to provided in modules. 
Although artifacts with scope compile are available for compile, compiling with scala-2.11 fails because of missing annotations defined in spark-tags. Setting scope of spark-tags to compile helps.

The error doesnt exist with spark-2.12. This fixed is therefore not needed on SDL 2.x.

@pgruetter: @MarkusRothSBB would like to have a 1.2.x release including this fix asap if possible

### DIFF
--- a/sdl-jms/pom.xml
+++ b/sdl-jms/pom.xml
@@ -39,6 +39,16 @@
 				<!-- do not include sdl-core into fat-jar for this module, as sdl-core is normally put into classpath separately -->
 				<sdl-core.scope>provided</sdl-core.scope>
 			</properties>
+			<dependencies>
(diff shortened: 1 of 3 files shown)


pgruetter (approved):
Fine be me. If there's only this PR for a release, do you still want to increase the version number on the 2.x.x branch? Otherwise the numbers will be out-of-sync and it gets more complicated comparing the two.

PR #294, opened 2021-01-20
Bugfix/merge lost commits

zzeekk:
Some changes got lost in previous merges. I cherry picked them again and made the unit tests work.


pgruetter (approved):
Need to discuss how to avoid losing commits. Guess it's hard to find a better solution :thinking:

PR #298, opened 2021-01-22
small bug fixes

zzeekk:
- remove config file permission check as it doesnt work on all hadoop filesystems
- allow appName in state filename to contain underscore
- workaround enableOverwriteUnpartitionedSparkFileDataObjectAdls
- enhance RawFileDataObject to read/write custom Spark data formats


pgruetter (approved):
Tested, looks good

PR #304, opened 2021-01-27
Bugfix/positive number of partitions required

zzeekk:
Change:
- change _run_id partition column name to run_id

Fix:
- dont automatically repartition if numInitialHdfsPartitions = -1
- make sure numPartitionsRequired is not equal to 0

Note that these changes didn't resolve the bug we where facing, nevertheless they are good improvements.


pgruetter (approved):
:+1:

Issue #313, opened 2021-02-07
Action with many partitions is slow


zzeekk:
**Describe the bug**
Performance is slow with directories/tables with many partitions. Action takes ~15min creating a new partition with not much data. There are lots of the following entries in the logs:
INFO InMemoryFileIndex: Start listing leaf files and directories. Size of Paths: 0; threshold: 32

**To Reproduce**
Databricks Runtime 7.5 (includes Apache Spark 3.0.1, Scala 2.12, Hadoop 2.7.4), details see https://docs.databricks.com/release-notes/runtime/7.5.html
Input are CSV Files partitioned by 2 levels, first level has ~1100 partitions
Output is Hive table, partitioned by 1 level with ~1100 partitions

**Expected behavior**
Should not take longer than 1min.

**Context**
Maybe we have this problem:
https://stackoverflow.com/questions/39513505/spark-lists-all-leaf-node-even-in-partitioned-data

zzeekk:
@pgruetter - for you information

zzeekk:
Use saveMode = SDLSaveMode.OverwriteOptimized to write SparkFileDataObjects and HiveTableDataObjects without using dynamic partitioning.

PR #315, opened 2021-02-12
Feature/313 sdl savemode overwrite optimized

zzeekk:
see #313


pgruetter (approved):
Looks good, pushed a typo fix

Issue #316, opened 2021-02-13
Implement renaming files for partitioned DataObjects and with numberOfTasksPerPartition>1


zzeekk:
**Is your feature request related to a problem? Please describe.**
Renaming files to a well defined filename is not possible for partitioned DataObjects or if numberOfTasksPerPartition>1. But it is needed to export partitioned data.

**Describe the solution you'd like**
If numberOfTasksPerPartition=1 and the DataObject is partitioned, one file is created per partition and it should be renamed to the specified filename.
If numberOfTasksPerPartition>1, the files should get a number, e.g. filename=data.csv -> files renamed are data.1.csv, data.2.csv,...

Issue #314, opened 2021-02-07
Implement stop if no data for FileSubFeedAction


zzeekk:
**Is your feature request related to a problem? Please describe.**
FileTransferAction supports incremental processing by deleteDataAfterRead=true. Sometimes there are no new files to process. In these case it should be possible to stop the SDL run like it can be done with PartitionDiffMode.stopIfNoData=true for SparkAction's. 

**Describe the solution you'd like**
I think that deleteDataAfterRead=true is something which actually would belong into an execution mode. I therefore suggest to create an execution mode FileIncrementalMoveMode. It would implicitly delete data after read, the property deleteDataAfterRead would become deprecated. It will have another property stopIfNoData, throwing NoDataToProcessWarning if no files are found in the input.

zzeekk:
NoDataToProcessDontStopWarning isn't flexible enough if you have a CustomSparkAction with multiple inputs, and you want to execute the Action unless all inputs have no data.
It also doesnt support to skip several actions of the DAG, and maybe at the end of the DAG execute one of them again (e.g. to send a report).
If propose the following solution: every Action gets an executionCondition attribute, which can be used to check the state of the input subFeeds and skip the action if needed. Additionally a SubFeed get's an additional attribute isSkipped, which is set to true if the Action producing the SubFeed was skipped.
By Default a skipped SubFeed can be read by the next Action but get no data (empty DataFrame). If the Action wants to read a skipped SubFeed with the existing Data of the DataObject, an executionMode has to be set, e.g. ProcessAllMode to read all existing data.

PR #318, opened 2021-02-13
Feature/316 extend renaming files

zzeekk:
see #316 + ignore numInitialHdfsPartitions if Spark 3.0 AQE is enabled


pgruetter (approved):
Works flawlessly. Tested locally with a CSV file using the following configuration:
`sparkRepartition { numberOfTasksPerPartition = 20, filename = weatherTemperature.csv}`

which produces 20 CSV files
```
weatherTemperature.1.csv
...
weatherTemperature.20.csv
```

pgruetter:
One comment though: It seems that quite some code from #314 slipped in here, didn't it? Anyway, going to review that one next so you can merge both anyway.

zzeekk:
> One comment though: It seems that quite some code from #314 slipped in here, didn't it? Anyway, going to review that one next so you can merge both anyway.

Yes, thats right. It's build on top of #314. That one should be merged first. Sorry, i forgot to mention this.

Issue #325, opened 2021-03-07
Validate schema on write if configured


zzeekk:
**Is your feature request related to a problem? Please describe.**
When writing to a DataObject with schema defined, DataFrame should be validated against it on write.

**Describe the solution you'd like**
Add validation of schema on write

PR #327, opened 2021-03-12
Feature/325 validate schema on write

zzeekk:
see #325

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -324,3 +324,25 @@ Requirements:
 See Readme of [sdl-examples](https://github.com/smart-data-lake/sdl-examples) for a working example and instructions to setup python environment for IntelliJ  
 
 How it works: under the hood a PySpark DataFrame is a proxy for a Java Spark DataFrame. PySpark uses Py4j to access Java objects in the JVM.
+
+## Schema Evolution
+SmartDataLakeBuilder is built to support schema evolution where possible. This means that data pipelines adapt them self automatically to additional or removed columns and changes of data types if possible.
+
+To assert that a defined list of columns is always present in the schema of a specific DataObject, use its `schemaMin` attribute to define a minimal schema. The minimal schema is validated on read and write with Spark.
+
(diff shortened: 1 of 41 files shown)


docs/Reference.md
```diff
@@ -324,3 +324,25 @@ Requirements:
 See Readme of [sdl-examples](https://github.com/smart-data-lake/sdl-examples) for a working example and instructions to setup python environment for IntelliJ  
 
 How it works: under the hood a PySpark DataFrame is a proxy for a Java Spark DataFrame. PySpark uses Py4j to access Java objects in the JVM.
+
+## Schema Evolution
+SmartDataLakeBuilder is built to support schema evolution where possible. This means that data pipelines adapt them self automatically to additional or removed columns and changes of data types if possible.
```
pgruetter, line 329:
```suggestion
SmartDataLakeBuilder is built to support schema evolution where possible. This means that data pipelines adapt themselves automatically to additional or removed columns and changes of data types if possible.
```

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/SchemaValidation.scala
```diff
@@ -50,21 +51,49 @@ private[smartdatalake] trait SchemaValidation { this: DataObject =>
...
+
+  /**
+   * Validate the schema of a given Spark Data Frame `df` against a given expected schema.
+   *
+   * @param df The data frame to validate.
+   * @param schemaExpected The expected schema to validate against.
+   * @param role role used in exception message. Set to read or write.
+   * @throws SchemaViolationException is the `schemaMin` does not validate.
+   */
+  def validateSchema(df: DataFrame, schemaExpected: StructType, role: String): Unit = {
```
pgruetter, line 81: For the role, shouldn't we use an Enumeration here or do you think that's overkill?

zzeekk: As we dont use it for any logic (no tests/matching on it), but it's just for the exception text i would leave it like this. Otherwise yes, should be an enumeration.

pgruetter (approved):
LGTM

hiltpold (approved):

Issue #221, opened 2020-11-15
Remove enableJdbcCaseSensitivity


pgruetter:
**Is your feature request related to a problem? Please describe.**
As discussed in #219 we can probably get rid of the environment variable enableJdbcCaseSensitivity. 

**Describe the solution you'd like**
Remove Enviornment.enableJdbcCaseSensitivity and any functions that use it.

**Further requirements**
We should follow common conventions in relational databases when it comes to case sensitivity. In Oracle, table names are case insensitive unless you quote them: `"Schema"."Name"` is case sensitive, `schema.name` is not. In MSSQL backticks are used. 

In SDL/Hocon config quotes can be added to values as follows:
```
""Name""
`Name`
"`Name`"
```
The first " only tells HOCON that the value should be interpreted as a string and is removed by the Hocon parser.
If a db or table name starts with quotes, SDL needs to handle these names case sensitive. 
Note that in SQL the quoted names can also contain other special characters like slashes, dots or hyphens.

pgruetter:
Analyzed the requirements as stated above and I have to revise that:
MSSQL uses brackets [schema].[table] not backticks, backticks are used in MySQL i.e.

We should really leave it up to the JDBC driver to handle case sensitivity. So for this issue:
- Remove Environment.enableJdbcCaseSensitivity as originally stated
- Make sure that escape sequences in Hocon are passed on correctly, i.e. for SQL Server:
```
table {
   db = "["dbo"]" 
   name = "["nameWithSpecialCharacters"]"
}
```
or for case sensitive table names in Oracle:
```
table {
   db = "\""schema"\""
   name = "\""caseSensitiveTableName"\""
}
```

zzeekk:
> ```
> table {
>    db = "["dbo"]" 
>    name = "["nameWithSpecialCharacters"]"
> }
> ```
I agree, thats much better. But why do you use "["nameWithSpecialCharacters"]" and not "[nameWithSpecialCharacters]" for mssql? I guess the second " is absorbed by Hocon as well, but then it's not needed at all, right?

PR #328, opened 2021-03-17
Remove enableJdbcCaseSensitivity

pgruetter:
Solves #221 

Currently, case sensitivity is not handled cleanly and it's difficult to enter special characters in schema and / or table names. 
This PR tries to improve the situation by forwarding the users HOCON config (including quotes) to the JDBC driver.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
@@ -127,8 +127,23 @@ private[smartdatalake] abstract class SQLCatalog(connection: JdbcTableConnection
   // get spark jdbc dialect definitions
   protected val jdbcDialect: JdbcDialect = JdbcDialects.get(connection.url)
   protected val isNoopDialect: Boolean = jdbcDialect.getClass.getSimpleName.startsWith("NoopDialect") // The default implementation is used for unknown url types
-  def isDbExisting(db: String)(implicit session: SparkSession): Boolean
+  // use jdbcDialect to define identifiers used for quoting
+  protected val (quoteStart,quoteEnd) = {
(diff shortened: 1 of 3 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
```diff
@@ -149,68 +164,27 @@ private[smartdatalake] object SQLCatalog {
  */
 private[smartdatalake] class DefaultSQLCatalog(connection: JdbcTableConnection) extends SQLCatalog(connection) {
   override def isDbExisting(db: String)(implicit session: SparkSession): Boolean = {
-    val cntTableInCatalog =
-      if (Environment.enableJdbcCaseSensitivity)
-        s"select count(*) from INFORMATION_SCHEMA.SCHEMATA where TABLE_SCHEMA='$db'"
-      else
-        s"select count(*) from INFORMATION_SCHEMA.SCHEMATA where upper(TABLE_SCHEMA)=upper('$db')"
+    val cntTableInCatalog = s"select count(*) from INFORMATION_SCHEMA.SCHEMATA where TABLE_SCHEMA=$db"
```
zzeekk, line 167: What if <db> is a case sensitive name? Here we need to implement a more specific query, as if the <db> is configured with quotes in Hocon, it will never be found in a catalog. We would need to check if <db> is quoted, if yes remove the quotes and make a case sensitive comparision, otherwise we make a case insensitive comparision (upper both sides).
What do you think?

pgruetter: You're absolutely right of course, just passing the quoted string to the database doesn't work in case of isDbExisting. 
As discussed, fixed in https://github.com/smart-data-lake/smart-data-lake/pull/328/commits/49f04537424e96689962403388d70ffc04bf3da2.

zzeekk (approved):

Issue #329, opened 2021-03-22
Improve JDBC SDL query performance


zzeekk:
**Is your feature request related to a problem? Please describe.**
prepare and init phase with many JDBCTableDataObject is slow because JDBC connection is recreated for every query like db/table exists and pre/postSQL

**Describe the solution you'd like**
Reuse JDBC connection by creating a pool like for SFTP connections

PR #330, opened 2021-03-22
Feature/329 jdbc performance

zzeekk:
see #329

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
@@ -23,8 +23,10 @@ import java.sql.{DriverManager, ResultSet, Statement, Connection => SqlConnectio
 import com.typesafe.config.Config
 import io.smartdatalake.config.SdlConfigObject.ConnectionId
 import io.smartdatalake.config.{FromConfigFactory, InstanceRegistry}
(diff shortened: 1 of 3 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
```diff
@@ -109,6 +133,17 @@ case class JdbcTableConnection( override val id: ConnectionId,
     } else Map()
   }
 
+  // setup connection pool
+  val pool = new GenericObjectPool[SqlConnection](new JdbcClientPoolFactory)
+  pool.setMaxTotal(maxParallelConnections)
+  pool.setMaxIdle(1) // keep max one idle jdbc connection
```
pgruetter, line 139: Maybe we could make these parameters configurable? Especially the timeout: if you have many small feeds you might want to increase the value.

pgruetter (approved):
Looks good, added one comment but not a blocker.

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/JdbcTableConnection.scala
zzeekk: Good idea, i added connectionPoolMaxIdleTimeSec in JdbcTableConnection and SftpFileRefConnection

PR #334, opened 2021-03-26
implement support to register python UDFs for Spark SQL transformations

zzeekk:
Support to register python udfs in global config, which can then be used in Spark SQL.

nice feature


pgruetter (approved):
Good reason for the change: Because it's nice 😁

swissrussdev:
As valid a reason as any!

PR #340, opened 2021-04-16
Bugfix/div fixes

zzeekk:
- validate partition columns existing on write and read
- fix listing partitions on relative path
- implement including current data partition in KafkaTopicDataObject.listPartitions


zzeekk:
split up in smaller PR's

PR #341, opened 2021-04-16
validate partition columns existing on write and read

zzeekk:
Validate that partition columns are existing in DataFrame, when writing or reading from "DataObjects with CanHandlePartitions".

The validation doesn't cost much and ensures schema consistency early.


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HiveTableDataObject.scala
```diff
@@ -156,6 +158,7 @@ case class HiveTableDataObject(override val id: DataObjectId,
                              (implicit session: SparkSession, context: ActionPipelineContext): Unit = {
     require(!isRecursiveInput, "($id) HiveTableDataObject cannot write dataframe when dataobject is also used as recursive input ")
     validateSchemaMin(df, "write")
+    validateSchemaHasPartitionCols(df, "write")
```
pgruetter, line 161: Do you need to validate it again or is the one validation in init enough?

pgruetter (approved):

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HiveTableDataObject.scala
zzeekk: It doesn't cost much so i would do it again.

PR #342, opened 2021-04-16
fix use relative path for listing partitions

zzeekk:
Listing partitions is currently done on absolute path. Wrong results are produces if the partition layout pattern is already found in the base path of the DataObject. Note that this is a rare case as the pattern normally includes "=", but i struggled with it.

Listing Partitions is done on relative path. Path of files/directories are relativized against the base path of the DataObject before extracting partitions.


pgruetter (approved):
LGTM

PR #343, opened 2021-04-16
implement including current date partition in KafkaTopicDataObject.listPartitions

zzeekk:
There is a new flag "includeCurrentPartition: Boolean = false" on KafkaTopicDataObject to include the current date partition in the list of existing partitions. Default is to exclude it.

It might be useful to read/export all data before a scheduled maintenance.


pgruetter (approved):
👍

PR #345, opened 2021-04-21
Feature/sdl snowflake improvements

MarkusRothSBB:
Some improvements to the sdl-snowflake module:
-Ability to choose the role that Snowflake uses
-Improvement to the mixup of db/schema: Snowflake database is now called database, Snowflake schema is called db (consisent with schemas in other databases that are also called db)
-Snowflake database can and must only be set on SnowflakeTableConnection
-Snowflake schema (db) can and must only be set on SnowflakeTableDataObject
-Add the ability to add Snowflake table comments to the SnowflakeTableDataObject (since the table can get recreated for every write, the comment must be set every time)


zzeekk (approved):
lgtm :-)

PR #346, opened 2021-04-21
use options to customize kafka consumer creation

zzeekk:
KafkaConnection and KafkaTopicDataObject options are used to create consumer for detecting date partitions.
Rename datasourceOptions to options, as this is the convention applied for other DataObjects.

Customizing consumer option for detecting date partitions might be needed, e.g. timeouts.


pgruetter (approved):
LGTM

Issue #349, opened 2021-05-07
Recovery re-executes actions skipped with no-data


zzeekk:
**Describe the bug**
Recovery of a failed run re-executos actions skipped because they had no data. This should not happen as they have been skipped by intention and doesn't need to be re-executed. 

There is another problem with multiple attempts: the state file for an attempt only includes runtime information of the current attempt. If there are multiple attempts runtime information from older attempts get lost and actions are re-executed.

**Expected behavior**
Recovery should ignore actions skipped because they have no data.
Recovery should write full runtime information, including information from previous attempts.

pgruetter:
Included in Release 1.2.5 and 2.0.4.

PR #351, opened 2021-05-14
bugfix recover (#349)

zzeekk:
see #349


pgruetter (approved):
Works correctly now. On the first execution, I made a feed fail.
On the first recovery run, only the failed and subsequent feeds are executed (not the skipped feeds).
On the second recovery, same behavior.
For the third recovery, I removed the exception and the recovery successfully executed the remaining feeds. 
No data was overwritten because of re-executed feeds.

PR #360, opened 2021-06-14
Bugfix/spark tags dependency

pgruetter:
Add spark-tags dependency for sdl-deltalake

Somehow, sdl-deltalake doesn't always resolve the transitive dependency for spark-tags. When doing a fresh clone, maven would build correctly the first time. But when compiling a second time, I hit the error described here: https://stackoverflow.com/questions/42327777/spark-sql-errors. 

> error: missing or invalid dependency detected while loading class file 'package.class'. Could not access term annotation in package org.apache.spark, because it (or its dependencies) are missing.

### DIFF
--- a/sdl-deltalake/pom.xml
+++ b/sdl-deltalake/pom.xml
@@ -69,6 +69,10 @@
 		</dependency>
 		<dependency>
 			<groupId>org.apache.spark</groupId>
+			<artifactId>spark-tags_${scala.minor.version}</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.spark</groupId>
 			<artifactId>spark-sql-kafka-0-10_${scala.minor.version}</artifactId>
 			<version>${spark.version}</version>
 			<scope>runtime</scope>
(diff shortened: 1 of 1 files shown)


sdl-deltalake/pom.xml
```diff
@@ -67,6 +67,10 @@
 			<groupId>org.apache.spark</groupId>
 			<artifactId>spark-catalyst_${scala.minor.version}</artifactId>
 		</dependency>
+		<dependency>
```
zzeekk, line 70: Scope=compile should be enough. And please add a comment maybe even referencing the stackoverflow issue.

zzeekk (commented):
I had a similar issue in other projects and applied the same solution.

sdl-deltalake/pom.xml
pgruetter: Thanks, done in https://github.com/smart-data-lake/smart-data-lake/pull/360/commits/70668e7caf04a8106101e92638c749d3e87c410c.

Will wait with the merge until jms fix is finalized.

pgruetter:
@zzeekk Discussed this one already, but I need one review ;-)

zzeekk (approved):

Issue #355, opened 2021-06-03
Databricks 8.X compatibility


pgruetter:
**Is your feature request related to a problem? Please describe.**
Reach compatibility with Databricks Runtime 8.X

**Describe the solution you'd like**
SDL Version 2.0.4 does not run on Databricks Runtime 8.0 because of incompatible libraries. 
We use json4s-core version 3.6.6, Databricks updated this to a milestone (!) release 3.7.0-M5 starting with Databricks Runtime 8.0.
Json4s is used in ActionDAGRunState.scala.

**Additional context**
It's possible that more libraries will follow after json4s and this was just the first one that popped up.

pgruetter:
In json4s 3.7.0-M5, Formats was split into separate classes which is why ShortTypeHints is not found anymore.
Further changes are necessary though.

zzeekk:
@pgruetter: An alternative solution would be to get rid of json4s. I used it because kxbmap-configs was not yet able to write data. With the new version used for "fail on superfluous keys" it's also possible to write Hocon or Json. The code for writing the state files could be migrated to kxbmap-configs then...

pgruetter:
@zzeekk Good point, we should get rid of json4s now. I read into kxbmap and did a first try in branch https://github.com/smart-data-lake/smart-data-lake/tree/replace-json4s-with-kxbmap. I'm currently stuck on a few ConfigWriters and Readers so I just added some dummies so it compiles.

For example, I don't understand why ConfigWriter[RuntimeInfo] needs anything special, it's a case class that uses types that should all be defined (out-of-the-box or in ActionDAGRunState now like LocalDateTime). I used a dummy to just output toString but of course this can't be parsed back by the ConfigReader.

pgruetter:
The version from https://github.com/smart-data-lake/smart-data-lake/commit/587b79863946dd140322736a151394d6f68ab7d4 works correctly. 

It does introduce a breaking change regarding durations:
In the past, we used java's toString: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toString--
Kxbmap transforms these values according to Hocon: https://github.com/lightbend/config/blob/master/HOCON.md#duration-format

When migrating to the latest SDL version, one needs to manually update the durations in the latest state file from
`PT5.35S` to `535ms`.
Not sure if we should overwrite ConfigWriter and -Reader to be backward compatible.

zzeekk:
[...]

PR #361, opened 2021-06-26
Replace json4s with kxbmap

pgruetter:
Solves #355 by removing json4s dependency. 

Databricks Runtime 8 updated the json4s dependency from 3.6.6 to 3.7.0-M5. The newer version is not compatible with SDL.
This was a good time to get rid of json4s, as the functionality we need is now included in kxbmap. This was actually documented in ActionDAGRunState as technical debt.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1124,17 +1124,6 @@
 				</exclusions>
 			</dependency>
 			<dependency>
-				<groupId>org.json4s</groupId>
-				<artifactId>json4s-jackson_${scala.binary.version}</artifactId>
-				<version>3.6.6</version>
-				<exclusions>
-					<exclusion>
-						<groupId>com.fasterxml.jackson.core</groupId>
-						<artifactId>*</artifactId>
-					</exclusion>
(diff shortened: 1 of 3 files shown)


zzeekk:
@pgruetter: i would not remove json4s dependency management from pom.xml. This is dependency mgmt copied from Spark source... 
The problem we have with Databricks Runtime 8 is that we used json4s and compiled against the original json4s version used by Spark. But json4s is not included in our fat-jar. Now Databricks Runtime 8 has a newer json4s (and another build of Spark...) and SDL is compiled against the wrong version.
It should be enough to not use json4s in SDL. We keep dependency mgmt for our fat-jar-with-spark profile, but the normal fat-jar profile doesn't include json4s and Databricks Runtime 8 should be consistent with their Spark distribution...

pgruetter:
@zzeekk OK, makes sense. I actually tested it before removing the dependency. So what you're writing is correct, it does work on Databricks Runtime 8.

zzeekk (approved):

Issue #211, opened 2020-10-30
Define retention period for date partitioned data object


zzeekk:
For some data there exists regulatory rules about rentention period, other data get useless after some time. Some project have already implemented cleanup jobs for date partitioned data objects.

Cleaning up date partitions after a defined period of time could be integrated into SDLB by configuring a retention period:
```
dataobject-test {
  type = HiveTableDataObject
  retentionPeriod {
    period = 7d
    col = dt // must be a partition column for now
    format = yyyyMMdd // this is needed because partition column normally have dataType=string
  }
}
```
If col is not a partition column a configuration exception is thrown.
After every successfull write to the DataObject all partitions earlier than "now - retentionPeriod.period" are deleted.

zzeekk:
Another requirement is to compact partitions after a certain period of time or a defined number of new partitions.
I propose to implement a more generic approach for housekeeping. Similar to the executionMode of an Action, a DataObject can have a housekeepingMode, where different implementations can be chosen.

PR #364, opened 2021-06-30
Housekeeping for partitioned DataObjects (#211)

zzeekk:
see #211

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -348,7 +348,7 @@ Recipes for data pipelines with schema evolution:
 * HistorizeAction & DeduplicateAction supports schema evolution for incremental data. They consolidate the existing data & schema of the output DataObject with a potentially new schema of the input DataObject. Then they overwrite the whole output DataObject. They need a TransactionalSparkTableDataObject as output, which doesn't have a fixed schema, e.g. TickTockHiveTableDataObject.
 
 ## Housekeeping
-SmartDataLakeBuilder supports housekeeping for DataObjects by specifying the HousekeepingMode mode.
+SmartDataLakeBuilder supports housekeeping for DataObjects by specifying the HousekeepingMode.
 
 The following HousekeepingModes are currently implemented:
 * PartitionRetentionMode: Define partitions to keep by configuring a retentionCondition. 
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -346,3 +346,26 @@ The following list describes specific behaviour of DataObjects:
 Recipes for data pipelines with schema evolution:
 * CopyAction supports schema evolution if all data is processed by overwriting the whole output DataObject. It needs an output DataObject which doesn't have a fixed schema, e.g. HiveTableDataObject.
 * HistorizeAction & DeduplicateAction supports schema evolution for incremental data. They consolidate the existing data & schema of the output DataObject with a potentially new schema of the input DataObject. Then they overwrite the whole output DataObject. They need a TransactionalSparkTableDataObject as output, which doesn't have a fixed schema, e.g. TickTockHiveTableDataObject.
+
+## Housekeeping
+SmartDataLakeBuilder supports housekeeping for DataObjects by specifying the HousekeepingMode mode.
+
+The following HousekeepingModes are currently implemented:
+* PartitionRetentionMode: Define partitions to keep by configuring a retentionCondition. 
+  retentionCondition is a spark sql expression working with the attributes of PartitionExpressionData returning a boolean with value true if the partition should be kept.
+* PartitionArchiveCompactionMode: Archive and compact old partitions.
+  * Archive partition reduces the number of partitions in the past by moving older partitions into special "archive partitions". 
+    archiveCondition defines a spark sql expression working with the attributes of PartitionExpressionData returning archive partition values as Map\[String,String\].
+    If return value is the same as input partition values, the partition is not touched. Otherwise all files of the partition are moved to the corresponding partition. 
+    Be aware that the value of the partition columns changes for these files/records.
+  * Compact partition reduces the number of files in a partition by rewriting them with Spark.
+    compactPartitionExpression defines a sql expression working with the attributes of PartitionExpressionData returning true if this partition should be compacted.
+    Once a partition is compacted, it is marked as compacted and will not be compacted again. It is therefore ok to return true for all partitions which should be compacted, regardless if they have been compacted already.
+
+Example - cleanup partitions with partition layout dt=<yyyymmdd> after 90 days:
+```
+housekeepingMode = {
+  type = PartitionRetentionMode
+  retentionCondition = "datediff(now(), to_date(elements['dt'], 'yyyyMMdd')) <= 90"
+}
+```
--- a/sdl-core/src/main/scala/io/smartdatalake/util/misc/CompactionUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/misc/CompactionUtil.scala
@@ -1,0 +1,129 @@
+/*
+ * Smart Data Lake - Build your data lake the smart way.
+ *
+ * Copyright © 2019-2021 ELCA Informatique SA (<https://www.elca.ch>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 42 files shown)


pgruetter:
Tried to release spark-extensions 2.0.12 for you, but the build fails. Will check again later.

About the implementation:
1. Is there a reason why TickTockHiveTableDataObject does not have housekeepingMode ?
2. How would I go about compacting timestamp partitions. I tried 
`archivePartitionExpression = "if( elements('dt') < add_months(current_date, -12), map('dt', trunc(elements('dt'), \"YEAR\" )), elements)"`
but that's not working
`Undefined function: 'elements'. This function is neither a registered temporary function nor a permanent function registered in the database 'default'.;`
Not sure about the exact syntax, but how do you use "elements" ?

zzeekk:
@pgruetter: Thanks for the Release of spark-extensions. I checked - it still wants to deploy over bintray, which is shutdown. We have to migrate to deploy directly to sonatype / maven central. If we do this we might loose the history of the old artifacts. Would you mind? This happend to me with spark-temporalquery where i have created a new repository. Here we might deploy to the existing smartdatalake repository, but i dont know what happens to existing artifacts in the first deployment...

Good questions 
1) I'm not sure Housekeeping makes sense for TickTockHiveTableDataObject, because it is normally not partitioned, and for now we have only HousekeepingModes working with partitioned. Do you see a use case for Housekeeping with TickTockHiveTableDataObject? Should we at it even just for completness?
2) You can use `elements['dt']` or `elements.dt` to reference map entries. For now you only specified archivePartitionExpression, which moves a partition into another partition. To compact that partition later, you would also need to configure compactPartitionExpression which should return true for partitions which should be compacted. Note that in case of yearly archive partitions you must until all partitions of the year are archived for doing the compaction, as it is done only once.
3) thanks, deleted :-)

pgruetter:
> If we do this we might loose the history of the old artifacts. Would you mind? 

So it's possible that old versions will be removed from maven central and only the latest version remains? Not sure if this is a problem or not, but it's not very clean. The new deployment works directly in maven with a profile, right? Could we execute a manual deployment of all the old versions? Anyway, we need to switch in any case as the current configuration will never work again so let's try and re-upload older versions if needed.

> 1. I'm not sure Housekeeping makes sense for TickTockHiveTableDataObject, because it is normally not partitioned, and for now we have only HousekeepingModes working with partitioned. Do you see a use case for Housekeeping with TickTockHiveTableDataObject? Should we at it even just for completness?

Never thought about if it makes sense or not, but I see quite a few partitioned TickTockHiveTables in my configurations. So yes, even just for completeness, let's implement it too. 

> 2. You can use `elements['dt']` or `elements.dt` to reference map entries. For now you only specified archivePartitionExpression, which moves a partition into another partition. To compact that partition later, you would also need to configure compactPartitionExpression which should return true for partitions which should be compacted. Note that in case of yearly archive partitions you must until all partitions of the year are archived for doing the compaction, as it is done only once.

So is the scaladoc wrong then? In Scaladoc for HousekeepingMode it says
`archivePartitionExpression = "if( elements('run_id') < runId - 1000, map('run_id', elements('run_id') div 1000), elements)"`
Does that need to be
`archivePartitionExpression = "if( elements['run_id'] < runId - 1000, map('run_id', elements['run_id'] div 1000), elements)"`
?

Did I understand the rest correctly: I want to move old monthly partitions into a bigger yearly partition. 
I can not do it as I did, moving everything older than 6 months, because it is only executed once. So executed today, the January partition is moved into a year 2021 partition. Executed next month, February is NOT archived because 2021 is already "done" ?

Still, for this particular case, I think the compactPartitionExpression is not needed as this only compacts the file within a partition. I'm trying to combine multiple "months" partitions into yearly partitions. Correct ?

zzeekk:
Ok, I'll try to change the deployment to sonatype for spark-extensions.
Good catch, scala doc is wrong... i'll fix it.
And yes, `archivePartitionExpression = "if( elements.dt < add_months(current_date, -12), map('dt', trunc(elements.dt, \"YEAR\" )), elements)` should move monthly partitions after 12 months into yearly partitions. Then you might want to compact that yearly partition to reduce number of files after 24 months with `compactPartitionExpression = "if( elements.dt < add_months(current_date, -12)`.

docs/Reference.md
```diff
@@ -346,3 +346,26 @@ The following list describes specific behaviour of DataObjects:
 Recipes for data pipelines with schema evolution:
 * CopyAction supports schema evolution if all data is processed by overwriting the whole output DataObject. It needs an output DataObject which doesn't have a fixed schema, e.g. HiveTableDataObject.
 * HistorizeAction & DeduplicateAction supports schema evolution for incremental data. They consolidate the existing data & schema of the output DataObject with a potentially new schema of the input DataObject. Then they overwrite the whole output DataObject. They need a TransactionalSparkTableDataObject as output, which doesn't have a fixed schema, e.g. TickTockHiveTableDataObject.
+
+## Housekeeping
+SmartDataLakeBuilder supports housekeeping for DataObjects by specifying the HousekeepingMode mode.
```
pgruetter, line 351:
```suggestion
SmartDataLakeBuilder supports housekeeping for DataObjects by specifying the HousekeepingMode.
```

sdl-core/src/main/scala/io/smartdatalake/util/misc/CompactionUtil.scala
```diff
@@ -0,0 +1,129 @@
...
+   * Compacting hadoop partitions is not supported out-of-the-box by hadoop, as files need to be read with the correct format and written again.
+   * The following steps are used to compact partitions with Spark:
+   * 1. Check if compaction is already in progress by looking for a special file "_SDL_COMPACTING" in data objects root hadoop path. If it exists
+   *    and is not older than 12h exit compaction with Exception. Otherwise create/update special file "_COMPACTION". If the file is older than 12h
+   *    the compaction process is assumed to be crashed.
+   * 2. As step 5 is not atomic (delete and move are two operations), we need to check for possibly incomplete compactions of previous crashed runs and fix them.
+   *    Incomplete compactions are marked with a special file "_SDL_MOVING" in the temporary path.
+   *    Incomplete compacted partitions must be moved from temporary path to hadoop path (see step 5)
+   *    and marked as compacted (see step 6).
+   * 3. Filter already compacted partitions from given partitions by looking for "_SDL_COMPACTED" file, see step ...
```
pgruetter, line 40:
```suggestion
   * 3. Filter already compacted partitions from given partitions by looking for "_SDL_COMPACTED" file, see step 5
```
?

sdl-core/src/main/scala/io/smartdatalake/util/misc/CompactionUtil.scala
```diff
@@ -0,0 +1,129 @@
...
+   * 4. Data from partitions to be compacted is rewritten into a temporary path under this data objects hadoop path.
+   * 5. Partitions to be compacted are deleted from the hadoop path and moved from the temporary path to the hadoop path. This should be done one-by-one to reduce risk of data loss.
+   *    To recover in case of unexpected abort between delete and move, a special file "_SDL_MOVING" is created in temporary path before deleting hadoop path.
+   *    After moving the temporary path, this file is deleted again. Mark compacted partitions by creating a special file "_SDL_COMPACTED" and
+   * 6  Delete "_SDL_COMPACTING" file created in step 1.
```
pgruetter, line 45:
```suggestion
   * 6.  Delete "_SDL_COMPACTING" file created in step 1.
```

sdl-core/src/main/scala/io/smartdatalake/util/misc/SparkExpressionUtil.scala
```diff
@@ -102,6 +102,25 @@ private[smartdatalake] object SparkExpressionUtil {
     }
   }
 
+  /**
+   * Evaluate an expression against a each entry of a list of case class instances
```
pgruetter, line 106:
```suggestion
   * Evaluate an expression against a each entry of a list of case class instances
```
```suggestion
   * Evaluate an expression against each entry of a list of case class instances
```

pgruetter (approved):
Just a few minor comments. Couldn't test the functionality yet, but fine for me to merge already.

Issue #344, opened 2021-04-16
New Transformer API


zzeekk:
**Is your feature request related to a problem? Please describe.**
Problems:
- It is often not possible to reuse CustomDfTransformers between different Actions because they contain specific logic at some point.
- It would be nice to split complex CustomDfTransfomers into smaller steps, so they are visible in the Metadata with name and description, see also #46 
- It would be nice to not configure CustomDfTransfomers with generic options, but parsed from their Scala case class attributes as it's done for Actions and DataObjects

**Describe the solution you'd like**
Let Actions have a list of transformers configured instead of a single transformer. The list of transformers would be applied according to their order in the list, and pass the DataFrame from one to the other.
Each transformer entry will have the following attributes: type, name, description and any other configuration needed by their Scala case class definition.
They will be parsed in the same way as Actions or DataObjects. The type can be a generic transformer included in the SDL jars (e.g. StandardizeDatatypes, FilterBlacklist, ...) or any class name implementing the corresponding interface.
A special transformer can be used to wrap existing CustomDfTransformers. 

**Additional context**
Copy-, Historize-, and DeduplicateAction all have configurable transformations which pollute their API: columnBlacklist,                       columnWhitelist, additionalColumns, filterClause and standardizeDatatypes. Their execution order is hardcoded. They should be converted to the new Transformer API so they can be used more flexible. In a future version these attributes can then be removed.

zzeekk:
Here an example to use new transformer API:

Instead of 
```
  loadJdbc {
    type = CopyAction
    inputId = ab-csv-org
    outputId = ab-reduced-hsqldb
    transformer { 
      className = com.sample.ReduceNycCSVTransformer
    }
    metadata {
      feed = ab-jdbc
    }
  }
```
you can write and chain multiple transformations (with optional name and description)
```
  loadJdbc {
    type = CopyAction
    inputId = ab-csv-org
    outputId = ab-reduced-hsqldb
    transformers [{ 
      type = ScalaClassDfTransformer
      className = com.sample.ReduceNycCSVTransformer
      name = "Reduce NYC dataset"
      description = "This transformation selects id and name from NYC dataset"
    },{ 
      type = FilterTransformer
      filterClause = "name != 'abc'"
    }]
    metadata {
      feed = ab-jdbc
    }
  }
```

[...]

PR #368, opened 2021-07-12
new transformer API (#344)

zzeekk:
see #344, it includes an example configuration...


MarkusRothSBB (approved):
I really love these changes! It will make the framework a lot more discoverable, extendable and easy to use. Looks (very!) good to me. Looking forward to rolling this version out at SBB.

pgruetter (approved):
I totally agree, these changes are awesome and it looks like you had to put a lot of work into this. So thanks :+1: 

Please check the remarks, it looks like in the documentation you use `transformers [{` but I'm pretty sure it has to be `transformers = [{`. At least that's how I had to use it.

I also look forward in using these changes in the next project already. Maybe we can soon publish a version 2.0.5, but I'll work with the snapshot until we see that everything is working correctly.

Issue #375, opened 2021-08-11
DeltaLake: <table> is not a Delta table


pgruetter:
**Describe the bug**
When executing an action that writes into a Delta Lake table on Windows, the feed fails with the following error
```Exception in thread "main" io.smartdatalake.workflow.TaskFailedException: Task int-5g failed. Root cause is 'AnalysisException: `data/int/int-5g-antennas` is not a Delta table.' ```

**To Reproduce**
The issue can be reproduced with the following application.conf:
```
dataObjects {
  opendata-swiss-5g-antennas {
    type = WebserviceFileDataObject
    url = "https://data.geo.admin.ch/ch.bakom.mobil-antennenstandorte-5g/data/ch.bakom.mobil-antennenstandorte-5g_en.json"
  }

  stg-5g-antennas {
    type = JsonFileDataObject
    path = "data/stg/5g-antennas.json"
    schema = "`crs` STRUCT<`properties`: STRUCT<`name`: STRING>, `type`: STRING>,`features` ARRAY<STRUCT<`geometry`: STRUCT<`coordinates`: ARRAY<BIGINT>, `type`: STRING>, `properties`: STRUCT<`description`: STRING, `id`: BIGINT, `lang`: STRING, `powercode_en`: STRING>, `type`: STRING>>,`name` STRING,`type` STRING"
  }

  int-5g-antennas {
    type = DeltaLakeTableDataObject
    path = "data/int/~{id}"
    table {
      db = default
      name = "int_antennas"
      primary-key = [id]
    }
  }
}

actions {
  stg-5g {
    type = FileTransferAction
    inputId = opendata-swiss-5g-antennas
    outputId = stg-5g-antennas
    metadata {
      feed = 5g-stg
    }
  }

  int-5g {
    type = HistorizeAction
    inputId = stg-5g-antennas
    outputId = int-5g-antennas
    transformer.sqlCode = "select features.properties.id, features.geometry, features.properties.powercode_en from (select explode(features) as features from stg_5g_antennas)"
    metadata {
      feed = 5g-int
    }
  }

}
```

**Expected behavior**
When executing the above 2 feeds, it is expected that a DeltaLake table is created.

**Additional context**
Tried a similar pipeline on Linux and didn't run into theses problems.

**Comment by pgruetter** [comment]
It's not a Windows problem per se, the following code works correctly (not using SDLB):

```
def main(args: Array[String]): Unit = {
  val spark = SparkSession
    .builder()
[...]

PR #384, opened 2021-08-24
update delta lake (#375)

zzeekk:
see #375

Please note postfix -2 on the branch name, as i squashed some commits.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/hdfs/Partition.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/hdfs/Partition.scala
@@ -19,7 +19,7 @@
 package io.smartdatalake.util.hdfs
 
 import org.apache.spark.annotation.DeveloperApi
-import org.apache.spark.sql.Column
+import org.apache.spark.sql.{Column, DataFrame}
 import org.apache.spark.sql.functions._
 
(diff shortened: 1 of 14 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/hdfs/Partition.scala
```diff
@@ -137,6 +137,17 @@ private[smartdatalake] object PartitionValues {
     diffPartitionValues(existingPartitionValues, expectedPartitionValues, partitionColCombinations)
   }
 
+  /**
+   * Read DataFrame and convert to PartitionValues
+   * @param df DataFrame with partition columns only selected. All columns must be of type string.
```
pgruetter, line 142: It's possible to give something else than string, right? If it's really a "must", then you could insert an assert to check.
```suggestion
   * @param df DataFrame with partition columns only selected. All columns will be handled as string.
```

pgruetter (approved):

sdl-core/src/main/scala/io/smartdatalake/util/hdfs/Partition.scala
zzeekk: Right, done.

hiltpold (approved):

Issue #365, opened 2021-07-02
Connectivity: create DataObject to read from airbyte connectors


zzeekk:
**Is your feature request related to a problem? Please describe.**
Provide more connectors for API's of different cloud services

**Describe the solution you'd like**
Create a DataObject that executes singer tap (python script) and reads its stdout (json messages: schema, data & state).
Dynamically create a DataFrame with the schema provided by the schema message.

**Additional context**
See also https://www.singer.io/#taps, https://github.com/singer-io/getting-started/blob/master/docs/SPEC.md#state-message

zzeekk:
Maybe Airbyte has better Connectors, see: https://hackernoon.com/6-biggest-differences-between-airbyte-and-singer-udj3w0y
Protocol Definition: https://github.com/airbytehq/airbyte/blob/master/airbyte-protocol/models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml
And the parser is already in Java: https://github.com/airbytehq/airbyte/blob/master/airbyte-workers/src/main/java/io/airbyte/workers/protocols/airbyte/DefaultAirbyteStreamFactory.java

PR #385, opened 2021-08-24
implement DataObjects with state (#365)

zzeekk:
As preparation for #365 this PR implements execution mode DataObjectStateIncrementalMode and trait CanCreateIncrementalOutput to be implemented by DataObjects.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
@@ -20,9 +20,9 @@ package io.smartdatalake.app
 
 import java.io.File
 import java.time.LocalDateTime
-
 import com.typesafe.config.Config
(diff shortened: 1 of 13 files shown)


zzeekk:
@hiltpold: i renamed method initIncrementalMode to setState as suggested by @pgruetter 

@pgruetter: you were right, setState has to be called in init-phase but also after every increment in postExec.

hiltpold (approved):

pgruetter (approved):
LGTM
Only thing I didn't see / test is what happens when you have a DataObject with a state but you don't specify a state file on the command line ?

zzeekk:
> LGTM
> Only thing I didn't see / test is what happens when you have a DataObject with a state but you don't specify a state file on the command line ?

It will be treated as a normal DataObject, no set/getState called.

pgruetter:
> It will be treated as a normal DataObject, no set/getState called.

OK, maybe as an improvement we can find a way to at least warn about that. But it's probably going to be hard.

PR #395, opened 2021-09-18
Prevent file names starting with . in WebserviceFileDataObject (crc f…

Geheiner:
…iles still have original name though)

In this fix, I simply remove the dot of the filenames generated by WebserviceFileDataObject if it is just at the start of the filename.
I noticed that the .crc file still stays as before. No Idea how to change that.
Here is how the files look like after the fix:
image

I tested the fix with the configuration provided in the issue. The atteched screenshot shows the resulting output filenames


zzeekk (approved):
Yes, that's the right place 👍 

crc-Files are hidden files, they must have a dot in front.

I want to write a test for the filename translation, but didn't succeed yet because of another thing which i dont understand yet. If you're in a hurry i can make a separate PR for the test and you can merge this one.

Thanks!

PR #398, opened 2021-10-01
evolving schema with DeltaLakeTableDataObject with SDLSaveMode.Append

kapag-x:
apply mergeSchema option instead of overwriteSchema when writing to DeltaLakeTableDataObject with SDLSaveMode.Append

allows to support schema evolution with DeltaLakeTableDataObject and SDLSaveMode.Append


zzeekk (approved):
I suggest to add option "mergeSchema" in the partitions.isEmpty if-branch. Like this Append with SchemaEvolution is supported as well here without a special condition for it.

You can merge this PR if this is ok for you.

pgruetter:
@kapag-x @zzeekk 
Just created an issue: https://github.com/smart-data-lake/smart-data-lake/issues/399
Will this also be solved with this PR ?

PR #405, opened 2021-10-27
Bugfix/fix clear skipped on ignore filter

zzeekk:

1. fix parsing spark event listener info if appName contains special characters
2. fix reading data frame from skipped SubFeed if filters are ignored

1. If no application name is given feedSel is used as replacement. feedSel might contain special characters which break the regexp for parsing event info.
2. If a SubFeed is skipped because there is no data to process, an empty DataFrame is created for this SubFeed as input for CustomSparkAction. There is a constellation where an input SubFeed is skipped but inputIdsToIgnoreFilter is set for this input. In that case a DataFrame with the all input DataObjects data should be created, as filters incl. isSkipped can be ignored.


pgruetter (approved):
LGTM

Issue #406, opened 2021-11-02
Upgrade to Spark 3.2


zzeekk:
**Is your feature request related to a problem? Please describe.**
Spark 3.2 is out - it has many updated dependencies which might be important for projects (avro 1.10, hadoop 3.3)

**Describe the solution you'd like**
Update Spark dependency and dependencyManagement in sdl-parent pom to spark version 3.2

pgruetter:
We need to discuss if and how we can handle different spark versions. In Databricks i.e. the latest LTS version only supports 3.1.2 and is supported until Sep 23, 2023. If we only support Spark 3.2, we force everyone to move to a newer Databricks runtime version which are usually only supported for 6 months.

PR #421, opened 2021-11-18
updates for Spark3.2 (#406)

zzeekk:
see #406


zzeekk:
Delta lake version 1.0.0 does not support Spark 3.2.0. We need to wait for 1.1.x which is on the way.
See also https://github.com/delta-io/delta/issues/733

pgruetter (approved):
LGTM

PR #423, opened 2021-11-22
Feature/state logs

kapag-x:
- adapted StateChangeLogger to log only for the action for which the notification was emitted.
- added new attribute changedActionId : Option[ActionId] to StateListener.notifyState:
     changedActionId = single action whose state changed, if applicable. (Final notification provides the final state of the whole DAG.)

- with previous version, log emits several times the same status for the same action


sdl-core/src/main/scala/io/smartdatalake/app/StateListener.scala
```diff
@@ -53,6 +54,10 @@ trait StateListener {
 
   /**
    * notifyState is called whenever an action is finished (succeeded or failed)
+   *
+   * @param state of the currently active part of the DAG
+   * @param context information
+   * @param changedActionId : single action whose state changed, if applicable. (Final notification provides the final state of the whole DAG.)
```
zzeekk:
```suggestion
   * @param changedActionId : if notification is triggered by an action state changes, this is the action corresponding.
```

sdl-core/src/main/scala/io/smartdatalake/app/StateListener.scala
```diff
@@ -53,6 +54,10 @@ trait StateListener {
 
   /**
    * notifyState is called whenever an action is finished (succeeded or failed)
```
zzeekk:
```suggestion
   * notifyState is called whenever an action is finished (succeeded or failed) and at the end of the DAG execution (success or failure).
   * It always includes the state of all actions of the DAG.
   * At the end of the DAG execution notifyState is called with the final state. In this case state.isFinal = true and changedActionId is empty.
```

zzeekk (approved):

zzeekk (approved):

Issue #376, opened 2021-08-16
Support for Snowflake Snowpark


zzeekk:
**Is your feature request related to a problem? Please describe.**
With Snowpark Spark can be executed inside Snowflake (or a flavor of Spark?)

**Describe the solution you'd like**
Implement an Action to execute Snowpark Transformations.

**Additional context**
Documentation: https://docs.snowflake.com/en/developer-guide/snowpark/index.html
It seems that Snowflake copied or integrated large pieces of the Spark API. It must be analyzed if this could be used as Spark Version inside SDL or if it has to be handled as a separate Execution Engine. In that case there might be library problems to research...

MarkusRothSBB:
This would be very interesting for SBB, as we are using Snowflake. HistorizeAction currently leads to a large amount of reads to the target Snowflake table, making processing non-trivial amounts of data very slow. If HistorizeAction would work directly inside Snowpark, I expect this to work orders of magnitude faster.

MarkusRothSBB:
Update: According to Snowflake, Snowpark is now available von Azure. We could start experimenting on our infrastructure at SBB.

MarkusRothSBB:
I looked into this a little bit: https://www.phdata.io/blog/spark-developers-guide-to-snowpark/

It seems that Snowpark is a completely separate library to spark, it just uses a similar API. I guess we could have a separate execution path for actions that have only Snowflake data objects as inputs and outputs that use Snowpark DataFrames rather than Spark DataFrames, but it seems like it would be a major undertaking.

It would be really nice to be able to ingest data from source systems with spark and then do the transformations within Snowflake using Snowpark within the same project, though.

zzeekk:
I checked the ScalaDoc - as you said it "is a completely separate library to spark, it just uses a similar API".
https://docs.snowflake.com/en/developer-guide/snowpark/reference/scala/com/snowflake/snowpark/index.html
Source code is not available, i guess Snowflake doesn't need to open-source it as it's not a copy of Spark but just a similar API.
I think we could integrate it as another "execution engine", up to now we have "Spark DataFrames" and "File Input&OutputStream". Now we could add "Snowpark DataFrames".
It would be nice to reuse existing Actions (Copy/Deduplicate/HistorizeAction), but for that we would need to generalize the Spark and Snowpark API.
In a first step i propose to create a simple CopyAction for Snowpark, probably duplicating a lot of Code from SparkSubFeedAction.

[...]

PR #431, opened 2021-12-10
refactor integrating SparkSession into ActionPipelineContext

zzeekk:

preparation for #376 - include SparkSession into ActionPipelineContext, so that it is only created when needed.


zzeekk:
@MarkusRothSBB : i will check the build error later today...

PR #433, opened 2021-12-15
Added origin exception for better problem analysis

### LINKED ISSUE #432: throw exception needed in testSchemaRegistry
**Describe the bug**
Please show the Exception if the try of the function testSchemaRegistry of KafkaConnection has an error. 

Reproduce**
Actual definition of the function:

def testSchemaRegistry(): Unit = {
    try {
      confluentHelper.foreach(_.test())
    } catch {
      case e:Exception => throw ConfigurationException(s"($id) Can not connect to schema registry (${schemaRegistry.get})")
    }
  } 

**Expected behavior**
throw also the Exception

ursin:
Here a small fix to [issue 432](https://github.com/smart-data-lake/smart-data-lake/issues/432). Currently, the original exception is lost, when an attempt to connect to a schema registry fails.
The change in the pull request adds the original exception to the thrown *ConfigurationException*.
 
For a user, it is clear, that the connection to the registry failed. But the message of the *ConfigurationException* does not help in analyzing and solving the real problem. These can be 

- Wrong config
- Corrupt key/trust store
- ...

### DIFF
--- a/sdl-kafka/src/main/scala/io/smartdatalake/workflow/connection/KafkaConnection.scala
+++ b/sdl-kafka/src/main/scala/io/smartdatalake/workflow/connection/KafkaConnection.scala
@@ -93,7 +93,7 @@ case class KafkaConnection(override val id: ConnectionId,
     try {
       confluentHelper.foreach(_.test())
     } catch {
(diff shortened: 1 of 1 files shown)


pgruetter (approved):
Thanks, LGTM

PR #439, opened 2021-12-22
Feature/376: Add Snowpark Support

MarkusRothSBB:
First step in solving #376. 
This PR adds a CustomSnowparkDfsTransformer, that works exactly like a CustomDfsTransformer, but uses the Snowpark API for the transformation. It is a drop-in replacement for CustomDfsTransformer and can operate on the same SnowflakeTableDataObjects, so that you can write to a SnowflakeTableDataObject with a Spark action and then continue with a Snowpkark action in the same DAG.

To enable transformations directly within Snowflake, without have to transfer the data out to Spark and back in again.

### What are the limitations of this approach?
This is based on a drastically stripped down version of a SparkSubFeed. Savemodes and everything else interesting has been removed. HistorizeAction and DeduplicateAction are not supported. To enable this, either a lot of duplicate code would be necessary, or the DataFrame needs to be abstracted in sdl-core, which would be a major undertaking.

Even though this first version is very limited, it does not break any existing functionality (neither of sdl-core of sdl-snowflake), so I suggest merging it and gathering some real-life experience with it.


sdl-core/src/main/scala/io/smartdatalake/app/StateListener.scala
```diff
@@ -20,7 +20,6 @@
 package io.smartdatalake.app
 
 import io.smartdatalake.config.ConfigurationException
-import io.smartdatalake.util.misc.SmartDataLakeLogger
```
MarkusRothSBB, line 23: This was just an unused import

.gitignore
```diff
@@ -9,38 +9,38 @@ venv
 *.log
 
 # User-specific stuff
-.idea/**/workspace.xml
```
MarkusRothSBB, line 12: I'm note sure if newer versions of IDEA work differently (I'm still on 2020.1), but all the autogenerated files in .idea-directories of the module directories were annoying when moving between branches. I hope this change doesn't break anything.

pom.xml
```diff
@@ -356,7 +356,9 @@
 		<splunk.version>1.6.5.0</splunk.version>
 		<sshj.version>0.21.1</sshj.version>
 		<kafka.version>2.8.0</kafka.version>
-		<snowflake.version>2.8.4</snowflake.version>
+		<snowflake-spark.version>2.9.2</snowflake-spark.version>
```
MarkusRothSBB, line 359: Library that bridges Spark API with Snowflake.

sdl-snowflake/src/main/scala/io/smartdatalake/smartdatalake.scala
```diff
@@ -0,0 +1,31 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake
```
zzeekk, line 20: It's not clean if you define an object in module sdl-snowflake on the level of io.smartdatalake. We should move this to a package io.smartdatalake.utils.snowflake, or to the language implementation you are working on...

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/connection/SnowflakeConnection.scala
```diff
@@ -69,14 +71,31 @@ case class SnowflakeTableConnection(override val id: ConnectionId,
           "sfWarehouse" -> warehouse
         )
       case _ => throw new IllegalArgumentException(s"($id) No supported authMode given for Snowflake connection.")
-    } else throw new IllegalArgumentException(s"($id) No authMode given for Snowflake connection.")
+    }
   }
 
-  override def factory: FromConfigFactory[Connection] = SnowflakeTableConnection
+  def getSnowparkSession(schema: String): Session = {
```
zzeekk, line 78: It seems to me that the snowpark session is not yet shared, as every call to getSnowparkSession creates a new session. Also it seems that a snowpark session is schema dependent. You probably need to maintain a map of created sessions per schema and return already created sessions if existing...

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataobject/SnowflakeTableDataObject.scala
```diff
@@ -19,93 +19,85 @@
...
+  extends TransactionalSparkTableDataObject
+    with CanCreateSnowparkDataFrame
+    with CanWriteSnowparkDataFrame {
 
-  /**
-   * Connection defines connection string, credentials and database schema/name
-   */
-  private val connection = getConnection[SnowflakeTableConnection](connectionId)
+  private val connection = getConnection[SnowflakeConnection](connectionId)
+  private var _snowparkSession: Option[Session] = None
```
zzeekk: Ah, you create a session per DataObject. But it could be shared by connection & schema, right?
Maybe you should add the schema to the SnowflakeConnection?

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataobject/SnowflakeTableDataObject.scala
```diff
@@ -19,93 +19,85 @@
...
+    df.write
       .format(SNOWFLAKE_SOURCE_NAME)
       .options(connection.getSnowflakeOptions(table.db.get))
       .options(Map("dbtable" -> (connection.database + "." + table.fullName)))
       .mode(finalSaveMode.asSparkSaveMode)
       .save()
 
-    if (comment.isDefined) {
-      val sql = s"comment on table ${connection.database}.${table.fullName} is '${comment.get}';"
+    if (comment != null && comment.isDefined) {
```
zzeekk: It should not happen that an Option is null. I see that the default value is missing in the definition as case class attribute. Can you fix the default value and remove the null testing here?

zzeekk (commented):
I have some comments to check, but it looks great and still simple :-)
I see that you closed this PR and guess there will be a replacement soon...

MarkusRothSBB:
Thanks! As far as I can tell, you closed this PR :). Maybe we can reopen it?

I'll check out your comments for sure.

zzeekk:
Oh, it got closed because i deleted the target branch, as "feature/376-refactor-spark-session" was already merged to "develop-spark3". As one can not change the target branch name in an existing PR, can you create another PR with target branch "develop-spark3"?

zzeekk:
Yeah, was able to reopen the PR and change the target branch.
There are now some conflicts to resolve... we should rebase your branch onto develop-spark3 and also squash all commits into one. I could do this, but it changes the commit history and we need to sync on this, otherwise you'll have git problems. We can discuss this in our meeting tomorrow.

.gitignore
zzeekk: That's interesting, because i have only one .idea directory in the root directory, and i have various installations. Let's check this on your machine...

MarkusRothSBB: I upgraded to IntelliJ 2021.3. If I revert to the original state of .gitignore, I see the following:
image
 
It's not really relevant for this particular PR, so I can remove the change, but at least on my machine it makes working with the project a lot better.

sdl-snowflake/src/main/scala/io/smartdatalake/smartdatalake.scala
MarkusRothSBB: This is a [package object](https://docs.scala-lang.org/tour/package-objects.html) that contains package-wide type aliases for the io.smartdatalake package in the sdl-snowflake module, as described in the link. I think that the type aliases should be available in the entire package "io.smartdatalake" in the sdl-snowflake module, not inly in the "io.smartdatalake.utils.snowflake" package.

Of course, the type aliases are not strictly necessary, but I found that it makes working in classes with mixed Spark/Snowpark namespaces a lot easier.

Can you maybe elaborate a bit more on why you think this approach is not clean?

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/connection/SnowflakeConnection.scala
MarkusRothSBB: Every call to does create a new session, but the session is cached in SnowflakeTableDataObject._snowparkSession.

You need to pass a schema to create a Snowpark session, but you are not limited to that schema when using the session. It merely sets a default schema for the session. I always use fully qualified table names when interacting with Snowflake (via the Spark adapter or with Snowpark), so the default session is never used.

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataobject/SnowflakeTableDataObject.scala
MarkusRothSBB: You are right. I moved the cache down to the connection. A valid schema is needed to create the session, but after that, that schema is ignored, so there really only needs to be a single cached session per connection.

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataobject/SnowflakeTableDataObject.scala
MarkusRothSBB: Fixed. We had some trouble compiling the project due to generic values in the case classes extending ParsableFromConfig, so we did some experiments with removing generics. Forgot to revert this.

Issue #440, opened 2022-01-01
Implement Language server to provide code completion for configuration


zzeekk:
**Is your feature request related to a problem? Please describe.**
To find which properties an Action or DataObject supports, i find myself often checking scaladoc and code. It would be nice to have code completion for configuration files available in different IDE's like IntelliJ or VSCode.

**Describe the solution you'd like**
Code completion can be provided to IDE's by a language server implementation, see https://microsoft.github.io/language-server-protocol. A language server for SDL Hocon Files should be created. 
A first step in this direction has already been made: SDL's configuration schema is defined as json schema, see also http://smartdatalake.ch/JsonSchemaViewer.

In these issue a language server should be implemented, that takes SDL's configuration schema as input, and can suggest valid entries for a specific cursor position in a given configuration file. In order to lookup the valid configuration entries for a specific cursor position, the language server needs to parse Hocon Files keeping the position/origin of the parsed tokens. Interestlingly the standard Hocon Parser already supports this to a certain amount, but there seems to be a bug with endLineNumber.
- Tokenizer.tokenize returns an Iterator<Token>, each Token contains it's origin
- Parsable.rawParseDocument tokenizes an input and creates a SimpleConfigDocument.
- SimpleConfigDocument has ConfigNodeRoot configNodeTree, including all parsed tokens with their origin - but endLineNumber is not properly set
- ConfigParser then creates a SimpleConfig, including the origin for all entries - but endLineNumber is not properly set

A first step would therefore be to improve Hocon Parser to set endLineNumber correctly, and add start/endColumn or start/endPosition.

zzeekk:
Interesting Ressources:

Sample language server using LSP4J framework:
https://medium.com/ballerina-techblog/implementing-a-language-server-how-hard-can-it-be-part-1-introduction-c915d2437076
https://github.com/ballerina-platform/ballerina-lang/blob/master/language-server/modules/langserver-core
https://github.com/ballerina-platform/lsp4intellij

Intellij Hocon Plugin:
https://github.com/AVSystem/intellij-hocon

ANTLR
Hocon Grammar:
https://github.com/jdevelop/go-hocon/blob/master/parser/HOCON.g4
Code Completion:
https://github.com/mike-lischke/antlr4-c3
https://neuroning.com/post/implementing-code-completion-for-vscode-with-antlr/
Incremental Parsing:
https://github.com/dberlin/antlr4ts/blob/incremental/IncrementalParsing.md

[...]

PR #441, opened 2022-01-02
Export sdl configuration as json schema

zzeekk:
Export SDL configuration schema as json schema. This contributes to #440.

### DIFF
--- a/sdl-lang/src/main/scala/io/smartdatalake/meta/GenericTypeUtil.scala
+++ b/sdl-lang/src/main/scala/io/smartdatalake/meta/GenericTypeUtil.scala
@@ -1,0 +1,202 @@
+package io.smartdatalake.meta
+
+import com.github.takezoe.scaladoc.{Scaladoc => ScaladocAnnotation}
+import io.smartdatalake.config.InstanceRegistry
+import io.smartdatalake.util.misc.SmartDataLakeLogger
+import io.smartdatalake.workflow.action.{Action, ActionMetadata}
+import io.smartdatalake.workflow.connection.{Connection, ConnectionMetadata}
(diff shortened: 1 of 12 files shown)


sdl-lang/src/main/scala/io/smartdatalake/meta/GenericTypeUtil.scala
```diff
@@ -0,0 +1,202 @@
...
+ */
+private[smartdatalake] object GenericTypeUtil extends SmartDataLakeLogger {
+
+  def getReflections = new Reflections("io.smartdatalake")
+
+  /**
+   * Finds all relevant types according to the config and generates GenericTypeDefs for them.
+   * A final GenericTypeDef is marked with isFinal=true and represents an SDL case class used in configuration.
+   * The final types include all attributes. Inherited attributes are marked with isOverride=true.
+   * GenericTypeDefs include hte hierarchy of interfaces implemented by a type. They are listed in superTypes.
```
pgruetter:
```suggestion
   * GenericTypeDefs include the hierarchy of interfaces implemented by a type. They are listed in superTypes.
```

pgruetter (approved):
One question: Is it correct, that only the 4 sections global, connections, dataObjects, actions are parsed? If you define an arbitrary section like "defaults" or something, they are not exported correctly?

zzeekk:
> One question: Is it correct, that only the 4 sections global, connections, dataObjects, actions are parsed? If you define an arbitrary section like "defaults" or something, they are not exported correctly?

Schema Exporter only exports the schema our the hocon-file must obey. You're right that this is only defined for the 4 sections global, connections, dataObjects, actions. And the schema allows for arbitrary other sections.

Issue #426, opened 2021-12-03
Extend OAuth Support


ursin:
**Is your feature request related to a problem? Please describe.**
OAuth authentication is currently only possible for keycloak protected webservices. Other services use different OAuth flows.

**Tasks:**
- Test current OAuth implementation for different OAuth protected webservices
- Verify if the current implementation work or has to be extended

**Describe the solution you'd like**
The final solution has to be discussed. But the area of application makes currently only sense for webservices.

zzeekk:
implemented, see OAuthMode

PR #444, opened 2022-01-06
Feature/atlas exporter

zzeekk:
Adds Atlas metadata exporter implementation

### DIFF
--- a/sdl-lang/src/main/resources/atlasExporter.purview.conf
+++ b/sdl-lang/src/main/resources/atlasExporter.purview.conf
@@ -1,0 +1,40 @@
+
+atlasServer {
+  url: "https://xyz.purview.azure.com/catalog/"
+  auth {
+    token: "CLEAR#xyz"
+  }
+}
+
+appName: "develop"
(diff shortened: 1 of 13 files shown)


sdl-lang/src/main/resources/atlasExporter.purview.conf
```diff
@@ -0,0 +1,40 @@
+
+atlasServer {
+  url: "https://xyz.purview.azure.com/catalog/"
+  auth {
+    token: "CLEAR#xyz"
```
pgruetter, line 5: If we want to use Purview, we should implement the proper OAuth for authentication. Using a token is just for testing purposes as you need to get the token manually first.

pgruetter (approved):
Thanks for the integration!

sdl-lang/src/main/resources/atlasExporter.purview.conf
zzeekk: Ok, could be a topic in issue #426.

PR #446, opened 2022-01-09
Improvements for getting-started Part 3

zzeekk:


pgruetter (approved):
LGTM

hiltpold (approved):

sdl-deltalake/src/main/scala/io/smartdatalake/workflow/dataobject/DeltaLakeTableDataObject.scala
```diff
@@ -256,7 +256,7 @@ case class DeltaLakeTableDataObject(override val id: DataObjectId,
             dfWriter
               .mode(finalSaveMode.asSparkSaveMode)
               .option("mergeSchema", allowSchemaEvolution)
-              .saveAsTable(table.fullName)
+              .save() // it seems generally more stable to work without Table API
```
Geheiner, line 259: @zzeekk  Have you tested that this still correctly updates the hive Metastore? I'm wondering how it knows the table name now

zzeekk: @Geheiner: It's ok as we only use this if the table is already existing. Also the hadoopPath we use is checked against the hadoopPath of the metastore for this table. If there are differences the hadoopPath of the metastore is used and a warning logged.

PR #461, opened 2022-02-01
StatusInfo REST-API-Server

Geheiner:
### What changes are included in the pull request?
https://github.com/smart-data-lake/smart-data-lake/issues/450

Simple Rest-Server that starts along with SDLB if you define in your globalconfig the following :
global {
  statusInfo {
  }
}
See StatusInfoRestApiConfig for config options

It serves the content of ActionDAGRunState under 
http://localhost:4440/api/v1/state
and ActionPipelineContext under :
http://localhost:4440/api/v1/context

PR #471, opened 2022-02-22
Websocket for live status + DagExporter

### LINKED ISSUE #450: Embedded Server to get Live Status Info
Currently there is no convenient way to get specific information about the current state of SDL, including the DAG.
The only workaround for now is to parse the log files, which is tedious and might not contain the information you are looking for. 

A simple webserver that allows to query the current state of SDL and the DAG.
In a first version, it will communicate via REST.
In a second version, it will communicate via a websocket.

**Additional context**

Geheiner:
Websocket API that allows communication of updates on the Execution DAG.
DagExporter: CommandLineTool that Takes as input a SDL Config and prints it's containing Actions to STDOUT, in a simplified JSON-Format. 

These changes also include the unmerged, but approved changes https://github.com/smart-data-lake/smart-data-lake/pull/461 as they depend on each other


sdl-core/src/main/scala/io/smartdatalake/statusinfo/websocket/IncrementalStatusInfoListener.scala
```diff
@@ -0,0 +1,56 @@
...
+
+    val updateJSON =
+      if (changedActionId.isDefined) {
+        val changedActions = state.actionsState.filter(_._1 == changedActionId.get)
+
+        if (changedActions.size != 1) {
+          logger.warn(s"Not exactly one changedAction! Got: $changedActions")
+        }
+
+        val changedAction = changedActions.head
```
zzeekk, line 47: What if changedActions is empty? We should throw IllegalArgumentException in that case...

sdl-core/src/main/scala/io/smartdatalake/statusinfo/websocket/IncrementalStatusInfoListener.scala
```diff
@@ -0,0 +1,56 @@
...
+        SDLMessage(SDLMessageType.StatusUpdate, Some(StatusUpdate(Some(changedAction._1.id), Some(changedAction._2), context.phase, state.finalState)))
+      }
+      else {
+        SDLMessage(SDLMessageType.EndConnection, Some(StatusUpdate(None, None, context.phase, state.finalState)))
+      }
+
+    activeSockets.foreach(socket => socket.getRemote.sendString(writePretty(updateJSON)(ActionDAGRunState.formats + new EnumNameSerializer(SDLMessageType) + new EnumNameSerializer(ExecutionPhase))))
```
zzeekk: I suggest to move the following into a private instance val, maybe even an implicit val?
ActionDAGRunState.formats + new EnumNameSerializer(SDLMessageType) + new EnumNameSerializer(ExecutionPhase)

sdl-core/src/main/scala/io/smartdatalake/statusinfo/websocket/SDLMessage.scala
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.statusinfo.websocket
+
+import io.smartdatalake.statusinfo.websocket.SDLMessageType.SDLMessageType
+
+case class SDLMessage(msgType: SDLMessageType, statusUpdate: Option[StatusUpdate] = None, log: Option[ActionLog] = None)
```
zzeekk, line 31: Can we add a comment that either statustUpdate or log is filled, depending on the messages type?

sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
```diff
@@ -72,10 +90,11 @@ private[smartdatalake] object ActionDAGRunState {
     {case obj: DataObjectId => obj.id}
   ))
 
-  implicit private lazy val workflowReflections = ReflectionUtil.getReflections("io.smartdatalake.workflow")
+  implicit private lazy val workflowReflections: Reflections = ReflectionUtil.getReflections("io.smartdatalake.workflow")
 
   private lazy val typeHints = ShortTypeHints(ReflectionUtil.getTraitImplClasses[SubFeed].toList ++ ReflectionUtil.getSealedTraitImplClasses[ExecutionId], "type")
-  implicit private val formats: Formats = Json4sCompat.getStrictSerializationFormat(typeHints) + new EnumNameSerializer(RuntimeEventState) +
+  implicit val formats: Formats = Serialization.formats(typeHints)
+    .withStrictArrayExtraction.withStrictMapExtraction.withStrictOptionParsing + new EnumNameSerializer(RuntimeEventState) +
```
zzeekk: This seems to be a wrong merge conflict resultion. We need to use Json4sCompat.getStrictSerializationFormat to be compatible with older json4s versions of Spark 3.1.
```suggestion
    Json4sCompat.getStrictSerializationFormat(typeHints) + new EnumNameSerializer(RuntimeEventState) +
```

zzeekk (approved):
Nice!
I have some minor comments and there is one wrong merge conflict resolution to fix. Otherwise fine for me 👍

sdl-core/src/main/scala/io/smartdatalake/statusinfo/websocket/IncrementalStatusInfoListener.scala
Geheiner: Two lines above I make a check about that. I was not sure if this should really abort the entire SBLB execution (which would happen if I throw an exception) so I opted for a warn. But I can change that to an Exception

PR #475, opened 2022-02-25
typo download typ

mand35:
in that present stage the download-departures actions is still a of typ download. Just to stay consistent and prevent readers confusion.


pgruetter:
@Geheiner Could you please double-check if this proposal makes sense? Thanks

Geheiner (approved):

PR #476, opened 2022-02-25
part3: Note about data scheme

mand35:
A comment is added about the predefined json data scheme used in the new implementation.

### DIFF
--- a/docs/getting-started/part-3/custom-webservice.md
+++ b/docs/getting-started/part-3/custom-webservice.md
@@ -200,7 +200,7 @@ This logic is implemented in the next code snipped and should replace the code c
 }
 
 ```
-Note, in the *Init* case, the pre-defined **schema** (see the `ext-departures` dataObject definition in the config) is used for the creation of an empty/dummy json file, which then gets converted to the DataFrame. 
+Note, in the *Init* phase, the pre-defined **schema** (see the `ext-departures` dataObject definition in the config) is used to create an empty/dummy json file, which then gets converted to the DataFrame.
 
 Don't be confused about some comments in the code. They will be used in the next chapter. 
 If you re-compile the code of this project and then restart the program with the previous commands
--- a/docs/getting-started/part-3/custom-webservice.md
+++ b/docs/getting-started/part-3/custom-webservice.md
@@ -200,7 +200,7 @@ This logic is implemented in the next code snipped and should replace the code c
 }
 
 ```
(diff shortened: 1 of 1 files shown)


docs/getting-started/part-3/custom-webservice.md
```diff
@@ -200,6 +200,8 @@ This logic is implemented in the next code snipped and should replace the code c
 }
 
 ```
+Note, in the *Init* case, the pre-defined **schema** (see the `ext-departures` dataObject definition in the config) is used for the creation of an empty/dummy json file, which then gets converted to the DataFrame.
```
pgruetter:
```suggestion
Note, in the *Init* phase, the pre-defined **schema** (see the `ext-departures` dataObject definition in the config) is used to create an empty/dummy json file, which then gets converted to the DataFrame.
```

pgruetter: @zzeekk I think this is not strictly true. We don't create an empty Json file, we generate a DataFrame on the fly with the given schema, correct?

pgruetter (changes requested):

docs/getting-started/part-3/custom-webservice.md
mand35: I am happy if we can describe it in a correct way. I only want to point out that the defined schema comes in place here. This is the important bit for me. 
Thanks for finding to correct description of the present command. 
```Scala
Seq[String]().toDF("responseString")
.select(from_json($"responseString", DataType.fromDDL(schema)).as("response"))                                                                                                                                                            
.select(explode($"response").as("record"))                                                                                                                                                                                                
.select("record.*")                                                                                                                                                                                                                       
.withColumn("created_at", current_timestamp()) 
```

docs/getting-started/part-3/custom-webservice.md
```diff
@@ -200,6 +200,8 @@ This logic is implemented in the next code snipped and should replace the code c
 }
 
 ```
+Note, in the *Init* phase, the pre-defined **schema** (see the `ext-departures` dataObject definition in the config) is used to create an empty/dummy json file, which then gets converted to the DataFrame.
```
Geheiner:
```suggestion
Note, in the *Init* phase, the pre-defined **schema** (see the `ext-departures` dataObject definition in the config) is used to create an empty/dummy json string, which then gets converted to the DataFrame.
```

Geheiner (changes requested):

PR #477, opened 2022-02-25
part3 typo .get

mand35:
The `DepartureQueryParameters` do not have a method `get`. 
The solutions file is correct, but the code snipped is not. Corrected now.

Call to `DepartureQueryParameters` is corrected, now without `get`

otherwise the compiler throws an error


pgruetter (approved):
LGTM, @Geheiner can you confirm?

PR #486, opened 2022-03-15
implement GenericDataFrame (#376)

zzeekk:
Proposal for a GenericDataFrame with implementations for Spark and Snowpark.

It allows to write generic transformations by using the GenericDataFrame API, but it's also possible to write transformations directly with Spark or Snowpark. In prepare phase supported SubFeedTypes are collected from input, output and transformations, and then decided which one two use.

This is tested with Spark (all unit tests) and Snowpark (read/write/create tables, CopyAction with generic transformations) and there also a data pipeline combining Spark and Snowpark incl. schema convertion (see io.smartdatalake.workflow.snowflake.SparkAndSnowparkDataPipelineIT).

Breaking changes in config and transformation implementation:
- removal of deprecated static transformation configuration in CopyAction, DeduplicateAction and HistorizeAction, e.g. columnBlacklist, columnWhitelist, additionalColumns, filterClause, standardizeDatatypes
- renaming of CustomSparkAction -> CustomDataFrameAction
- renaming of SparkIncrementalMode -> DataFrameIncrementalMode
- renaming of pacakge io.smartdatalake.workflow.spark.customlogic -> io.smartdatalake.workflow.customlogic
- renaming of transformers:
  - PythonCodeDfTransformer -> PythonCodeSparkDfTransformer
  - ScalaClassDfsTransformer -> ScalaClassSparkDfsTransformer
  - ScalaClassDfTransformer -> ScalaClassSparkDfTransformer
  - ScalaCodeDfsTransformer -> ScalaCodeSparkDfsTransformer
  - ScalaCodeDfTransformer -> ScalaCodeSparkDfTransformer
  - ScalaNotebookTransformer -> ScalaNotebookSparkDfTransformer
  - RepartitionTransformer -> SparkRepartitionTransformer
  - StandardizeDatatypesTransformer -> StandardizeSparkDatatypesTransformer

Because of the breaking changes and the major refactoring i increased version number to 2.3.0-SNAPSHOT.

Internal changes (not final...):
- schema & schemaMin attribute of DataObjects is now a Option[GenericSchema]
- DataObjects get/writeDataFrame method became get/writeSparkDataFrame, and there is a new get/writeDataFrame method working with GenericDataFrames.
- Generic objects have a variable subFeedType(s) which declares the implementations it supports, e.g. SparkSubFeed, SnowparkSubFeed or DataFrameSubFeed if it implements / can handle all implementations of GenericDataFrames. 
- An implementation of DataFrameSubFeed (e.g. SparkSubFeed) must implement a companion object DataFrameSubFeedCompanion with helper methods implementations. The helper object can be instantiated dynamically by using DataFrameSubFeed.getCompanion(subFeedType).

Open points / Limitations: 
- DeduplicateAction and HistorizeAction dont support yet processing SnowparkDataFrames. This will be another PR.

@Geheiner: you promised to make a review ;-)
@pgruetter, @MarkusRothSBB: your feedback in general would be appreciated


sdl-core/src/main/scala/io/smartdatalake/definitions/SDLSaveMode.scala
```diff
@@ -50,6 +50,7 @@ object SDLSaveMode extends Enumeration {
   val Ignore: Value = Value("Ignore")
 
   /**
+   * Spark only optimization.
```
Geheiner, line 53: would be nice to enforce that somewhere in the code instead of just the in the doc

sdl-core/src/main/scala/io/smartdatalake/workflow/action/sparktransformer/StandardizeDatatypesTransformer.scala
```diff
@@ -23,25 +23,25 @@ import com.typesafe.config.Config
...
  * @param name         name of the transformer
  * @param description  Optional description of the transformer
  */
-case class StandardizeDatatypesTransformer(override val name: String = "standardizeDatatypes", override val description: Option[String] = None) extends ParsableDfTransformer {
+case class StandardizeDatatypesTransformer(override val name: String = "standardizeDatatypes", override val description: Option[String] = None) extends SparkDfTransformer {
   override def transform(actionId: ActionId, partitionValues: Seq[PartitionValues], df: DataFrame, dataObjectId: DataObjectId)(implicit context: ActionPipelineContext): DataFrame = {
     df.castAllDecimal2IntegralFloat
   }
-  override def factory: FromConfigFactory[ParsableDfTransformer] = StandardizeDatatypesTransformer
+  override def factory: FromConfigFactory[GenericDfTransformer] = StandardizeDatatypesTransformer
```
Geheiner: If this is supposed to be spark specific, why use the GenericDfTransformer?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
```diff
@@ -0,0 +1,290 @@
...
+  /**
+   * returns data frame which consists of those rows which contain at least a null in the specified columns
+   */
+  def getNulls(cols: Seq[String]): GenericDataFrame = {
+    import helper._
+    filter(cols.map(col(_).isNull).reduce(_ or _))
+  }
+
+  /**
+   * Count nlets of this data frame with respect to specified columns.
```
Geheiner: typo?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  /**
+   * compares df with df2
+   *
+   * @param other : data frame to comapre with
+   * @return true if both data frames have the same cardinality, schema and an empty symmetric difference
+   */
+  def isEqual(other: GenericDataFrame): Boolean = {
+    // As a set-theoretic function symmetricDifference ignores multiple occurences of the same row.
```
Geheiner, line 149: why do we ignore multiple occurences of the same row?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkDataFrame.scala
```diff
@@ -0,0 +1,269 @@
...
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Column, DataFrame, RelationalGroupedDataset, Row}
+
+import scala.reflect.runtime.universe
+import scala.reflect.runtime.universe.typeOf
+
+case class SparkDataFrame(inner: DataFrame) extends GenericDataFrame {
+  override def subFeedType: universe.Type = typeOf[SparkSubFeed]
+  override def schema: SparkSchema = SparkSchema(inner.schema)
+  override def join(other: GenericDataFrame, joinCols: Seq[String]): SparkDataFrame = {
```
Geheiner, line 39: so we only support inner joins? For more complex joins, CustomTransformers shall be written in the specific language from what I understood, right?

MarkusRothSBB:
This is phenomenal work! Looking fantastic, and might have a staggering impact on our work with Snowflake. @ursin offered to look into the changes and test them with our Snowflake environment. He will give you feedback directly.

sdl-core/src/main/scala/io/smartdatalake/workflow/SubFeed.scala
```diff
@@ -92,109 +89,101 @@ trait SubFeedConverter[S <: SubFeed] {
...
+  override def breakLineage(implicit context: ActionPipelineContext): DataFrameSubFeed
+  override def clearPartitionValues(breakLineageOnChange: Boolean = true)(implicit context: ActionPipelineContext): DataFrameSubFeed
+  override def updatePartitionValues(partitions: Seq[String], breakLineageOnChange: Boolean = true, newPartitionValues: Option[Seq[PartitionValues]] = None)(implicit context: ActionPipelineContext): DataFrameSubFeed
+  override def clearDAGStart(): DataFrameSubFeed
+  override def clearSkipped(): DataFrameSubFeed
+  def isStreaming: Option[Boolean]
+  def withDataFrame(dataFrame: Option[GenericDataFrame]): DataFrameSubFeed
+  def withPartitionValues(partitionValues: Seq[PartitionValues]): DataFrameSubFeed
+  def withFilter(partitionValues: Seq[PartitionValues], filter: Option[String]): DataFrameSubFeed
+  def applyFilter: DataFrameSubFeed = {
```
Geheiner: So this is how code looks like that is agnostic to Spark or Snowflake! Quite impressive!

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkSubFeed.scala
```diff
@@ -0,0 +1,243 @@
...
+  }
+  override def structType(fields: Map[String,GenericDataType]): GenericDataType = {
+    assert(fields.values.forall(_.isInstanceOf[SparkDataType]), s"Unsupported subFeedType(s) ${fields.values.filterNot(_.isInstanceOf[SparkDataType]).map(_.subFeedType.typeSymbol.name).toSeq.distinct.mkString(", ")} in method structType")
+    val sparkFields = fields.map{ case (name,dataType) => StructField(name, dataType.asInstanceOf[SparkDataType].inner)}.toSeq
+    SparkDataType(StructType(sparkFields))
+  }
+  /**
+   * Construct array from given columns removing null values (Snowpark API)
+   */
+  override def array_construct_compact(columns: GenericColumn*): GenericColumn = {
```
Geheiner, line 208: So here we you extended the Spark API with a function from the snowpark API! Interesting possibilities :)

sdl-snowflake/src/test/scala/io/smartdatalake/workflow/snowflake/SparkAndSnowparkDataPipelineIT.scala
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * This is an integration test with data pipeline combining Spark and Snowpark
+ * The first Action will write with Spark to Snowflake, creating a new table in Snowflake.
+ * The second Action will read the new Snowflake table with Snowpark and create another Snowflake table.
+ * The SubFeed between Action1 and Action2 will apply schema conversion which is needed in init phase, as the tables in Snowflake dont exist yet.
+ * Action2 will also apply two generic transformation, e.g. filter and add runId column.
+ * It needs to be run manually because you need to provide a Snowflake environment.
+ * Please configure this in SnowflakeConnectionConfig.
+ */
+object SparkAndSnowparkDataPipelineIT extends App {
```
Geheiner, line 48: From what I understand, the underlying execution engine is still spark even when you apply snowflake transformations, correct? Now sure how this works now  for executing snowpark transformations?
Everything seems to work with a spark session

sdl-core/src/main/scala/io/smartdatalake/workflow/action/sparktransformer/FilterTransformer.scala
```diff
@@ -35,19 +36,25 @@ import scala.util.{Failure, Success, Try}
...
  * @param description  Optional description of the transformer
  * @param filterClause Spark SQL expression to filter the DataFrame
+ * @param subFeedTypeForValidation When parsing the configuration the runtime subFeedType for validating the filter expression is not yet known.
+ *                                 By default SparkSubFeed langauge is used, but you can configure a different one if needed.
  */
-case class FilterTransformer(override val name: String = "filter", override val description: Option[String] = None, filterClause: String) extends ParsableDfTransformer {
+case class FilterTransformer(override val name: String = "filter", override val description: Option[String] = None, filterClause: String, subFeedTypeForValidation: String = typeOf[SparkSubFeed].typeSymbol.fullName) extends GenericDfTransformer {
+  private val validationHelper: DataFrameSubFeedCompanion = DataFrameSubFeed.getHelper(subFeedTypeForValidation)
+  import validationHelper._
   private val filterClauseExpr = Try(expr(filterClause)) match {
```
Geheiner, line 45: you can remove the unused val now

sdl-snowflake/src/test/scala/io/smartdatalake/workflow/snowflake/SparkAndSnowparkDataPipelineIT.scala
Geheiner: I was searching for a while for this test until I found it in this module :D
I know this may seem a bit far-fetched now but: If we start to support both Spark and Snowflake, it seems a bit weird to have all Spark-Specific Stuff in the core module  but all Snowflake specific stuff in a separate module.
I know for obvious historical reasons Spark is everywhere in SDL right now. Would it make sense to also move Spark-Specific stuff to a specific module sdl-spark in the future?

Geheiner (commented):
Here is my first batch of comments.
I agree with my predecessors, amazing work that opens up amazing possibilities. Looking forward to the discussions.

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkDataFrame.scala
```diff
@@ -0,0 +1,269 @@
...
+
+import scala.reflect.runtime.universe
+import scala.reflect.runtime.universe.{Type, typeOf}
+
+case class SnowparkDataFrame(inner: DataFrame) extends GenericDataFrame {
+  override def subFeedType: universe.Type = typeOf[SnowparkSubFeed]
+  override def schema: SnowparkSchema = SnowparkSchema(inner.schema)
+  override def join(other: GenericDataFrame, joinCols: Seq[String]): SnowparkDataFrame = {
+    other match {
+      case sparkOther: SnowparkDataFrame => SnowparkDataFrame(inner.join(sparkOther.inner, joinCols))
```
Geheiner: Some remaining Copy-Paste naming issues.
Replace all 36 occurences of spark with snowpark

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkDataFrame.scala
```diff
@@ -0,0 +1,269 @@
...
+      case _ => throw new IllegalStateException(s"Unsupported subFeedType ${other.subFeedType.typeSymbol.name} in method join")
+    }
+  }
+  override def select(columns: Seq[GenericColumn]): SnowparkDataFrame = {
+    assert(columns.forall(_.subFeedType =:= subFeedType), s"Unsupported subFeedType(s) ${columns.filter(c => !(c.subFeedType =:= subFeedType)).map(_.subFeedType.typeSymbol.name).toSet.mkString(", ")} in method select")
+    SnowparkDataFrame(inner.select(columns.map(_.asInstanceOf[SnowparkColumn].inner)))
+  }
+  override def groupBy(columns: Seq[GenericColumn]): SnowparkGroupedDataFrame = {
+    assert(columns.forall(_.subFeedType =:= subFeedType), s"Unsupported subFeedType(s) ${columns.filter(c => !(c.subFeedType =:= subFeedType)).map(_.subFeedType.typeSymbol.name).toSet.mkString(", ")} in method select")
```
Geheiner: 1. There are typos for these asserts: here it should say "in method groupBy"
2. I have seen many similar definitions of these asserts. I think we should extract some common methods from them and call this generic method. If you search for "columns.filter(c => !(c.subFeedType =:= subFeedType)" you will find 4 occurences in SnowparkDataFrame and 4 occurences in SparkDataFrame. We could define this in GenericDataFrame

sdl-core/src/main/scala/io/smartdatalake/workflow/action/sparktransformer/StandardizeDatatypesTransformer.scala
zzeekk: Good question. StandardizeDatatypesTransformer extends SparkDfTransformer. But here we give the generic supertype for extraction from hocon config.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
zzeekk: This is code moved during refactoring... but you're right, it should be n-lets. I'll fix that.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
zzeekk: Well, in a set multiple occurences of the same object are ignored.
As symmetricDifference is a set theoretic function, it's implemented like that.
But isEqual should not ignore multiple occurences of the same object are ignored. Thats why there is an additional condition on the count (you can ask @kaelte about details of set theory).
Note that this is also code moved during refactoring, and not new...

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkSubFeed.scala
zzeekk: Yes, that was a thing we need and i implemented this in Spark with flatten.
Interestingly such a function exists in Snowpark...

sdl-snowflake/src/test/scala/io/smartdatalake/workflow/snowflake/SparkAndSnowparkDataPipelineIT.scala
zzeekk: No. The Spark Session is only started when an Action executes with SubFeedType=SparkSubFeed.
SubFeedType is evaluated at runtime based on supported SubFeedTypes of inputs, outputs and transformations.
The second copy Action selected SubFeedType=SnowparkSubFeed and is executed with Snowpark, see corresponding Log about SubFeedType.
You're right, we could create an sdl-spark module in the future. In sdl-core we need Hadoop for file access and spark-core to evaluate expressions (for example executionCondition).

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkDataFrame.scala
zzeekk: We can extend this interface in the future. That's just the baseline we need for features currently implemented in SDL. A leftJoin would make sense, for sure.
On the other side i'm not yet sure if we should push writing generic transformations.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
Geheiner: What is a n-let?

zzeekk: You might know a duplet, right? From @kaelte i learned that there is also a onelet, threelet, fourlet....

kaelte: A row of a dataframe df is a n-let
iff df contains n-1 other rows which have the values.
A 2-let is also know as _doublet_, 3-let as _triplet_.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
kaelte: `symmetricDifference` is based on [DataSet.except](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#except(other:org.apache.spark.sql.Dataset[T]):org.apache.spark.sql.Dataset[T]) which removes n-lets.

Since Spark 2.4.0 there exists [DataSet.exceptAll](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#exceptAll(other:org.apache.spark.sql.Dataset[T]):org.apache.spark.sql.Dataset[T]) which _preserves the duplicates_ (even if they occure more than twice).

If the definition of `symmetricDifference` is changed to using [DataSet.exceptAll](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#exceptAll(other:org.apache.spark.sql.Dataset[T]):org.apache.spark.sql.Dataset[T]) then `isequal` will not need to compare anymore the cardinalities.

In Mathematics, a set contains only one instance of each element, in contrast to a [multiset](https://en.wikipedia.org/wiki/Multiset). A Spark Dataset is actually a Datamultiset.

sdl-core/src/main/scala/io/smartdatalake/config/ConfigToolbox.scala
```diff
@@ -59,7 +59,7 @@ object ConfigToolbox {
     val name = "interactive"
     val globalConfig = GlobalConfig()
     val context = ActionPipelineContext(name, name, SDLExecutionId.executionId1, instanceRegistry, Some(LocalDateTime.now()), SmartDataLakeBuilderConfig(name, Some(name)), phase = ExecutionPhase.Exec, serializableHadoopConf = defaultHadoopConf, globalConfig = globalConfig)
-    globalConfig._sparkSession = Some(sparkSession)
+    globalConfig._sparkSession = Option(sparkSession)
```
kaelte, line 62: Why is Option better than Some?

zzeekk: If sparkSession=null then Option returns None, else Some(sparkSession)
Background: getDefaultActionPipelineContext is used in different context for unit tests and interactive tools. Sometimes you want to use an existing SparkSession, sometimes you want to work without SparkSession (e.g. create it on-demand).

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkDataFrame.scala
zzeekk: done

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
Geheiner: OK, thanks! It's one of these words that's not easy to google ;)

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkSubFeed.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+  override def stringType: SnowparkDataType = SnowparkDataType(StringType)
+  override def arrayType(dataType: GenericDataType): SnowparkDataType = {
+    dataType match {
+      case snowparkDataType: SnowparkDataType => SnowparkDataType(ArrayType(snowparkDataType.inner))
+      case _ => throw new IllegalStateException(s"Can not create SnowparkDataType for ${dataType.subFeedType.typeSymbol.name}")
+    }
+  }
+  override def structType(fields: Map[String,GenericDataType]): SnowparkDataType = {
+    assert(fields.values.forall(_.isInstanceOf[SnowparkDataType]), s"Unsupported subFeedType(s) ${fields.values.filterNot(_.isInstanceOf[SparkDataType]).map(_.subFeedType.typeSymbol.name).toSeq.distinct.mkString(", ")} in method structType")
```
Geheiner: similar comment here as for
https://github.com/smart-data-lake/smart-data-lake/pull/486#discussion_r832001789
I have seen many similar definitions of these asserts. I think we should extract some common methods from them and call this generic method. If you search for "Unsupported subFeedType(s) " you will find 
 21 occurences in SnowparkDataFrame and in SparkDataFrame. We could define these in GenericDataFrame, would be less error prone for wrong method names in the error message.

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkSubFeed.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+  override def count(column: GenericColumn): SnowparkColumn = {
+    column match {
+      case snowparkColumn: SnowparkColumn => SnowparkColumn(functions.count(snowparkColumn.inner))
+      case _ => throw new IllegalStateException(s"Unsupported subFeedType ${column.subFeedType.typeSymbol.name} in method count")
+    }
+  }
+  override def explode(column: GenericColumn): SnowparkColumn = {
+    // TODO: check if this can be done by a udf?
+    throw new NotImplementedError("explode array is not implemented in Snowpark")
```
Geheiner, line 162: I found this implementation in snowflake DataFrame API:
com.snowflake.snowpark.DataFrame#flatten
https://docs.snowflake.com/en/sql-reference/functions/flatten.html
I think it does the same thing, but it needs a DataFrame...

zzeekk: You're right, the function exists on DataFrame-level and not on Column-level. In my opinion this is due the fact, that Snowflake treats nested datatypes as variant in SQL. And as Snowpark gets translated to SQL, it's difficult to implement a flatten function on column level, which would need to parse the variant and apply flatten at some level, as the schema is not known at plan-time, but only at execution-time...
I think Spark is more sophisticated with nested data types than Snowpark. Spark knows the schema at plan-time, but in a Snowpark UDF its schema-on-read, means you get a variant and try to cast it to what ever you want:
https://docs.snowflake.com/en/developer-guide/snowpark/reference/scala/com/snowflake/snowpark/types/Variant.html
This also means there are no nested Structs, but only nested Maps...

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkDataFrame.scala
Geheiner: One additional use-case of generic transformations that comes to mind is Unit-Testing:
As Snowflake does not allow you to do proper Unit-Tests with connecting to the cloud, using generic functions would be a sort of workaround to allow you to to switch execution engines for Unit-Testing. 
So for that, the user would need to replace the SnowflakeTableDataObject that he uses in production with for example HiveTableDataObjects in the unit-tests and therefore SDLB would automatically switch to Spark during Unit-Tests.
Not sure how realistic this is given that having a different engine for your tests may make the code behave differently for tests... It might be worth a shot to try it out when migrating a SDL project from Spark to Snowpark.
For SDL-Internal transformations we already do that implicitly by using Spark as Engine all the time when there is no SnowflakeTableDataObject involved.

sdl-core/src/main/scala/io/smartdatalake/definitions/ExecutionMode.scala
```diff
@@ -303,6 +301,7 @@ case class SparkStreamingMode(checkpointLocation: String, triggerType: String =
  * @param applyCondition Condition to decide if execution mode should be applied or not. Define a spark sql expression working with attributes of [[DefaultExecutionModeExpressionData]] returning a boolean.
  *                       Default is to apply the execution mode if given partition values (partition values from command line or passed from previous action) are not empty.
  */
+// TODO: this is not Spark specific and should be renamed to DataFrameIncrementalMode!
```
Geheiner: as discussed, we will proceed with the renaming within this PR

sdl-core/src/main/scala/io/smartdatalake/workflow/action/CustomSparkAction.scala
```diff
@@ -45,11 +47,12 @@ import org.apache.spark.sql.SparkSession
  * @param recursiveInputIds      output of action that are used as input in the same action
  * @param inputIdsToIgnoreFilter optional list of input ids to ignore filter (partition values & filter clause)
  */
+// TODO: should be renamed to CustomDataFrameAction
```
Geheiner: as discussed, we will proceed with the renaming within this PR

Geheiner:
I noticed that SplunkDataObject is the only place that explicitly uses spark.implicits._, and that in one small substep.
So I guess it will only work with Spark because of that. Might be not so complicated to generify it, though I'm not sure if it needs to be a priority. Does it make sense to open a ticket for that?
There are probably others that don't need to be Spark Specific per se but that still are. We'll do them when needed, i guess

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/GenericDfsTransformerDef.scala
```diff
@@ -0,0 +1,151 @@
...
+  private[smartdatalake] def applyTransformation(actionId: ActionId, partitionValues: Seq[PartitionValues], dfs: Map[String,GenericDataFrame])(implicit context: ActionPipelineContext): (Map[String,GenericDataFrame], Seq[PartitionValues]) = {
+    val transformedDfs = transform(actionId, partitionValues, dfs)
+    val transformedPartitionValues = transformPartitionValues(actionId, partitionValues).map(_.values.toSeq.distinct)
+      .getOrElse(partitionValues)
+    (transformedDfs,transformedPartitionValues)
+  }
+}
+
+/**
+ * Interface to implement Spark-DataFrame transformers working with many inputs and many outputs (n:m)
```
Geheiner: I guess this comment is meant for SparkDfsTransformer and not GenericDfsTransformer?

sdl-core/src/main/scala/io/smartdatalake/workflow/SubFeed.scala
```diff
@@ -92,109 +89,102 @@ trait SubFeedConverter[S <: SubFeed] {
...
+    dataObject match {
+      case input: UserDefinedSchema if input.schema.isDefined =>
+        input.schema.map(dataObject.createReadSchema)
+      case input: SchemaValidation if input.schemaMin.isDefined =>
+        input.schemaMin.map(dataObject.createReadSchema)
+      case _ => None
     }
   }
+
+  /**
```
Geheiner: This block of functions can be put into another seperate trait GenericColumnFunctions and then you can implement it in Objects called SparkFunctions and SnowparkFunctions. Would be easier to follow.
Right now the trait DataFrameSubFeed a bit of a SwissArmyKnife that does more things than just "A SubFeed transports references to data between Actions.". If I call .sql and get a GenericColumn, i'm in some middle step within an action, so it does not fit the definition

sdl-core/src/main/scala/io/smartdatalake/workflow/SubFeed.scala
```diff
@@ -92,109 +89,102 @@ trait SubFeedConverter[S <: SubFeed] {
...
+  def arrayType(dataType: GenericDataType): GenericDataType
+  def structType(colTypes: Map[String,GenericDataType]): GenericDataType
+  /**
+   * Get a DataFrame with the result of the given sql statement.
+   * @param dataObjectId Snowpark implementation needs to get the Snowpark-Session from the DataObject. This should not be used otherwise.
+   */
+  def sql(query: String, dataObjectId: DataObjectId)(implicit context: ActionPipelineContext): GenericDataFrame
+}
+object DataFrameSubFeed {
+  def getHelper(tpe: Type): DataFrameSubFeedCompanion = ScalaUtil.companionOf[DataFrameSubFeedCompanion](tpe)
```
Geheiner: From what I have seen, getHelper is only used to be able to use spark/snowparkfunctions. This indicates to me that the split suggested in the previous comment makes sense. Then the Object can be renamed DataFrameFunctions and getHelper can be renamend getFunctions

zzeekk:
> I noticed that SplunkDataObject is the only place that explicitly uses spark.implicits._, and that in one small substep. So I guess it will only work with Spark because of that. Might be not so complicated to generify it, though I'm not sure if it needs to be a priority. Does it make sense to open a ticket for that? There are probably others that don't need to be Spark Specific per se but that still are. We'll do them when needed, i guess

As discussed, one can not create connectors for Snowflake/Snowpark. SplunkDataObject is implemented for Spark, it returns a Spark DataFrame, and we can use Spark to load data from Splunk into Snowflake if needed.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/GenericDfsTransformerDef.scala
zzeekk: Done

sdl-core/src/main/scala/io/smartdatalake/workflow/action/CustomSparkAction.scala
zzeekk: Done

sdl-core/src/main/scala/io/smartdatalake/definitions/ExecutionMode.scala
zzeekk: Done

sdl-core/src/main/scala/io/smartdatalake/workflow/SubFeed.scala
zzeekk: Done, created trait DataFrameFunctions

sdl-core/src/main/scala/io/smartdatalake/workflow/SubFeed.scala
zzeekk: Done, we have now a getCompanion and a getFunctions method

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkDataFrame.scala
zzeekk: Ok, created a helper method in DataFrameSubFeed

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataframe/snowflake/SnowparkSubFeed.scala
zzeekk: Ok, created a helper method in DataFrameSubFeed

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkSubFeed.scala
```diff
@@ -0,0 +1,244 @@
...
+    this.dataFrame.foreach(_.inner.persist) // Spark's persist & cache can be called without referencing the resulting DataFrame
+    this
+  }
+  override def unpersist: SparkSubFeed = {
+    this.dataFrame.foreach(_.inner.unpersist) // Spark's unpersist can be called without referencing the resulting DataFrame
+    this
+  }
+  override def isStreaming: Option[Boolean] = dataFrame.map(_.inner.isStreaming)
+  override def hasReusableDataFrame: Boolean = dataFrame.isDefined && !isDummy && !isStreaming.getOrElse(false)
+  // TODO: still needed?
```
Geheiner: can we remove this unused method? I did not find any usages

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataobject/SnowflakeTableDataObject.scala
```diff
@@ -45,31 +51,35 @@ import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
...
+                                    override val schemaMin: Option[GenericSchema] = None,
                                     saveMode: SDLSaveMode = SDLSaveMode.Overwrite,
                                     connectionId: ConnectionId,
                                     comment: Option[String] = None,
                                     override val metadata: Option[DataObjectMetadata] = None)
                                    (@transient implicit val instanceRegistry: InstanceRegistry)
-  extends TransactionalSparkTableDataObject
-    with CanCreateSnowparkDataFrame
-    with CanWriteSnowparkDataFrame {
+  extends TransactionalSparkTableDataObject {
```
Geheiner, line 63: SnowflakeTableDataObject extends something Spark-Specific?... That sounds wrong

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/CanCreateSparkDataFrame.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import io.smartdatalake.workflow.dataframe.GenericDataFrame
+import io.smartdatalake.util.hdfs.PartitionValues
+import io.smartdatalake.workflow.{ActionPipelineContext, DataFrameSubFeed}
+import io.smartdatalake.workflow.dataframe.spark.{SparkDataFrame, SparkSubFeed}
+import org.apache.spark.sql.{DataFrame, DataFrameReader, DataFrameWriter}
+
+import scala.reflect.runtime.universe.{Type, typeOf}
+
+private[smartdatalake] trait CanCreateSparkDataFrame { this: DataObject with CanCreateDataFrame => // CanCreateDataFrame is not extended because a DataObject can implement multiple CanCreate*DataFrame traits.
```
Geheiner: I don't get why it's a problem to have a DataObject extend multiple CanCreate*DataFrame traits. Is it because of conflicting implementations of methods?
I tried to reconstruct it with a simple code snippet, but did not get what's the problem:

    trait Base {
    def blub: Int
    }
    
    trait SubA extends Base{
    override def blub = 1
    }
    
    trait SubB extends Base {
    override def blub: Int = 2
    }
    
    case class SubSub() extends SubB with SubA {
    
    }
    val subsub1 = SubSub()
    println(subsub1.blub)

sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
```diff
@@ -309,8 +347,9 @@ private[smartdatalake] abstract class SparkActionImpl extends ActionSubFeedsImpl
   /**
    * apply transformer to SubFeeds
    */
-  protected def applyTransformers(transformers: Seq[DfsTransformer], inputPartitionValues: Seq[PartitionValues], inputSubFeeds: Seq[SparkSubFeed], outputSubFeeds: Seq[SparkSubFeed])(implicit context: ActionPipelineContext): Seq[SparkSubFeed] = {
+  protected def applyTransformers(transformers: Seq[GenericDfsTransformerDef], inputPartitionValues: Seq[PartitionValues], inputSubFeeds: Seq[DataFrameSubFeed], outputSubFeeds: Seq[DataFrameSubFeed])(implicit context: ActionPipelineContext): Seq[DataFrameSubFeed] = {
     val inputDfsMap = inputSubFeeds.map(subFeed => (subFeed.dataObjectId.id, subFeed.dataFrame.get)).toMap
+    // TODO !!!
```
Geheiner: what is there TODO ?

sdl-snowflake/src/main/scala/io/smartdatalake/workflow/dataobject/SnowflakeTableDataObject.scala
zzeekk: No that's correct. The SnowflakeTableDataObject can be used with Spark (Snowflake Spark Connector) and Snowflake (Snowpark).

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/spark/SparkSubFeed.scala
zzeekk: Ok, I'll remove it

sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
zzeekk: No changes against the same method in SparkActionImpl / develop-spark3. I'll remove the TODO.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/CanCreateSparkDataFrame.scala
zzeekk: You're right. There is no problem with inheritance. It was already improved in CanWriteSparkDataFrame, i changed this also in CanReadSparkDataFrame.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/Table.scala
```diff
@@ -21,10 +21,10 @@ package io.smartdatalake.workflow.dataobject
 /**
  * Table attributes
  *
- * @param db optional override of db defined by connection
- * @param name table name
- * @param query optional select query
- * @param primaryKey optional sequence of primary key columns
+ * @param db          optional override of database-schema defined by connection. Called db for backwards-compatibility.
```
Geheiner: @zzeekk @pgruetter 
FYI the db parameter is incorrectly named, I spent some time finding that one out getting it to work with snowflake.
From what I understand, changing the name of db to schema would be a massive breaking change so I understand that we don't want to do it...

zzeekk: In my opinion the comment is still wrong, because there is no database-schema defined in the connection...

Geheiner: thanks, I updated the comment

docs/Reference.md
```diff
@@ -305,11 +305,11 @@ Specifying options allows to reuse a transformation in different settings.
 #### Java/Scala
 You can use Spark Dataset API in Java/Scala to define custom transformations.
```
Geheiner: I have never seen an example of an Extension of a CustomTransformer working with Datasets, only DataFrames. Would be interesting for some of our collegues.

sdl-snowflake/src/test/scala/io/smartdatalake/workflow/snowflake/SparkAndSnowparkDataPipelineIT.scala
```diff
@@ -61,15 +63,16 @@ object SparkAndSnowparkDataPipelineIT extends App {
   val srcDO = HiveTableDataObject( "src1", Some(tempPath+s"/${srcTable.fullName}"), table = srcTable, numInitialHdfsPartitions = 1)
   srcDO.dropTable
   instanceRegistry.register(srcDO)
-  val tgt1Table = Table(Some("test"), "tgt1", None, Some(Seq("lastname")))
-  val tgt1DO = SnowflakeTableDataObject( "tgt1", tgt1Table, connectionId = "sfCon")
+  val tgt1Table = Table(Some(System.getenv("SNOWFLAKE_SCHEMA")), "tgt1", None, Some(Seq("lastname")))
```
Geheiner, line 66: @zzeekk  FYI I changed your test to use environment variables. I hope it makes it a bit easier to see how to run the test (no confusion with db vs schema)

Geheiner (approved):
I'm done :)

sdl-snowflake/src/test/scala/io/smartdatalake/workflow/snowflake/SparkAndSnowparkDataPipelineIT.scala
zzeekk: ok

docs/Reference.md
zzeekk: I'll change that to "Spark DataFrame API", but we could create a Transformer working with Dataset in the future. I'll create feature request for that.

PR #494, opened 2022-04-04
fix naming of input views when chaining SQL transformations

zzeekk:
Temporary views created by SQL transformers are postfixed with "_sdltemp" to avoid naming conflicts.
It's implemented in backward compatible way - old naming convention is replaced with the new one if Environment.replaceSqlTransformersOldTempViewName=true - default is true.

Using the DataObject id as name for the temporary views in SQL transformers could have strange side effects if the name after replacing special chars was the same as a potential table in the metastore. Should SQL choose the temporary view or the table to read from? With Snowpark it's even not possible to create such views.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/ActionHelper.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/ActionHelper.scala
@@ -131,5 +131,20 @@ private[smartdatalake] object ActionHelper extends SmartDataLakeLogger {
     val invalidCharacters = "[^a-zA-Z0-9_]".r
     invalidCharacters.replaceAllIn(str, "_")
   }
+
+  /**
+   * Create a valid temporary view name for SQL transformation.
+   * Apart from replacing special characters, a postfix is added to make the name unique in case the input name is also an existing table.
+   * @param inputName name of the input the temporary view should be created for
+   */
+  def createTemporaryViewName(inputName: String) : String = {
+    replaceSpecialCharactersWithUnderscore(inputName) + TEMP_VIEW_POSTFIX
+  }
+
+  def replaceLegacyViewName(sql: String, inputViewName: String): String = {
+    sql.replaceAll("\\s"+inputViewName.stripSuffix(ActionHelper.TEMP_VIEW_POSTFIX)+"(\\s|\\.|$)", s" $inputViewName" + "$1")
+  }
+
+  val TEMP_VIEW_POSTFIX = "_sdltemp"
 }
 
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameOneToOneActionImpl.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameOneToOneActionImpl.scala
@@ -91,8 +91,8 @@ abstract class DataFrameOneToOneActionImpl extends DataFrameActionImpl {
    * apply transformer to SubFeed
    */
   protected def applyTransformers(transformers: Seq[GenericDfTransformerDef], inputSubFeed: DataFrameSubFeed, outputSubFeed: DataFrameSubFeed)(implicit context: ActionPipelineContext): DataFrameSubFeed = {
-    val transformedSubFeed = transformers.foldLeft(inputSubFeed){
-      case (subFeed, transformer) => transformer.applyTransformation(id, subFeed)
+    val (transformedSubFeed, _) = transformers.foldLeft((inputSubFeed,Option.empty[String])){
+      case ((subFeed,previousTransformerName), transformer) => (transformer.applyTransformation(id, subFeed, previousTransformerName), Some(transformer.name))
     }
     // Note that transformed partition values are set by execution mode.
     outputSubFeed.withDataFrame(transformedSubFeed.dataFrame)
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameOneToOneActionImpl.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameOneToOneActionImpl.scala
@@ -18,12 +18,10 @@
  */
 package io.smartdatalake.workflow.action
 
-import io.smartdatalake.workflow.action.generic.transformer.GenericDfTransformerDef
-import io.smartdatalake.workflow.action.spark.customlogic.CustomDfTransformerConfig
-import io.smartdatalake.workflow.dataobject.{CanCreateDataFrame, CanCreateSparkDataFrame, CanWriteDataFrame, CanWriteSparkDataFrame, DataObject}
+import io.smartdatalake.workflow.action.generic.transformer.{GenericDfTransformerDef, SQLDfTransformer}
 import io.smartdatalake.workflow.dataframe.spark.SparkSubFeed
+import io.smartdatalake.workflow.dataobject.{CanCreateDataFrame, CanWriteDataFrame, DataObject}
 import io.smartdatalake.workflow.{ActionPipelineContext, DataFrameSubFeed, SubFeed}
-import org.apache.spark.sql.Column
 
 import scala.reflect.runtime.universe.{Type, typeOf}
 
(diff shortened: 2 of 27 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/SQLDfsTransformer.scala
```diff
@@ -22,47 +22,72 @@ package io.smartdatalake.workflow.action.generic.transformer
...
-    dfs.foreach {
-      case (dataObjectId,df) =>
-        val objectId =  ActionHelper.replaceSpecialCharactersWithUnderscore(dataObjectId)
-        // Using createTempView does not work because the same data object might be created more than once
-        df.createOrReplaceTempView(objectId)
+    // register all inputs as temporary table
+    val inputViewNameOptions = dfs.map {
+      case (inputName,df) =>
+        val inputViewName =  ActionHelper.createTemporaryViewName(inputName)
+        // Using createTempView does not work because the same temporary view is created more than once (Init & Exec phase...)
```
Geheiner: Confusing comment since the function createTempView is nowhere to be found, i suggest to remove it

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/SQLDfsTransformer.scala
```diff
@@ -22,47 +22,72 @@ package io.smartdatalake.workflow.action.generic.transformer
...
-          val preparedSql = SparkExpressionUtil.substituteOptions(dataObjectId, Some(s"transformers.$name.code"), sql, options)
-          functions.sql(preparedSql,dataObjectId)
+          var preparedSql = SparkExpressionUtil.substituteOptions(actionId, Some(s"transformers.$name.code"), sql, options ++ inputViewNameOptions)
+          // for backward compatibility the temp view name from versions <= 2.2.x is replaced with the new temp view name including a postfix.
+          if (Environment.replaceSqlTransformersOldTempViewName) {
+            inputViewNameOptions.values.foreach(inputViewName => preparedSql = ActionHelper.replaceLegacyViewName(preparedSql, inputViewName))
+          }
+          // create DataFrame from SQL
+          logger.debug(s"($actionId.transformers.$name) Preparing DataFrame $outputName from SQL statement: $preparedSql")
+          functions.sql(preparedSql, outputDataObjectId)
```
Geheiner, line 87: I know it's a bit late for that comment now but... having to find out dataObjectIds everywhere just to please snowpark starts to get really confusing. Why can't we have information about the snowparksession in a similar way as for the spark session? Maybe we could have a map of sessions defined in the actionpipelineContext? The signature of the .sql function annoyed me before, but I thought if it s only used in a few places then let's leave it. Now it looks like it might be worth challenging it

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ActionHelper.scala
```diff
@@ -131,5 +131,20 @@ private[smartdatalake] object ActionHelper extends SmartDataLakeLogger {
...
+  /**
+   * Create a valid temporary view name for SQL transformation.
+   * Apart from replacing special characters, a postfix is added to make the name unique in case the input name is also an existing table.
+   * @param inputName name of the input the temporary view should be created for
+   */
+  def createTemporaryViewName(inputName: String) : String = {
+    replaceSpecialCharactersWithUnderscore(inputName) + TEMP_VIEW_POSTFIX
+  }
+
+  def replaceLegacyViewName(sql: String, inputViewName: String): String = {
```
Geheiner, line 144: could you add a unit test for that? Would make it easier to see what it does exactly

Geheiner, line 144: could you add y unit test for that? would make it easier to understand what it does

sdl-core/src/test/scala/io/smartdatalake/workflow/action/CustomDataFrameActionTest.scala
```diff
@@ -162,8 +163,9 @@ class CustomDataFrameActionTest extends FunSuite with BeforeAndAfter {
     tgtDO1.dropTable
     instanceRegistry.register(tgtDO1)
 
-    val customTransformer = SQLDfsTransformer(code = Map(tgtDO1.id -> s"select * from ${srcDO1.id.id}"))
+    val customTransformer = SQLDfsTransformer(code = Map(tgtDO1.id.id -> s"select * from ${srcDO1.id.id}"))
     val action1 = CustomDataFrameAction("action1", List(srcDO1.id), List(tgtDO1.id), transformers = Seq(customTransformer))
+    instanceRegistry.register(action1)
```
Geheiner, line 168: funny that you did not need that before

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/SQLDfsTransformer.scala
```diff
@@ -22,47 +22,72 @@ package io.smartdatalake.workflow.action.generic.transformer
...
     code.map {
-      case (dataObjectId,sql) =>
+      case (outputName,sql) =>
         val df = try {
-          val preparedSql = SparkExpressionUtil.substituteOptions(dataObjectId, Some(s"transformers.$name.code"), sql, options)
-          functions.sql(preparedSql,dataObjectId)
+          var preparedSql = SparkExpressionUtil.substituteOptions(actionId, Some(s"transformers.$name.code"), sql, options ++ inputViewNameOptions)
+          // for backward compatibility the temp view name from versions <= 2.2.x is replaced with the new temp view name including a postfix.
+          if (Environment.replaceSqlTransformersOldTempViewName) {
+            inputViewNameOptions.values.foreach(inputViewName => preparedSql = ActionHelper.replaceLegacyViewName(preparedSql, inputViewName))
```
Geheiner: use a different name than preparedSql. It's not the same than the one above, right?

sdl-core/src/test/scala/io/smartdatalake/workflow/action/generic/transformer/SQLDfTransformerTest.scala
```diff
@@ -0,0 +1,71 @@
...
+  instanceRegistry.register(tgtDO1)
+
+  val action1 = CopyAction("action1", srcDO1.id, tgtDO1.id)
+  instanceRegistry.register(action1)
+
+  val emptyDf = Seq((1,"a")).toDF("num","str")
+
+  test("options and view name token are replaced") {
+    val customTransformer = SQLDfTransformer(code = s"select num, %{option1} from %{inputViewName_src1}")
+    customTransformer.transformWithOptions(action1.id, Seq(), SparkDataFrame(emptyDf), srcDO1.id, Map("option1" -> "str"))
```
Geheiner, line 58: IF possible it would be great to see by what these placeholders get replaced. Helps to understand what's happening, now I need to debug the tests to find it out

sdl-core/src/test/scala/io/smartdatalake/workflow/action/CustomDataFrameActionTest.scala
zzeekk: It's because we need to lookup the Snowflake connection...

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/SQLDfsTransformer.scala
zzeekk: Yes its the same :-) But i could use foldLeft instead...

Geheiner: use foldleft

sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameOneToOneActionImpl.scala
```diff
@@ -91,8 +91,8 @@ abstract class DataFrameOneToOneActionImpl extends DataFrameActionImpl {
    * apply transformer to SubFeed
    */
   protected def applyTransformers(transformers: Seq[GenericDfTransformerDef], inputSubFeed: DataFrameSubFeed, outputSubFeed: DataFrameSubFeed)(implicit context: ActionPipelineContext): DataFrameSubFeed = {
-    val transformedSubFeed = transformers.foldLeft(inputSubFeed){
-      case (subFeed, transformer) => transformer.applyTransformation(id, subFeed)
+    val (transformedSubFeed, _) = transformers.foldLeft((inputSubFeed,Option.empty[String])){
```
Geheiner, line 94: test if all transformers.name are unique

zzeekk:
Created #495 as technical debt.

Geheiner (approved):

PR #500, opened 2022-04-06
revert intellij code suggestion to make default value more transparent

zzeekk:
I dont like the following code suggestion from Intellij's too much, as it hides the default value to be used, e,g, its not obvious to me that None.forall(_.toBoolean) == true.
I think the following is better:
.map(_.toBoolean).getOrElse(true)


Geheiner (approved):
agreed 👍

PR #508, opened 2022-04-12
Feature/documentation version dropdown

hiltpold:
A dropdown has been added to the JsonSchemaViewer from the documentation. It is now possible to change the version of the schema and display it. The current exporter of the schema has to be updated, as the "id" field in the schema json should equal the filename.

Better usability of the documentation.


Geheiner (approved):
LGTM, seems to work, tried it in my browser. FYI I changed the base branch to documentation.

mand35:
This is a neat feature. 
I am not sure if this is necessary, but it would be more consistent if the link to share also reference to the version selected. Otherwise we may reference to an object which does not exists anymore in the latest version.

static/json-schema-viewer/index.html
```diff
@@ -6,7 +6,17 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jquery-mobile/1.4.5/jquery.mobile.min.css" />
         <link rel="stylesheet" href="styles/json-schema-viewer.min.css" />
-
+        <style>
+            .dropdown-container{
+                display: flex;
+                align-items: center;
+                display: flex;
```
mand35, line 13: Is that repetition of `display` and `align-items` necessary here?

hiltpold: Good catch. Might be artifacts from me using vim :)

zzeekk:
> This is a neat feature. I am not sure if this is necessary, but it would be more consistent if the link to share also reference to the version selected. Otherwise we may reference to an object which does not exists anymore in the latest version.

Good point. It would be nice to have to add the version as URL parameter...

mand35:
looks great with that fixed link now. 👍

zzeekk (approved):

Issue #489, opened 2022-04-01
Create a Scala Transformer providing Dataset API


zzeekk:
**Is your feature request related to a problem? Please describe.**
Create a ScalaClassSparkDsTransformer, which allows to define a transformation by using Spark Dataset API.
Using Spark Dataset API allows using case classes to define and validate the schema required by the transformation.

**Describe the solution you'd like**
ScalaClassSparkDsTransformer allows to implements a one-to-one transformation in Scala with Spark Dataset API. An additional configuration parameter caseClass must be provided to define the schema of the input dataset. At runtime SDLB will receive a DataFrame an try to convert it to a corresponding Dataset.
The transformation can create an arbitrary Dataset as output, which is converted back to a DataFrame by SDLB.

**Additional context**
In the future we could also create a many-to-many transformation using Spark Dataset API, but i suggest to validate the approach in 1:1 transformation first.

Geheiner:
So, the next step is to support a generic number of inputs/outputs.
I think the best way to implement this is analoguous to how Scala Itself implemented Tuples:
We create Transformers for all possible numbers of input/output datasets. The number of classes that would be needed is the maximum number of inputs times the maximum number of outputs. I suggest to start by supporting a maximum number of 5 for now.
This would require to create 25 transformer classes (24 in addition to the one that was already created).
Here is the list of all possible combinations of transformers that would need to be built:
```
1:1

2:2
2:1
1:2

3:3,
3:2,3:1,
1:3,2:3

4:4
4:1,4:2,4:3,
1:4,2:4,3:4

5:5
5:1,5:2,5:3,5:4,
1:5,2:5,3:5,4:5

num combinations with 1 = 1
num combinations with 2 = 4
num combinations with 3 = 9
num combinations with 4 = 16 
num combinations with 5 = 25
```

Geheiner:
Now there exists a generic N to 1 transformer for Datasets, some improvements have been added in this PR: https://github.com/smart-data-lake/smart-data-lake/pull/545
The feature has been tested in a real setting and it works, the ticket can be closed

PR #513, opened 2022-04-22
implement ScalaClassSparkDsTransformer for one to one DataSet transfo…

Geheiner:
Implementing ScalaClassSparkDsTransformer : https://github.com/smart-data-lake/smart-data-lake/issues/489


Geheiner:
Hi @zzeekk , Thanks for your feedback this afternoon! It allowed me to implement this much faster than I had expected.
I did not create a Test yet that parses a config File. I will add one next week. But I thought this is a good time for a review, the test that I wrote shows that it's working

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.spark.sql.{DataFrame, Dataset, Encoder, SparkSession}
+
+import scala.reflect.runtime.universe.TypeTag
+
+/**
+ * Interface to define a custom Spark-Dataset transformation (1:1)
+ * When you implement this interface, you need to provide two case classes: One for your input Dataset
+ * and one for your output Dataset.
+ */
+trait CustomDsTransformer[In <: scala.Product, Out <: scala.Product] extends Serializable {
```
zzeekk:
```suggestion
trait CustomDsTransformer[In : Encoder, Out : Encoder] extends Serializable {
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
```diff
@@ -0,0 +1,65 @@
...
+  implicit def outputEncoder(implicit typeTag: TypeTag[Out]): Encoder[Out] = org.apache.spark.sql.Encoders.product[Out]
```
zzeekk:
```suggestion
  implicit def outputEncoder: Encoder[Out] = implicitly[Encoder[Out]]
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
```diff
@@ -0,0 +1,65 @@
...
+   * @param session      Spark Session
+   * @param options      Options specified in the configuration for this transformation
+   * @param inputDS      Input Dataset
+   * @param dataObjectId Id of DataObject of SubFeed
+   * @return Transformed DataFrame
+   */
+  def transform(session: SparkSession, options: Map[String, String], inputDS: Dataset[In], dataObjectId: String): Dataset[Out]
+
+  private[smartdatalake] def transformWithTypeConversion(session: SparkSession, options: Map[String, String], inputDf: DataFrame, dataObjectId: String)(implicit typeTag: TypeTag[In]): DataFrame = {
+    val inputDSEncoder = org.apache.spark.sql.Encoders.product[In]
```
zzeekk, line 46:
```suggestion
    val inputDSEncoder = implicitly[Encoder[In]]
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
```diff
@@ -0,0 +1,65 @@
...
+    transform(session, options, inputDf.as(inputDSEncoder), dataObjectId).toDF
```
zzeekk, line 47: Here you should be able to write "inputDf.as[In]"

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDsTransformer.scala
```diff
@@ -0,0 +1,65 @@
...
+ * @param name                 name of the transformer
+ * @param description          Optional description of the transformer
+ * @param transformerClassName class name implementing trait [[CustomDsTransformer]]
+ * @param options              Options to pass to the transformation
+ * @param runtimeOptions       optional tuples of [key, spark sql expression] to be added as additional options when executing transformation.
+ *                             The spark sql expressions are evaluated against an instance of [[DefaultExpressionData]].
+ */
+case class ScalaClassSparkDsTransformer(override val name: String = "scalaSparkTransform", override val description: Option[String] = None, transformerClassName: String, options: Map[String, String] = Map(), runtimeOptions: Map[String, String] = Map()) extends OptionsSparkDfTransformer {
+
+  private val customTransformer = CustomCodeUtil.getClassInstanceByName[CustomDsTransformer[InputDsType, OutputDsType]](transformerClassName)
```
zzeekk:
```suggestion
  private val customTransformer = CustomCodeUtil.getClassInstanceByName[CustomDsTransformer[_, _]](transformerClassName)
```

zzeekk: Couldn't we leave In and Out types unspecified here? It would also make InputDsType and OutputDsType superfluous.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
zzeekk: The encoder instance can be retrieved like that. But this is already implicitly available - so this function is actionally superfluous... where is it used?

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
zzeekk: We could make this more generic by telling Scala that there must be an implicit Encoder[In].

zzeekk (commented):
Nice. See my comments about making this even more generic.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
Geheiner: Yes, I could change this code to 
```suggestion
      implicit val inputDSEncoder = org.apache.spark.sql.Encoders.product[In]
    transform(session, options, inputDf.as[In], dataObjectId).toDF
```
But I think it's just less clear that way. I see the point of implicits if it were used in many places, but if it's just this one?

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDsTransformer.scala
Geheiner: Unfortunatelly, I need to provide some class that suits the Encoder. 
Putting _ results in this error:
`A needed class was not found. This could be due to an error in your runpath. Missing class: no Java class corresponding to _$1 found
java.lang.NoClassDefFoundError: no Java class corresponding to _$1 found
	at scala.reflect.runtime.JavaMirrors$JavaMirror.typeToJavaClass(JavaMirrors.scala:1361)
	at scala.reflect.runtime.JavaMirrors$JavaMirror.runtimeClass(JavaMirrors.scala:234)
	at scala.reflect.runtime.JavaMirrors$JavaMirror.runtimeClass(JavaMirrors.scala:69)
	at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder$.apply(ExpressionEncoder.scala:54)
	at org.apache.spark.sql.Encoders$.product(Encoders.scala:300)
	at io.smartdatalake.workflow.action.spark.customlogic.CustomDsTransformer.transformWithTypeConversion(CustomDsTransformer.scala:47)`
But it seams to not matter what class this is. So Instead of Writing InputDSType and OutputDSType, I can just use a new class abstract class DsType extends scala.Product for both input and output and the encoder stops complaining

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
Geheiner: It will be used by the user's code. It allows the user to map his data back to a DataSet of the Output Datatype.
Concretely, I need this so that my Unit Tests compiles in the File ScalaClassSparkDsTransformerTest at line 44:
inputDS.withColumn("doubled_rating", col("rating") * 2).**as[OutputDataSet]**

Geheiner: Applying your suggestion results in a Stackoverflow error:
`An exception or error caused a run to abort. 
java.lang.StackOverflowError
	at io.smartdatalake.workflow.action.spark.customlogic.CustomDsTransformer.outputEncoder(CustomDsTransformer.scala:33)`

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
Geheiner: When I apply your suggestions I get this error:
```
traits cannot have type parameters with context bounds `: ...' nor view bounds `<% ...'
trait CustomDsTransformer[In : Encoder, Out : Encoder] extends Serializable {
```
Tha'ts why I put <: scala.Product

Geheiner:
I'm done now, I added a unit-Test that parses a conf file.
One thing to note is that with Datasets, types get more important. So the InputDataObject needs to have a schema that corresponds to the InputDataSet. In the test, I ensured that by specifying schema = """name string, rating int""" in the config of the DataObject.

zzeekk (approved):

zzeekk:
LGTM. Please use "Squash and merge" and add issue (#489) at end of new committ message.

PR #516, opened 2022-04-22
Improve Housekeeping

### LINKED ISSUE #514: Housekeeping not deleting directories with more than two levels
Housekeeping mode correctly works with one level of partitioning. 
Example:
JsonFileDataObject partitioned by [run_id]: partition after 10 run_id. All files are moved, then the directories for the run_ids are deleted.

When using two levels of partitioning or more, only the lowest directory level is deleted. The upper levels stay empty. 
Example:
JsonFileDataObject partitioned by [run_id, domain]: partition after 10 run_id. All files are moved, but only the directory domain is deleted. The empty run_id folder is preserved.

Delete empty partitions when executing housekeeping.

### LINKED ISSUE #515: Housekeeping for DeltaLake
Housekeeping for DeltaLakeTableDataObject is not correctly implemented yet. If you just move around parquet files of a Delta Table, you will break that table as the transaction log is not updated.

The best solution is probably to execute an update statement on the Delta Table when executing housekeeping. 
Example: 
DeltaLakeTableDataObject partitioned by run_id. 
```
update <table> set run_id=0 where run_id=1;
update <table> set run_id=0 where run_id=2;
...
```
This will automatically move the files appropriately while delta lake maintains the transaction log.

**Additional context**
Could even be further optimized by compacting the update statements
`update <table> set run_id=0 where run_id=1-10;`
But I assume this is hard to do because the user can enter any kind of expression for housekeeping.

Also, if anything was archived, the delta table could be vacuumed to clean up old files.

zzeekk:
see #514, #515

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/hdfs/HdfsUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/hdfs/HdfsUtil.scala
@@ -137,6 +137,35 @@ private[smartdatalake] object HdfsUtil extends SmartDataLakeLogger {
   }
 
   /**
+   * Deletes parent directories of path if they are empty.
+   * Handles all parent directories up-to base path.
+   * Stops if a not empty directory is found.
+   */
+  def deleteEmptyParentPath( path: Path, basePath: Path )(implicit fs: FileSystem) : Unit = {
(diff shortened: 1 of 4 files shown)


pgruetter (approved):
Looks good, thanks! I didn't test it yet, will do after merge as it's easier to test the snapshot.

PR #525, opened 2022-05-17
Feature/522 init phase schema 2

### LINKED ISSUE #522: Improve Schema initializition for SparkFileDataObjects
A pipeline with a FileTransferAction followed by CopyAction cannot be run together in the first run, if no schema is defined for CopyActions input SparkFileDataObject. This is because it cannot infer schema in init-phase, as no files are copied yet.

There is a similar case with two CopyActions in a row: If the 2nd is working with FileIncrementalMoveMode and is executed standalone, there might be no file available for schema inference.

Init-Phase should do as much as possible to pass the schema for DataObjects to the next Action. SparkFileDataObject isn't yet there. We should save the schema information in a separate file in Init-Phase:
- when writing a DataFrame the schema could be written as Json-File ".../_schema/currentSchema.json"
- when writing a File into the directory (by FileTransferAction), we could create a sample file for future schema inference, e.g. for CsvFileDataObject a ".../_sample/<originalFilename>.csv" file.

I suggest to make this configurable in Environment object with the following defaults:
- update schema file in Init-Phase: default=true
- update sample file in Init-Phase: default=false

zzeekk:
- implement sample data file and schema file for SparkFileDataObject (#522)
- improve FilesObservation in SparkFileDataObject
- use FileSystem instead of FileContext API in HadoopFileDataObject

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
@@ -194,6 +194,26 @@ object Environment {
       .map(_.toBoolean).getOrElse(true)
   }
 
+  /**
+   * If enabled, the sample data file for SparkFileDataObject is updated on every load from a file-based Action, otherwise it's just updated if it's missing.
+   * The advantage of updating the sample file on every load is to enable automatic schema evolution.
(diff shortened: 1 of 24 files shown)


pgruetter (approved):
👍

PR #526, opened 2022-05-17
Fix Airbyte supportsIncremental optional

zzeekk:
Make AirbyteConnectorSpecification supportsIncremental configuration attribut optional

Assume true if not given, otherwise MSSQL CDC connector does not work.


mand35:
Now the feature looks good. Not sure if we can add a test for Airbyte, because we would need a "remote" system, typically with some kind of authorization.

mand35 (approved):
looks good.

PR #527, opened 2022-05-17
Feature/spark dataset transformation #489

Geheiner:
Continue Implementing #489: ScalaClassSparkDs2To1Transformer

@zkull I decided to let the user decide how he wants the transform method parameters to be resolved to the dataObjects.
There is now an Option called parameterResolution that defaults to dataObjectId (your way) but that can also be set to dataObjectOrdering (my way )


sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDs2to1Transformer.scala
```diff
@@ -0,0 +1,98 @@
...
+   * @param session      Spark Session
+   * @param options      Options specified in the configuration for this transformation
+   * @param inputDS      Input Dataset
+   * @param dataObjectId Id of DataObject of SubFeed
+   * @return Transformed DataFrame
+   */
+
+  def transform(session: SparkSession, options: Map[String, String], ds1: Dataset[In1], ds2: Dataset[In2]): Dataset[Out]
+
+  def getTypeTag: Type = throw new IllegalArgumentException("When using option \"parameterResolution\" -> \"dataObjectId\"," +
```
Geheiner: I'm curious if you see another way of doing this... I did not find any other way to get the the type tag of the User's Class other than forcing the user to pass it onto me

Geheiner:
A seamingly unrelated Test
"sdlb run with 2 actions and positive top-level partition values filter, recovery after action 2 failed the first time"
is failing on this assert:
 TestSDLPlugin.startupCalled was false (SmartDataLakeBuilderTest.scala:163)
No idea why... But my code is ready for review, I will look at the build problem later...

Geheiner:
Build problem solved, thanks @zkull ;)

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDs2to1Transformer.scala
```diff
@@ -0,0 +1,98 @@
...
+    // He could only do that by defining another method called transform that has a type-signature different from the one that is expected.
+    // We only keep the transform method that matches exactly the expected type signature
+    val transformMethodsOfSubclass = typeTagSubclass.members.filter(_.isMethod).filter(_.fullName.endsWith(".transform"))
+    val transformMethodsToPick = transformMethodsOfSubclass.filter(method => method.info.paramLists.head.map(_.typeSignature).map(_.typeSymbol) == typesOfParamsOfParentTransformMethod)
+
+    assert(transformMethodsToPick.size == 1)
+
+    val transformParameterNames: Seq[String] = transformMethodsToPick.head.info.paramLists.head.map(_.name).map(_.toString)
+
+    //Parameters at index one and two are sparkSession and options, see method signature of transform
```
Geheiner:
```suggestion
    //Parameters at index 0 and 1 are sparkSession and options, see method signature of transform
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDs2to1Transformer.scala
```diff
@@ -0,0 +1,98 @@
...
+package io.smartdatalake.workflow.action.spark.customlogic
+
+import io.smartdatalake.util.hdfs.PartitionValues
+import io.smartdatalake.workflow.dataobject.DataObject
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+
+import scala.reflect.runtime.universe.{Type, TypeTag, typeOf}
+
+/**
+ * Interface to define a custom Spark-Dataset transformation (1:1)
```
zzeekk: Comment is still for 1:1... but this is 2:1.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDs2to1Transformer.scala
```diff
@@ -0,0 +1,98 @@
...
+   * @param dataObjectId Id of DataObject of SubFeed
+   * @return Transformed DataFrame
+   */
+
+  def transform(session: SparkSession, options: Map[String, String], ds1: Dataset[In1], ds2: Dataset[In2]): Dataset[Out]
+
+  def getTypeTag: Type = throw new IllegalArgumentException("When using option \"parameterResolution\" -> \"dataObjectId\"," +
+    " you need to define method getTypeTag next to your transform method as follows: override def getTypeTag: universe.Type = typeOf[this.type]")
+
+  private[smartdatalake] def transformBasedOnDataObjectOrder(session: SparkSession, options: Map[String, String], inputDos: Seq[DataObject], dfs: Map[String, DataFrame])
```
zzeekk:
```suggestion
  private[smartdatalake] def transformBasedOnDataObjectOrder(session: SparkSession, options: Map[String, String], inputDOs: Seq[DataObject], dfs: Map[String, DataFrame])
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDs2to1Transformer.scala
```diff
@@ -0,0 +1,98 @@
...
+                                                            (implicit typeTag1: TypeTag[In1], typeTag2: TypeTag[In2]): DataFrame = {
+    val inputDS1Encoder = org.apache.spark.sql.Encoders.product[In1]
```
zzeekk:
```suggestion
    val inputDs1Encoder = org.apache.spark.sql.Encoders.product[In1]
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDs2to1Transformer.scala
```diff
@@ -0,0 +1,98 @@
...
+    val inputDf2 = dfs(inputDo2.id.id)
+    transform(session, options, inputDf1.as(inputDS1Encoder), inputDf2.as(inputDS2Encoder)).toDF
+  }
+
+  private[smartdatalake] def transformBasedOnDataObjectId(session: SparkSession, options: Map[String, String], dfs: Map[String, DataFrame])
+                                                         (implicit typeTag1: TypeTag[In1], typeTag2: TypeTag[In2]): DataFrame = {
+    val inputDS1Encoder = org.apache.spark.sql.Encoders.product[In1]
+    val inputDS2Encoder = org.apache.spark.sql.Encoders.product[In2]
+
+    val typeTagSubclass = getTypeTag
```
zzeekk: Tricky to get the Type of the subclass, yes. But you can use getClass which refers to the subclass, and convert it to a type:
```
    val mirror = scala.reflect.runtime.currentMirror
    val typeTagSubclass = mirror.classSymbol(this.getClass).toType
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDs2To1Transformer.scala
```diff
@@ -0,0 +1,76 @@
...
+ * Define a transform function that receives a SparkSession, a map of options and two DataSets and that has to return a Dataset.
+ * The Java/Scala class has to implement interface [[CustomDs2to1Transformer]].
+ * By default, the Parameter Names of the transform function must match the dataObjectIds provided in the the inputIds list of the config.
+ * By setting the option "parameterResolution"= "dataObjectOrdering", you can name your dataObjects however you want,
+ * but then the ordering in of the inputIds list must match the ordering of the parameters of the transform function.
+ *
+ * @param name           name of the transformer
+ * @param description    Optional description of the transformer
+ * @param className      class name implementing trait [[CustomDfsTransformer]]
+ * @param options        Either: "parameterResolution" -> "dataObjectId" or "parameterResolution" -> "dataObjectOrdering". Default: "parameterResolution" -> "dataObjectId"
```
zzeekk: parameterResolution is not a user specific option, but a configuration option of SDLB. Therefore i would create a specific parameter "parameterResolution" with default value "dataObjectId".

sdl-core/src/test/scala/io/smartdatalake/workflow/action/ScalaClassSparkDs2To1TransformerTest.scala
```diff
@@ -0,0 +1,227 @@
...
+      List(
+        new Directory(new File("target/src1DS2to1")),
+        new Directory(new File("target/src2DS2to1")),
+        new Directory(new File("target/tgt1DS2to1")),
+      )
+    }
+    directoriesToDelete.foreach(dir => dir.deleteRecursively())
+  }
+
+    test("One DS2To1 Transformation using config file: two different input types using dataObjectOrdering") {
```
zzeekk: indentation not right

sdl-core/src/test/scala/io/smartdatalake/workflow/action/ScalaClassSparkDs2To1TransformerTest.scala
```diff
@@ -0,0 +1,227 @@
...
+    val actual = tgt1DO.getSparkDataFrame().as[AnotherOutputDataSet].head()
+    assert(actual.added_rating == 15)
+    assert(actual.concatenated_name == "johndoe")
+  }
+
+  test("One DS2To1 Transformation using config file: two identical input types using dataObjectId") {
+
+    // setup DataObjects
+    // source has partition columns dt and type
+    val srcDO1 = CsvFileDataObject("src1", "target/src1DS2to1", partitions = Seq("name")
```
zzeekk: Can we use a tempPath here as well?
I suggest to use a different tempPath per unit test.

sdl-core/src/test/scala/io/smartdatalake/workflow/action/ScalaClassSparkDsTransformerTest.scala
```diff
@@ -59,14 +59,16 @@ class ScalaClassSparkDsTransformerTest extends FunSuite with BeforeAndAfter {
   before {
     instanceRegistry.clear()
   }
+  after {
```
zzeekk, line 62: It's sufficient to clear instanceRegistry before.

zzeekk (commented):
Nice, you got it working.
Please review the Naming of Ds und DO in variables and classes.
Dataset should be abbreviated Ds, and DataObject is better readable as DO.
I know that we are not consistent with DataFrame, which is often abbreviated as Df...

Further i'm still thinking about how to make this more generic to create an n:1 transformer.
By reflection we could search for a corresponding transform-method implementation, and probably we could also get the typeTag for the parameter type dynamically... and maybe we could also create the Product Encoder, which is an ExpressionEncoder dynamically, as it's apply method looks as follows:
```
object ExpressionEncoder {

  def apply[T : TypeTag](): ExpressionEncoder[T] = {
    val mirror = ScalaReflection.mirror
    val tpe = typeTag[T].in(mirror).tpe

    val cls = mirror.runtimeClass(tpe)
    val serializer = ScalaReflection.serializerForType(tpe)
    val deserializer = ScalaReflection.deserializerForType(tpe)

    new ExpressionEncoder[T](
      serializer,
      deserializer,
      ClassTag[T](cls))
  }
```

sdl-core/src/test/scala/io/smartdatalake/workflow/action/ScalaClassSparkDs2To1TransformerTest.scala
Geheiner: I don't know how to synchronize the dynamically generated location of the temp path with the location of the dataObjects defined in the config files. Maybe using environment variables would be an option, but I don't think this is a good idea. What do you think?
I'm writing to target because I got inspired by io.smartdatalake.app.SmartDataLakeBuilderTest ;)

Geheiner:
> Nice, you got it working. Please review the Naming of Ds und DO in variables and classes. Dataset should be abbreviated Ds, and DataObject is better readable as DO. I know that we are not consistent with DataFrame, which is often abbreviated as Df...
> 
> Further i'm still thinking about how to make this more generic to create an n:1 transformer. By reflection we could search for a corresponding transform-method implementation, and probably we could also get the typeTag for the parameter type dynamically... and maybe we could also create the Product Encoder, which is an ExpressionEncoder dynamically, as it's apply method looks as follows:
> 
> ```
> object ExpressionEncoder {
> 
>   def apply[T : TypeTag](): ExpressionEncoder[T] = {
>     val mirror = ScalaReflection.mirror
>     val tpe = typeTag[T].in(mirror).tpe
> 
>     val cls = mirror.runtimeClass(tpe)
>     val serializer = ScalaReflection.serializerForType(tpe)
>     val deserializer = ScalaReflection.deserializerForType(tpe)
> 
>     new ExpressionEncoder[T](
>       serializer,
>       deserializer,
>       ClassTag[T](cls))
>   }
> ```
Thanks for the review, it's a fun feature to implement!
If it's a completely generic transformer, then the user would have to pass a List[Any] as arguments, which would not be so nice. We would loose compile time type safety... Or do you have something else in mind?

Geheiner:
One test is still failing will look at it later...

Geheiner:
Finally I'm done. There were indeed some corner cases which were not covered in the beginning, such as passing on 3 DOs but having only 2DOs as arguments in transform method.
Further while debugging one of my tests I saw that in case of errors, the Stacktrace is not showing the stacktrace frames located in the transform method by default because it is called via reflection, so I logged them as error in addition. That was useful for me and will be useful for the users of the framework!

zzeekk (approved):

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsTransformer.scala
```diff
@@ -37,13 +37,14 @@ trait CustomDsTransformer[In <: Product, Out <: Product] extends Serializable {
    * @param session      Spark Session
    * @param options      Options specified in the configuration for this transformation
    * @param inputDS      Input Dataset
+   * @param dataObjectId name of the input Dataset
    * @return Transformed DataFrame
    */
-  def transform(session: SparkSession, options: Map[String, String], inputDS: Dataset[In]): Dataset[Out]
+  def transform(session: SparkSession, options: Map[String, String], inputDS: Dataset[In], dataObjectId: String): Dataset[Out]
```
Geheiner, line 43: @zzeekk 
Something I forgot to ask and now notice when I use it:
The signature of transform of CustomDsTransformer is a bit inconsistent with the one expected for the Custom DSNto1Transformer. Why do we need this dataObjectId here? We don't have it in the other one

PR #528, opened 2022-05-17
call move/compactPartition only if list is not empty

zzeekk:
call move/compactPartition only if list is not empty in PartitionArchiveCompactionMode

compactPartition is not implemented by DeltaLakeTableObject, but we want to use Archive Partition


pgruetter (approved):
Thanks for the quick fix

PR #532, opened 2022-05-19
Proposal for Nto1-transformer

zzeekk:
Tried to implement a N-to-1 Dataset transformer and surprisingly it works :-)

Transformer implementation can define any method with name "transform" and parameters of type Dataset, SparkSession and Map[String,String]

What do you think about this?

(needs some cleanup before merging)


Geheiner:
My initial reaction:
image

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsNto1Transformer.scala
```diff
@@ -0,0 +1,75 @@
...
+      case param if param.typeSignature =:= typeOf[SparkSession] =>
+        session
+      case param if param.typeSignature =:= typeOf[Map[String,String]] =>
+        options
+      case param => throw new IllegalStateException(s"Transform method parameter with unallowed parameter type ${param.typeSignature.typeSymbol.name.toString} found. Only parameters of type Dataset, SparkSession and Map[String,String] are allowed.")
+    }
+
+    // call method dynamically
+    val transformMethodInstance = getClass.getMethods.find(_.getName == methodName).get
+    transformMethodInstance.invoke(this, transformParameterInstances:_*).asInstanceOf[Dataset[_]].toDF
```
Geheiner, line 61: So, if I understand correctly:
Scala does not care about the enclosed type of the Datasets, that's why you can just give it Dataset[_].
The only place that was picky about knowing if its a Dataset[A] or Dataset[B] on compile-time was this Encoder, but you managed to implement a less picky one yourself 🥇

Geheiner:
```suggestion
    transformMethodInstance.invoke(this, transformParameterInstances:_*).asInstanceOf[DataFrame]
   ```
works as well and is less confusing IMO

File: sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsNto1Transformer.scala [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+ */
+trait CustomDsNto1Transformer extends Serializable {
+
+  private[smartdatalake] def transformBasedOnDataObjectId(session: SparkSession, options: Map[String, String], dfs: Map[String, DataFrame]): DataFrame = {
+
+    // lookup transform method
+    val methodName = "transform"
+    val mirror = scala.reflect.runtime.currentMirror
+    val typeTagSubclass = mirror.classSymbol(this.getClass).toType
+    val transformMethodsOfSubclass = typeTagSubclass.members.filter(_.isMethod).filter(_.name.toString == methodName)
```
Geheiner, line 40: What I find is incredible is that the user's transform method is never directly called now, we are finding it with reflection :D We are doing assembler-style programming with scala :P

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDs2To1Transformer.scala
```diff
@@ -57,7 +57,7 @@ case class ScalaClassSparkDs2To1Transformer(override val name: String = "ScalaCl
 
     options("parameterResolution") match {
       case "dataObjectId" => Map(outputDo.id.id -> customTransformer.transformBasedOnDataObjectId(context.sparkSession, options, dfs))
-      case "dataObjectOrdering" => Map(outputDo.id.id -> customTransformer.transformBasedOnDataObjectOrder(context.sparkSession, options, inputDos, dfs))
+      //case "dataObjectOrdering" => Map(outputDo.id.id -> customTransformer.transformBasedOnDataObjectOrder(context.sparkSession, options, inputDos, dfs))
```
Geheiner, line 60: I would still like to keep this as an option. I've seen in my project the need to re use the same DatasetTransformation in different places. I think it should also be possible with your implementation. Do you mind if I re-implement it for your new transformer?

Geheiner:
> Tried to implement a N-to-1 Dataset transformer and surprisingly it works :-)
> 
> Transformer implementation can define any method with name "transform" and parameters of type Dataset, SparkSession and Map[String,String]
> 
> What do you think about this?
> 
> (needs some cleanup before merging)

Well, looks like you did it!
I will re-add my ordering feature if that s ok for you, and then add a couple of more tests.
We could even do the N to N transformer now with the same kind of Logic I guess... But one thing at a time :D

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsNto1Transformer.scala
zzeekk: Mh, we get a Dataset[Product] and can cast that to a DataFrame which is a Dataset[Row]?!

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsNto1Transformer.scala
zzeekk: Yes, we already use similar "patterns" in ConfigParser...

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDs2To1Transformer.scala
zzeekk: Yes sure. I just made a draft how we could implement Nto1.

zzeekk:
Yes we could also do NtoM, but i agree on one thing at the time.

I suggest we work further on your branch, and you adapt it as needed with my proposal and your extensions.
Hope to merge this soon ;-)

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDsNto1Transformer.scala
Geheiner: well the unit test works with it :D

PR #535, opened 2022-05-22
implement schema providers for caseClass, jsonSchemaFile and xsdFile (#512)

### LINKED ISSUE #512: Define DataObject Schema by Scala CaseClass
For several DataObjects a schema can be specified. Currently this has to be done as Spark DDL. This is ok if you have no Java project, otherwise it would be nice to define the schema by a case class, as done for Spark Datasets.

Allow to specify a case class name as schema instead of Spark DDL string. Use prefix "caseClass#" to differentiate from DDL string.

zzeekk:
see #512

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
@@ -19,16 +19,26 @@
 
 package io.smartdatalake.util.misc
 
+import com.databricks.spark.xml.util.XSDToSchema
+import io.smartdatalake.config.{ConfigUtil, ConfigurationException}
+import io.smartdatalake.util.hdfs.HdfsUtil
+import io.smartdatalake.util.json.{SchemaConverter => JsonSchemaConverter}
 import io.smartdatalake.workflow.dataframe._
+import org.apache.hadoop.conf.Configuration
+import org.apache.spark.sql.Encoders
+import org.apache.spark.sql.catalyst.JavaTypeInference
 import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.{ArrayType, StructType}
+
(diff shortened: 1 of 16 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
```diff
@@ -87,18 +97,18 @@ object SchemaUtil {
    * @param ignoreNullability whether to ignore differences in nullability.
    * @return `true` iff `leftType` is a subset of `rightType`. `false` otherwise.
    */
-  private def deepIsTypeSubset(leftType: GenericDataType, rightType: GenericDataType, ignoreNullability: Boolean, caseSensitive: Boolean ): Boolean = {
-    if (leftType.typeName != rightType.typeName) false  /*fail fast*/
+  private def deepIsTypeSubset(leftType: GenericDataType, rightType: GenericDataType, ignoreNullability: Boolean, caseSensitive: Boolean): Boolean = {
```
pgruetter, line 100: In `schemaDiff`, the parameter is called `ignoreNullable`. Can you please use the same term here?

pgruetter (approved):
Nice feature, this saves us converting a given XSD/JSON schema to a DDL string 👍 
Added one comment, but looks good to me otherwise.

zzeekk:
Improvements to do for xsdSchema:
- support empty RowTag
- RowTag must be given as XPath expression, but RowTag of SparkDataSource is just a tag. If the tag is present in different places of the schema, every occurance is read by SparkDataSource - this might create confusion if schemas of the different places is not the same...

zzeekk:
see #548

sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
zzeekk: done

PR #543, opened 2022-06-15
Bugfix transformer dataframe output mapping and input partitionvalues

zzeekk:

CustomDataFrameAction:
- allow using intermediate DataFrames as Action output 
- allow  last transformer output to be unused in Action outputs
-> These restrictions are not really usefull.

Fix applying mainInput partitionValues to other inputs
-> This is how it should work. PartitionValues from mainInput should be applyed to other inputs of an Action if possible and those inputs have no partitionValues yet.

Implement DataFrameActionImpl.breakDataFrameOutputLineage
-> This is needed because Deduplicate- and HistorizeAction write different data than what they got as input. It depends on incrementalMode setting and might be different from partitionValues for DeduplicateAction. Its cleaner to force the next action to request a new DataFrame from the source.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/HistorizeAction.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/HistorizeAction.scala
@@ -169,7 +169,7 @@ case class HistorizeAction(
 
   private[smartdatalake] override val handleRecursiveInputsAsSubFeeds: Boolean = false
 
-  // DataFrame created by DeduplicateAction should not be passed on to the next Action, but must be recreated from the DataObject.
+  // DataFrame created by HistorizeAction should not be passed on to the next Action, but must be recreated from the DataObject.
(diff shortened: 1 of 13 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/action/HistorizeAction.scala
```diff
@@ -169,6 +169,9 @@ case class HistorizeAction(
 
   private[smartdatalake] override val handleRecursiveInputsAsSubFeeds: Boolean = false
 
+  // DataFrame created by DeduplicateAction should not be passed on to the next Action, but must be recreated from the DataObject.
```
pgruetter:
```suggestion
  // DataFrame created by HistorizeAction should not be passed on to the next Action, but must be recreated from the DataObject.
```

pgruetter (approved):
LGTM, just a small copy/paste typo

PR #554, opened 2022-07-29
Workaround for NPE with FilesObserver

zzeekk:
Avoid NPE in FilesObserver if there is no data to process

https://issues.apache.org/jira/browse/SPARK-39044

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/SparkFileDataObject.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/SparkFileDataObject.scala
@@ -18,7 +18,6 @@
  */
 package io.smartdatalake.workflow.dataobject
 
-import io.smartdatalake.config.SdlConfigObject
 import io.smartdatalake.config.SdlConfigObject.ActionId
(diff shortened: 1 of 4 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/SparkFileDataObject.scala
```diff
@@ -208,6 +210,12 @@ private[smartdatalake] trait SparkFileDataObject extends HadoopFileDataObject
         }
     }
 
+    // early check for no data to process.
+    // This also prevents an error on Databricks when using filesObserver if there are no files to process. The
```
pgruetter: The .... ?

pgruetter (approved):
👍 Just one comment that ends abruptly.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/SparkFileDataObject.scala
zzeekk: thanks, fixed

PR #556, opened 2022-08-03
Feature/512 improve parsing xsd schema

zzeekk:
1. Implement parsing complex xml schemas:
  - support for attributeGroup
  - avoid Stackoverflow Error for recursive definitions by implementing max recursion depth
  - nested lists with maxOccurs >0 on xs:sequence tag
 
2. Combine multiple rowTags to create schema for parser and DataObject. This is needed to parse same element name with different type in different branch.

see open point in #512


mand35 (approved):
This looks good. Hopefully, this gets also merged upstream.

PR #564, opened 2022-08-30
Fix doumentation ref in readme

iliassoto:
Fixed some broken links in the main README.md.


mand35 (approved):
First, I thought we can use relative paths, but since they would need to point to another branch, this seems not trivial. 
Great that you fixed that issue. Thanks.

PR #574, opened 2022-09-26
Comment for column map syntax in Table.foreignKeys

iliassoto:
Added a comment to know how to write the column mapping on Table.foreignKeys


mand35 (changes requested):
Could you please also add a full example for `foreignKey` since this the one shown in the schemaViewer, The ForeignKey class is not presented there.

mand35 (changes requested):
Sorry, I guess I was not clear. My point is that in the schema viewer the class `Table` and its attributes are listed. One of the attribute is `foreignKeys`. There we now have "only" description: `optional sequence of foreign key definitions. This is used as metadata for a data catalog.` 
The description you added is very good, but as far as you not add it to the attribute description of the table class, we will not see it in the SchemaViewer.

mand35 (approved):
This look good. Many thanks for adding an example. We need to do more to improve the usability. 👍

Issue #577, opened 2022-10-21
Warn if given columns in transformation don't exist


pgruetter:
**Is your feature request related to a problem? Please describe.**
When you define a BlacklistTransformer, it is not clear if the given columnBlacklist is applied to the source DataObject (which could still be CamelCase), the internal presentation (Spark) or the output DataObject. 
It's the internal one, so it needs to be in lowercase as CamelCase column names get converted. As user, you don't get a warning or error, the BlacklistTransformer simply doesn't do anything

**Describe the solution you'd like**
Print a warning if the given columns do not exist. 
Could apply to other transformers.

ddeuber:
The only other transformer which has the same problem is the WhitelistTransformer.

PR #581, opened 2022-10-28
Feature/577 warn if column does not exist

ddeuber:
see #577


zzeekk:
Nice code and tests!
Not sure about the implementation of lowercasing. The schema gets "lowercased" but the whitelist columns not. In my opinion we should leave it to the user if he wants to lowercase column names or not (or maybe he wants to make camelCase to snake_case...). Its only a best practice to use lower case column names, but we should not force it (there might still be some other code who does, thats right...).
@pgruetter: wdyt?

zzeekk (approved):
looks good!

PR #582, opened 2022-10-28
Feature/sdl agent

Geheiner:
This Pull Request adds a new Feature allowing to have a main instance of SDLB sending instructions to another Instance of SDLB, called the Remote Agent. This Remote Agent then executes the action in it's environment and returns the schema of the resulting data (Empty SparkSubfeed) if successful. Only metadata is sent over the connection between remote Agent and Main Instance. The main instance then fetches the result if needed using it's own connection. 
This feature adds support for a new config section called agents which allows to describe how to reach the agents as well as the connections that the remote agents have.
An Action can now contain the key agentId. If it is specified, the Action will be executed on that agent, which should habe been defined in the agents section

Supported ways of connecting Main Instance to Remote Agent:
- Simple, unsecured Jetty Websocket for development use
- Azure Relay service (require sdl-azure and an Azure subscription with an Relay that has a Hybrid Connection)

Only DataObjects that return SparkSubfeeds are supported in a first step


pom.xml
```diff
@@ -857,49 +857,42 @@
 				<groupId>org.eclipse.jetty</groupId>
 				<artifactId>jetty-http</artifactId>
 				<version>${jetty.version}</version>
-				<scope>provided</scope>
```
Geheiner: TODO Probably this entire commit to be reverted before this PR can be merged!

Geheiner: Problem solved by allowing to toggle the scope via the profile fat-jar-with-spark

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentServerController.scala
```diff
@@ -0,0 +1,70 @@
...
+        val dataObjects: Map[DataObjectId, DataObject] = getDataObjectConfigMap(configFromString)
+          .map { case (id, config) => (DataObjectId(id), parseConfigObjectWithId[DataObject](id, config)) }
+        instanceRegistryImplicit.register(dataObjects)
+
+        val actions: Map[ActionId, Action] = getActionConfigMap(configFromString)
+          .map { case (id, config) => (ActionId(id), parseConfigObjectWithId[Action](id, config)) }
+
+        instanceRegistryImplicit.register(actions)
+
+        val resultingSubfeeds = sdlb.agentExec(agentServerConfig.sdlConfig, agentInstruction.phase, SDLExecutionId.executionId1, LocalDateTime.now(), LocalDateTime.now(), Seq(), Seq(), None, Seq(), simulation = false, globalConfig = GlobalConfig())(instanceRegistryImplicit)
```
ddeuber: Here you pass a large number of parameters, which makes it hard to read. Since this is the only call to agentExec, you could define default values in the method declaration or set the values directly in the method.

sdl-core/src/main/scala/io/smartdatalake/communication/message/SDLMessage.scala
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Represents a message that can be sent from a Smart Datalake Builder Instance
+ *
+ * @param msgType      The type of the message. For message type Log, log must be filled.
+ *                     For message type StatusUpdate, statusUpdate must be filled.
+ * @param statusUpdate Status update sent to the web frontend
+ * @param log          Log message sent to the web frontend
+ */
+case class SDLMessage(msgType: SDLMessageType,
```
ddeuber, line 32: Is it correct that SDLMessage is used for two unrelated applications, namely
1. status updates
2. communication between SDL agents?

If yes, I would split it into different classes for each application, so that you do not introduce coupling.

sdl-core/src/main/scala/io/smartdatalake/communication/message/SDLMessageType.scala
```diff
@@ -17,13 +17,15 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-package io.smartdatalake.statusinfo.websocket
+package io.smartdatalake.communication.message
 
 object SDLMessageType extends Enumeration {
   type SDLMessageType = Value
 
   val Log: Value = Value("Log")
```
ddeuber: SDLMessageType.Log is currently unused

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
```diff
@@ -0,0 +1,98 @@
...
+  override def outputs: Seq[DataObject] = wrappedAction.outputs
+
+  override def executionCondition: Option[Condition] = wrappedAction.executionCondition
+
+  override def executionMode: Option[ExecutionMode] = wrappedAction.executionMode
+
+  override def metricsFailCondition: Option[String] = wrappedAction.metricsFailCondition
+
+   override def prepare(implicit context: ActionPipelineContext): Unit = {
+     runOnAgent( ExecutionPhase.Prepare)
```
ddeuber: Formatting is not fully consistent in this class.

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentClient.scala
```diff
@@ -0,0 +1,92 @@
...
+}
+
+case class AgentClient(agent: Agent) {
+  val socket = new AgentClientSocket()
+  val client = new WebSocketClient
+
+  def sendSDLMessage(message: SDLMessage): Unit = {
+    val uri = URI.create(agent.url)
+
+
```
ddeuber: Two newlines seems a bit excessive :)

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentClient.scala
```diff
@@ -0,0 +1,92 @@
...
+    client.start()
+
+    val fut = client.connect(socket, uri)
```
ddeuber: `fut` is a bit a strange name, I needed a moment to figure out its meaning. `future` or `sessionFuture` would make it easier to understand

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentClientSocket.scala
```diff
@@ -0,0 +1,64 @@
...
+import org.json4s.ext.EnumNameSerializer
+import org.json4s.jackson.Serialization.{read, writePretty}
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+import scala.collection.mutable.Map
+
+class AgentClientSocket() extends WebSocketAdapter with SmartDataLakeLogger {
+
+  val pendingResults = new mutable.HashMap[String, SDLMessage]()
```
ddeuber: Why do you need a map here? Currently, there is only one action sent for each client socket. If the idea is that you should be able to send multiple messages and receive multiple responses then map should be made thread-safe.

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentServerSocket.scala
```diff
@@ -0,0 +1,76 @@
...
+    logger.info("Received TEXT message: " + message)
+    val sdlMessage = read[SDLMessage](message)
+    val responseMessageOpt = agentController.handle(sdlMessage, config)
+    if(responseMessageOpt.isDefined) sendSDLMessage(responseMessageOpt.get)
+    else closeConnection()
+  }
+
+  def sendSDLMessage(sdlMessage: SDLMessage): Unit = {
+    val outputString = writePretty(sdlMessage)
+    getSession.getRemote.sendString(outputString)
```
ddeuber: Since you log the messages you receive from the server, it would also make sense to log what you send. This could help if someone encounters network problems. The same hold for the other way around: on the server outgoing messages should be logged as well.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
```diff
@@ -0,0 +1,98 @@
...
+    agentClient.sendSDLMessage(hoconInstructions)
+
+    val instructionId = hoconInstructions.agentInstruction.get.instructionId
+
+    while (agentClient.socket.isConnected && !agentClient.socket.pendingResults.contains(instructionId)) {
+      Thread.sleep(1000)
+      println(s"Waiting for ${agent.id.id} to finish $instructionId...")
+    }
+    if(!agentClient.socket.isConnected){
+      throw new RuntimeException(s"Lost connection to ${agent.id.id}!")
```
ddeuber: This error is also thrown when a task failed on the remote agent. 

Error on remote agent:
image

Error on main / cloud SDLB:
image


This is not ideal, because the user will first have to check the log of the remote agent to find out whether something went wrong there or if there was a network problem. One idea would be to catch the exception on the remote agent and send the exception to the client.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
```diff
@@ -0,0 +1,98 @@
...
+    val hoconInstructions = AgentClient.prepareHoconInstructions(wrappedAction, context.instanceRegistry.getConnections, agent, executionPhase)
+    agentClient.sendSDLMessage(hoconInstructions)
+
+    val instructionId = hoconInstructions.agentInstruction.get.instructionId
+
+    while (agentClient.socket.isConnected && !agentClient.socket.pendingResults.contains(instructionId)) {
+      Thread.sleep(1000)
+      println(s"Waiting for ${agent.id.id} to finish $instructionId...")
+    }
+    if(!agentClient.socket.isConnected){
```
ddeuber: Jetty has a default idle timeout of 5min. After this time, the connection is closed if no message was sent:

image

We either have to disable the IdleTimeout (see https://www.eclipse.org/jetty/javadoc/jetty-10/org/eclipse/jetty/io/IdleTimeout.html#:~:text=A%20value%20that%20is%20less%20than%20or%20zero%20disables%20the%20idle%20timeout%20checks.), or add some heartbeat to keep the connection alive.

ddeuber: We should check what happens when the connection is temporarily lost. If you have long-running actions, it would not be so nice if the job crashes because you had a network instability over this time.

ddeuber (commented):
I did some local testing, I did not have time yet to try with Azure Relay. Generally, it works well, I just found some connection and error handling issues we should fix.

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentClientSocket.scala
Geheiner: Yes, you are right, we can remove the map. I anticipated that we would need it, but for now it serves no purpose. I'll remove it

sdl-core/src/main/scala/io/smartdatalake/communication/message/SDLMessage.scala
Geheiner: Yes. The original idea was to have a single Message case Class SDLMessage for all custom communication from SDLB.
But perhaps this constraint is starting to make it too complicated. @zzeekk WDYT? Should we create a new case class SDLAgentMessage which only covers Agent Communication?

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentServerSocket.scala
Geheiner: good idea, I'll add that

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentServerController.scala
Geheiner: ok, I declared them as default parameters

sdl-core/src/main/scala/io/smartdatalake/communication/message/SDLMessageType.scala
Geheiner: removed

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
Geheiner: We should discuss how far we want to implement the communication capabilities with Jetty. For me it's just the default implementation of the AgentClient, but in practice I imagine that it will rarely be used and instead we will mostly use AzureRelayAgentClient or their alternatives for Google or AWS

ddeuber: Azure Relay does not suffer from this problem, after 10minutes there is still no timeout.

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AzureRelayAgentClient.scala
```diff
@@ -0,0 +1,61 @@
...
+    val tokenProvider: TokenProvider = TokenProvider.createSharedAccessSignatureTokenProvider(connectionParams.getSharedAccessKeyName, connectionParams.getSharedAccessKey)
+    val client = new HybridConnectionClient(new URI(connectionParams.getEndpoint.toString + connectionParams.getEntityPath), tokenProvider)
+
+    val connection = client.createConnectionAsync.get
+    val messageStr = writePretty(message)(ActionDAGRunState.formats + new EnumNameSerializer(SDLMessageType) + new EnumNameSerializer(ExecutionPhase))
+    logger.info("Sending " + messageStr)
+    connection.writeAsync(ByteBuffer.wrap(messageStr.getBytes)).join()
+    val byteBuffer = connection.readAsync.get
+    // If the read operation is still pending when connection closes, the read result returns null.
+    val response = if (byteBuffer != null) {
```
ddeuber: The exception printed when the AgentServer goes down is a bit cryptic:

image

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
ddeuber: Yes, we should discuss this, because in order to use Jetty in production we will probably need to set up authentication and encryption. Regarding the original comment: Azure Relay can deal with temporarily lost connections as long as there is an active connection when the messages are sent.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
```diff
@@ -0,0 +1,87 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.workflow.action
+
+import io.smartdatalake.communication.agent.{AgentClient, AzureRelayAgentClient, JettyAgentClient}
```
ddeuber: There are unused imports.

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentServer.scala
```diff
@@ -0,0 +1,25 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.communication.agent
+
+abstract class AgentServer {
+  def start(config: JettyAgentServerConfig, serverController: AgentServerController): Unit
```
ddeuber: JettyAgentServerConfig is a bit a misnomer for implementations which do not use Jetty.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
Geheiner: I completely agree with the points regarding Jetty
I'm not sure if Encrpytion is set up with AzureRelay by default, I did not check this yet.
"as long as there is an active connection when the messages are sent." What does this mean in practice for us?

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AzureRelayAgentClient.scala
```diff
@@ -0,0 +1,61 @@
...
+import io.smartdatalake.workflow.{ActionDAGRunState, ExecutionPhase}
+import org.json4s.Formats
+import org.json4s.ext.EnumNameSerializer
+import org.json4s.jackson.Serialization.{read, writePretty}
+
+import java.net.URI
+import java.nio.ByteBuffer
+
+case class AzureRelayAgentClient() extends AgentClient with SmartDataLakeLogger {
+  implicit val format: Formats = ActionDAGRunState.formats + new EnumNameSerializer(SDLMessageType) + new EnumNameSerializer(ExecutionPhase)
```
ddeuber: This format ```ActionDAGRunState.formats + new EnumNameSerializer(SDLMessageType) + new EnumNameSerializer(ExecutionPhase)``` appears in multiple locations, I think it would make sense to extract it into a common variable.

ddeuber (commented):
Azure Relay works very well! I think the two files sdl-dependencies.txt and sdl-dependencies-2.txt can be removed.

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AgentServer.scala
Geheiner: yes, this is a leftover of the refactoring that I still need to address

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
ddeuber: It means it is fine if you have some long running job on the AgentServer and you loose connection as long as you regain the connection before sending the response. I think this is enough for us because the probability that your job crashes because there is a short connection loss is small. If there is a longer network problem there is not much you can do anyway.

sdl-core/src/main/scala/io/smartdatalake/communication/message/SDLMessage.scala
Geheiner: After discussing with @zzeekk we decided to keep SDLMessage as is. Reason: It is foreseen that in the future, the Agent will send StatusInfo messages to the main Instance via Azure Relay Connection

sdl-core/src/main/scala/io/smartdatalake/communication/agent/AzureRelayAgentClient.scala
Geheiner: I improved this exception, thanks for spotting the issue!

sdl-core/src/main/scala/io/smartdatalake/workflow/action/ProxyAction.scala
Geheiner: I improved this exception by doing as you say: catching it at the Agent, sending it to the main instance and then throwing the exception at the Agent.
 thanks for spotting the issue!

ddeuber (approved):
Looks good to me, I think its time to merge :)

PR #583, opened 2022-10-28
fix skipping actions with skipped actions as inputs

zzeekk:
If a DataObject finds no data in getDataFrame method it can throw a NoDataToProcessWarning exception. This results in the action beeing skipped. The subsequent actions should receive input SubFeed with isSkipped=true set, so that they can decide in their executionCondition if the action should be executed (default is skip if at least one input SubFeed is skipped).

This did not work as expected, as the NoDatatoProcessWarning was not enriched with SubFeeds having isSkipped=true set. This Bug resolves the issue.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/Action.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/Action.scala
@@ -190,8 +190,7 @@ private[smartdatalake] trait Action extends SdlConfigObject with ParsableFromCon
     }
     // check execution condition result
     if (!executionConditionResult.get._1 && !context.appConfig.isDryRun) {
-      val outputSubFeeds = outputs.map(output => InitSubFeed(dataObjectId = output.id, partitionValues = Seq(), isSkipped = true))
-      throw new TaskSkippedDontStopWarning(id.id, executionConditionResult.get._2.get, Some(outputSubFeeds))
+      throw new TaskSkippedDontStopWarning(id.id, executionConditionResult.get._2.get, Some(ActionHelper.createSkippedSubFeeds(outputs)))
     }
   }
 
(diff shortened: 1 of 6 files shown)


pgruetter:
Thanks a lot for this fix! I'll try to test later today.

pgruetter (changes requested):
It works as expected for default executionMode, but unfortunately I've hit a problem when trying this fix with `executionMode = { type = FailIfNoPartitionValuesMode }`.

If I have four stage actions stage_A to stage_D, only stage_D is suceeded, the rest is SKIPPED.
I configured the executionCondition to execute as long as one is SUCCEEDED. 
Triggering the subsequent action works again now, but I get the following exception during Exec phase:
`Exception in thread "main" io.smartdatalake.util.dag.TaskFailedException: Task int_validate failed. Root cause is 'AssertionError: assertion failed: (Action~int_validate) partitions run_id=7 missing for DataObject~stage_A'`

zzeekk:
Might be a problem rising from the following small improvement of this PR:
https://github.com/smart-data-lake/smart-data-lake/pull/583/files#diff-ff6f681dc3cf5b5518400dbd31f0289852680e9994c8f0872fd5e12999f233ddL237

For skipped SubFeeds checking that partition values are existing is not correct. I added the following commit to this PR to avoid it: https://github.com/smart-data-lake/smart-data-lake/pull/583/commits/b58bfd6ab6d3937474df8830aa3ab2a3b4a0ea53

pgruetter (approved):
Thanks again, now it works. Had to set SDL_ENABLE_SPARK_FILE_DATA_OBJECT_NO_DATA_CHECK=false though. 
Let's discuss later if there is a reasonable default when overwriting executionCondition.

PR #584, opened 2022-11-01
sdlPlugin - NPE because classloader is null at this point

zzeekk:
Fixing potential NPE with sdlPlugin is configured.
@pgruetter: now it works - i had the ClassNotFoundException because there was a space at the end of the class name :-(

Also found the bug with SparkFileDataObject still throwing 'IllegalStateException: (...) No FileSourceScanExec found in execution plan to check if there is data to process'. This was in case of calling getDataFrame method with partition values, but no existing files for the given partitions.


pgruetter (approved):
Thanks a lot! Oh no, the dreaded space at the end of a name...

About the other fix: If I read it correctly, I can now set remove the setting SDL_ENABLE_SPARK_FILE_DATA_OBJECT_NO_DATA_CHECK again, correct?

PR #588, opened 2022-11-04
databricks blog: add notebook config typesafe

iliassoto:
Added a small description for typesafe configuraiton from a databricks notebook


mand35 (approved):
thanks for updating

PR #592, opened 2022-11-14
fix Action with multiple inputs skipped if one input has no data (NoD…

### LINKED ISSUE #593: Action skipped if input has no data
**Describe the bug**
In version 2.3.1:
If an action has multiple inputs, the action could be skipped if one of the input DataObjects doesn't have any data. 
This happens even after explicitly setting the execution condition with an expression.

**To Reproduce**
Create a m:n action with partitioning, with one of the inputs having no data to process.

**Expected behavior**
The action gets executed.

zzeekk:
fix Action with multiple inputs skipped if one input has no data (NoDataToProcessWarning thrown in getSparkDataFrame method)


pgruetter (approved):
Tested, works correctly. Thanks a lot for the quick fix.

PR #596, opened 2022-12-02
Feature/518 wrong info in schema viewer

### LINKED ISSUE #518: Schema Viewer shows wrong information
**Describe the bug**

1. Type of WebserviceFileDataObject.partitionsDefs is strange:
http://smartdatalake.ch/json-schema-viewer/index.html#viewer-page?v=2-23-11
WebservicePartitionDefinition is actually a case class and its attribute should be rendered.

2. Definition (text for this element) is wrong, e.g. from another element?!
http://smartdatalake.ch/json-schema-viewer/index.html#viewer-page?v=2-23-1

3. Definition text should be converted to HTML to render special characters and new lines.

4. Translation section (beyond definition text) can be removed as we dont provide translations.

ddeuber:
#518 : Fix for point 2


zzeekk (approved):
Great you found the problem!

PR #607, opened 2022-12-17
Bugfix for subsequent actions skipped with DataFrameIncrementalMode

zzeekk:
### What changes are included in the pull request? Why are the changes needed?
Up to know executionCondition and executionMode was applied in Init- and Exec-Phase. As applying executionMode might be performance intensive (listing partitions, finding highwatermark), the result was cached and recalculated for specific cases. The logic to check if recalculation was needed had a bug if there were multiple inputs. In any case the recalculation was needed very often.
Solution: dont apply executionCondition and executionMode in Init-Phase, but only in Exec-Phase. This simplifies a lot, without changing the main functionality. It has an also desired side effect - early validation is no longer skipped in Init-Phase if there is no data for an Action.
With the changes of PR, Init-Phase is responsible for validating the Datapipeline, Exec-Phase checks what data needs to be processed, executes the pipeline skipping unneeded Actions.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
@@ -40,6 +40,7 @@ import scopt.OptionParser
 import java.io.File
 import java.time.{Duration, LocalDateTime}
 import scala.annotation.tailrec
+import scala.util.Try
 
 /**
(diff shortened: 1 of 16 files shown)


pgruetter (approved):
Great, works correctly now. Thanks a lot for the refactoring!
Will try to do some more regression tests tomorrow but we can merge this one already then I can test with a proper SNAPSHOT release.

Issue #610, opened 2023-01-03
Support DataObjectStateIncrementalMode for KafkaTopicDataObject


zzeekk:
**Is your feature request related to a problem? Please describe.**
Often Kafka should be read incrementally. This can currently be done using SparkStreamingMode, but it would be nice to also support  DataObjectStateIncrementalMode in order to store last loaded state in SDLB state file.

**Describe the solution you'd like**
Implement CanCreateIncrementalOutput for KafkaTopicDataObject.
Further an additional execution mode KafkaStateIncrementalMode would be interesting, which uses Kafka comitted Offsets to store it's state. This execution mode would allow SDLB to run stateless.

Issue #140, opened 2020-07-03
Kafka Topic Schema Evolution


zzeekk:
**Is your feature request related to a problem? Please describe.**
An existing Kafka topic with avro schema in schema registry might need a schema evolution if used as output and a changed schema arrives from the source. Currently the schema must be updated manually in schema registry, but often it would be nice if this is done automatically.

**Describe the solution you'd like**
The schema registry client supports updating a schema. If the new schema is compatible it should be updated automatically. There are two relevant compatibility levels to check if a schema update is compatible:
- for app internal topics, the new schema must be able to read old messages
- for app external topics, the new schema must be able to read old messages and the old schema must be able to read new messages.

**Additional context**
Some code snipptets:
```
  def setOrUpdateSchema(subject:String, newSchema:Schema): (Int,Schema) = {
    val latestSchema = if (schemaExists(subject)) Some(getLatestSchemaDefFromConfluent(subject)) else None
    val newSchemaDef = if (latestSchema.isEmpty) schemaRegistry.register(subject, newSchema)
    else if (!latestSchema.get._2.equals(newSchema)) {
      if (checkSchemaLatestReadable(latestSchema.get._2, newSchema)) {
        // schema is compatible
        registerSchema(subject, newSchema)
      } else {
        // schema is not compatible
        throw new SchemaEvolutionException(msg)
      }
    } else {
      // schema is equal
      latestSchema.get
    }
    // return
    newSchemaDef
  }
  
  // check if the data produced by the latest schema can be read by the new schema
  // this is needed to do schema evolution for internal topics, as internal readers will be updated at the same time
  def checkSchemaLatestReadable(latestSchema:Schema, newSchema:Schema): Boolean = {
    logger.debug("schemas", "latest"->latestSchema, "new"->newSchema)
    //SchemaCompatibility.checkReaderWriterCompatibility(newSchema, latestSchema).getType == SchemaCompatibility.SchemaCompatibilityType.COMPATIBLE
    val validatorRead = new SchemaValidatorBuilder().canReadStrategy.validateLatest
    Try{validatorRead.validate(newSchema, Seq(latestSchema).asJava)}.isSuccess
  }

  // check if the data produced by the new schema could be read by the latest schema and vice versa
  // this is needed to do schema evolution for external topics, because external readers might not update at the same time
  private def checkSchemaMutualReadable(latestSchema:Schema, newSchema:Schema): Boolean = {
[...]

PR #611, opened 2023-01-03
Feature/610 kafka incremental

zzeekk:
#610: 
- implementation of DataObjectStateIncrementalMode for Kafka
- Refactoring of ExecutionMode configuration, so that every module can define its own ExecutionModes
- implementation of KafkaStateIncrementalMode

#140:
- implementation of Schema Evolution for Kafka Topics with Schema Registry, e.g. checks that schema update is backward compatible


sdl-core/src/main/scala/io/smartdatalake/workflow/action/executionMode/CustomPartitionMode.scala
```diff
@@ -0,0 +1,71 @@
...
+}
+
+object CustomPartitionMode extends FromConfigFactory[ExecutionMode] {
+  override def fromConfig(config: Config)(implicit instanceRegistry: InstanceRegistry): CustomPartitionMode = {
+    extract[CustomPartitionMode](config)
+  }
+}
+
+trait CustomPartitionModeLogic {
+  def apply(options: Map[String,String], actionId: ActionId, input: DataObject with CanHandlePartitions, output: DataObject with CanHandlePartitions, givenPartitionValues: Seq[Map[String,String]], context: ActionPipelineContext): Option[Seq[Map[String,String]]]
```
ddeuber, line 85: Some Scaladoc for the traits and the apply function would be helpful when implementing custom subclasses. For example, I still don't understand the purpose of `givenPartitionValues` and I needed some time to figure out that the actual partition values can be found in `dataObject.listPartitions`. The same holds for the trait `CustomModeLogic`.

sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
```diff
@@ -481,22 +527,48 @@ private case class TopicPartitionOffsets(topicPartition: TopicPartition, startOf
...
-  private def getOffsetsForSpark(tp: TopicPartition, ot: Option[Long], defaultOffset: Int): String = {
+  def getOffsetForSpark(partition: Int, offset: Option[Long], defaultOffset: Int): String = {
     // if partition is empty we get no offset, but we have to define one for spark. Define defaultOffset for this.
-    s""""${tp.partition}":${ot.getOrElse(defaultOffset)}"""
+    s""""${partition}":${offset.getOrElse(defaultOffset)}"""
+  }
+
+  /**
+   * Parse starting/endingOffset used for Spark Kakfa data source from string
+   * Expected format: "<partitionNb:integer>:<offset:long>"
```
ddeuber: Typo in the comment: the expected format should be `"<partitionNb:integer>":<offset:long>`.

sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
```diff
@@ -481,22 +527,48 @@ private case class TopicPartitionOffsets(topicPartition: TopicPartition, startOf
     Seq.fill(size)(this.copy(startOffset = endOffset))
   }
 
+  def getStartOffsetForSpark: String = getOffsetForSpark(topicPartition.partition, startOffset, TopicPartitionOffsets.defaultOffsetEarliest)
```
ddeuber, line 585: Is there a way to set another offset for the first execution in the configuration? I could imagine some use cases where you do not want to load the whole data in Kafka (e.g. after renaming the consumer group).

ddeuber (commented):
LGTM, I only have some minor remarks.

sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
zzeekk: Good hint. I would delegate this to the system managing the state - in case of KafkaStateIncrementalMode the current offsets for a consumer group are managed by Kafka. You can use `kafka-consumer-groups` tool to change the current offsets, see also https://stackoverflow.com/questions/29791268/how-to-change-start-offset-for-topic.
In case of DataObjectStateIncrementalMode you would need to manipulate SDLB state files.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/executionMode/CustomPartitionMode.scala
zzeekk: done

sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
zzeekk: fixed

ddeuber (approved):

Issue #601, opened 2022-12-09
Loading Schema from file should be done lazy


zzeekk:
**Describe the bug**
Loading DataObjects schema from files (XSD or JsonSchema) is done when parsing the configuration. This means it is done for all DataObjects regardless of if they are selected for execution or not. This approach has two problems: it slows down parsing the configuration, and it might fail on environments without access to these files.

**To Reproduce**
  stg-test-report {
    type = XmlFileDataObject
    path = data/stg-test-report
    schema = "xsdfile#data/stg-test-report/sample.xsd;TestReport"
    rowTag = TestReport
  }

**Expected behavior**
The schema file should only be read in prepare phase of a DataObject, e.g the variable should be made "lazy" and read in prepare phase.

zzeekk:
Can be enabled by setting Environment.parseSchemaFilesLazy & compileScalaCodeLazy. Default for both is false.
Environment settings can now also be set by using global.environment section in configuration files. Like that they can be set easily in the configuration for specific environments.

PR #612, opened 2023-01-03
Bugfix/601 loading schemas lazy

zzeekk:
see #601, includes some refactoring in Environment.scala so that environment settings can also be read from global config (only system properties and environment variables for now)

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
@@ -111,7 +111,7 @@ object Environment {
    * Set path level of user home to determine basedir automatically (Default=2 -> /user/myUserHome)
    */
   def hdfsAclsUserHomeLevel: Int = {
-    if (_hdfsAclsLimitToBasedir.isEmpty) {
+    if (_hdfsAclsUserHomeLevel.isEmpty) {
       _hdfsAclsUserHomeLevel = Some(
         EnvironmentUtil.getSdlParameter("hdfsAclsUserHomeLevel")
          .map(_.toInt).getOrElse(2)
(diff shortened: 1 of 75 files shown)


sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
```diff
@@ -50,90 +51,150 @@ object Environment {
...
+  var _hdfsAclsLimitToBasedir: Option[Boolean] = None
 
   /**
    * Set path level of user home to determine basedir automatically (Default=2 -> /user/myUserHome)
    */
-  var hdfsAclsUserHomeLevel: Int = {
-    EnvironmentUtil.getSdlParameter("hdfsAclsUserHomeLevel")
-      .map(_.toInt).getOrElse(2)
+  def hdfsAclsUserHomeLevel: Int = {
+    if (_hdfsAclsLimitToBasedir.isEmpty) {
```
pgruetter:
```suggestion
    if (_hdfsAclsUserHomeLevel.isEmpty) {
```

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/DataObject.scala
```diff
@@ -59,6 +60,24 @@ trait DataObject extends SdlConfigObject with ParsableFromConfig[DataObject] wit
...
+    // check lazy parsed schema
+    this match {
+      case x: UserDefinedSchema => try {
+        x.schema.foreach(_.columns)
+      } catch {
+        case e: Exception => throw ConfigurationException.fromException(s"($id) error parsing 'schema'", "schema", e)
+      }
+      case _ => Unit
+    }
+    // check lazy parsed schemaMin (note that it can match schema and schemaMin, an we need therefore two match statements)
```
pgruetter:
```suggestion
    // check lazy parsed schemaMin (note that it can match schema and schemaMin, and we therefore need two match statements)
```

pgruetter (changes requested):
Looks good in general, but I think there's one copy-paste error that needs to be corrected.

pgruetter (approved):

Issue #600, opened 2022-12-09
Schema viewer does not indicate whether a field is required


ddeuber:
The schema viewer does not indicate whether a field is required although it is described in the legend.

An example of a required field is JsonFileDataObject.path:

![image](https://user-images.githubusercontent.com/10706639/206730484-bfa97eb2-511c-42f9-b4f3-acda8491455f.png)

ddeuber:
I noticed that the field `id` is marked as required in the JSON schema, so in the end we would have

![image](https://user-images.githubusercontent.com/10706639/206745792-5e3b5d4f-d9f4-4d15-bfa3-5f7eed58506f.png)

This might be counterintuitive because the user might think he needs to set the `id` inside the object but it is already defined as the key of the object. At least we should mention in the description of the `id` that this is in fact the key used for the object. WDYT @zzeekk?

ddeuber:
We determined that the `id` should not be part of the JSON schema.

PR #591, opened 2022-11-11
589 column encryption

mand35:
### What changes are included in the pull request?
Add a new transformer, which encrypts a list of columns in the dataFrame. 
Furthermore, the counterpart, the decryptor is implemented. Both use java cipher with. 

### Why are the changes needed?
Data security becomes more and more important. Often very sensitive data is mixed with data necessary to be analysed. Therefore, we need to hide the sensitive data by encrypting it

### Aspects to consider and to be discussed
We use a GCM method with initial vector (IV), which is stored at the beginning of the encrypted message. Therewith cells with the same value result in different ciphertexts.

Issue #580, opened 2022-10-25
Can't use same ExcelFileDataObject for write and read


pgruetter:
**Describe the bug**
See the example below. Reading an Excel from a WebserviceFileDataObject creates a file called result.xlsx in the specified path. 
When reading this same ExcelFileDataObject with a second action, the following error is thrown:

> Exception in thread "main" io.smartdatalake.util.dag.TaskFailedException: Task int-bfs-commune-register failed. Root cause is 'FileNotFoundException: C:\<path-to>\bfs-commune-register (Access is denied)'
> Caused by: java.io.FileNotFoundException: C:\<path-to>bfs-commune-register (Access is denied)
> 	at java.base/java.io.FileInputStream.open0(Native Method)
> 	at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
> 	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
> 	at org.apache.hadoop.fs.RawLocalFileSystem$LocalFSFileInputStream.<init>(RawLocalFileSystem.java:146)
> 	at org.apache.hadoop.fs.RawLocalFileSystem.open(RawLocalFileSystem.java:275)
> 	at org.apache.hadoop.fs.ChecksumFileSystem$ChecksumFSInputChecker.<init>(ChecksumFileSystem.java:161)
> 	at org.apache.hadoop.fs.ChecksumFileSystem.open(ChecksumFileSystem.java:372)
> 	at org.apache.hadoop.fs.FileSystem.open(FileSystem.java:976)
> 	at com.crealytics.spark.excel.WorkbookReader$.readFromHadoop$1(WorkbookReader.scala:60)

If you manually change the path (only for reading the file in int_*) to "bfs-commune-register/result.xlsx", the exception changes to:

> Exception in thread "main" io.smartdatalake.util.dag.TaskFailedException: Task int-bfs-commune-register failed. Root cause is 'IllegalStateException: (stg-bfs-commune-register) No FileSourceScanExec found in execution plan to check if there is data to process'
> Caused by: java.lang.IllegalStateException: (stg-bfs-commune-register) No FileSourceScanExec found in execution plan to check if there is data to process
> 	at io.smartdatalake.workflow.dataobject.SparkFileDataObject$.$anonfun$getFilesProcessedFromSparkPlan$2(SparkFileDataObject.scala:425)
> 	at scala.Option.getOrElse(Option.scala:189)
[...]

PR #616, opened 2023-01-14
Release 2.4.0

pgruetter:
## Bugfixes and improvements
#518 Schema Viewer shows wrong information
#580 Can't use same ExcelFileDataObject for write and read
#600 Schema viewer does not indicate whether a field is required
#601 Loading Schema from file should be done lazy
Leading underscores are preserved when normalizing column names
ExecutionMode and executionCondition are only applied in exec phase

## Features 
#591 Column encryption
#610 Support DataObjectStateIncrementalMode for KafkaTopicDataObject

## Dependencies
Bump commons-net from 3.1 to 3.9.0


pgruetter:
@zzeekk I would argue that the list of changes would warrant a version bump to 2.4.0. WDYT ?

zzeekk:
Yes, we should bump to version 2.4.0, because some classes have been relocated which might be used by some installation (CustomPartitionModeLogic:  io.smartdatalake.definitions -> io.smartdatalake.workflow.action.executionMode=

pgruetter:
@zzeekk : There you go: https://github.com/smart-data-lake/smart-data-lake/pull/617

zzeekk (approved):

PR #624, opened 2023-02-24
Minor bugfixes and improvements

zzeekk:
- fix dbutils-api scala version with property scala.minor.version
- improve exception message if key of DataObject or Connection is not found
- implement SQLQueryExpectation
- general improvements for Expectations
- fix NPE in ExpectationDefaultImpl
- make fail simulation on missing inputSubFeeds configurable
- make more from sdl-core public

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
@@ -312,7 +312,6 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
    * Similar to [[startSimulation]], but Actions and DataObjects are parsed from config files defined through appConfig.configuration attribute.
    */
   def startSimulationWithConfigFile(appConfig: SmartDataLakeBuilderConfig, initialSubFeeds: Seq[SparkSubFeed], dataObjectsState: Seq[DataObjectState] = Seq(), failOnMissingInputSubFeeds: Boolean = true)(session: SparkSession): (Seq[DataFrameSubFeed], Map[RuntimeEventState,Int]) = {
-    Environment._failSimulationOnMissingInputSubFeeds = Some(failOnMissingInputSubFeeds)
     val config = ConfigLoader.loadConfigFromFilesystem(appConfig.configuration.get, session.sparkContext.hadoopConfiguration)
     ConfigParser.parse(config, this.instanceRegistry)
     startSimulation(appConfig, initialSubFeeds, dataObjectsState, failOnMissingInputSubFeeds)(this.instanceRegistry, session)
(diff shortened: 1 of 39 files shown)


sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -288,32 +288,37 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
...
     (subFeeds.map(_.asInstanceOf[SparkSubFeed]), stats)
   }
 
   /**
-   * Starts a simulation run and registers all SDL first class objects that are defined in the config file which path is defined in parameter appConfig
+   * Similar to [[startSimulation]], but Actions and DataObjects are parsed from config files defined through appConfig.configuration attribute.
    */
-  def startSimulationWithConfigFile(appConfig: SmartDataLakeBuilderConfig, initialSubFeeds: Seq[SparkSubFeed], dataObjectsState: Seq[DataObjectState] = Seq())(session: SparkSession): (Seq[SparkSubFeed], Map[RuntimeEventState,Int]) = {
+  def startSimulationWithConfigFile(appConfig: SmartDataLakeBuilderConfig, initialSubFeeds: Seq[SparkSubFeed], dataObjectsState: Seq[DataObjectState] = Seq(), failOnMissingInputSubFeeds: Boolean = true)(session: SparkSession): (Seq[DataFrameSubFeed], Map[RuntimeEventState,Int]) = {
+    Environment._failSimulationOnMissingInputSubFeeds = Some(failOnMissingInputSubFeeds)
```
ddeuber: Why do you need to set `Environment._failSimulationOnMissingInputSubFeeds` here when it is set again in `startSimulation`?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/Expectation.scala
```diff
@@ -133,6 +133,7 @@ abstract class ExpectationDefaultImpl extends Expectation {
     import org.apache.spark.sql.functions._
     expectation.map { expectationStr =>
       val valueStr = value match {
+        case null => "null"
```
ddeuber: Are there other possibilities than passing null around in the metrics? I get paranoid when I see null values in Scala :)

ddeuber (approved):

sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
zzeekk: Correct, i removed it.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/Expectation.scala
zzeekk: You're right, not nice in Scala.
Background: the null value is from a Spark df.collect.
I fixed it and convert null to None.

PR #628, opened 2023-03-03
Feature/spark lab

zzeekk:
This PR adds a SmartDataLakeBuilderLab class to the code base, which servers as entry point for using SDLB objects interactively.
For now it supports accessing Spark DataObjects with a simplified interface, including code completion for selecting a DataObject from the list, like a catalog.

This "Catalog" is created by code generation after the regular compilation of an SDLB project. The LabCatalogGenerator command line tool is responsible for that. In Maven you need to add maven-exec-plugin and additional compile phase afterwards, see also scaladoc of LabCatalogGenerator for configuration of the corresponding Maven Plugins.

The SDLB projects needs sdl-lang as addtional dependency.


Geheiner (approved):
Looks good to me!
Amazing how much we can do with the framework with so little changes

sdl-lang/src/main/scala/io/smartdatalake/lab/SmartDataLakeBuilderLab.scala
```diff
@@ -39,3 +39,11 @@ case class SmartDataLakeBuilderLab[T](
   @transient val context: ActionPipelineContext = ConfigToolbox.getDefaultActionPipelineContext(session, registry)
   @transient val data: T = dataObjectCatalogFactory(registry, context)
 }
+
+object SmartDataLakeBuilderLab {
+  /**
+   * Writing into DataObjects using SmartDataLakeBuilderLab is disabled by default, as this should be done through data pipelines and not interactively.
+   * It can be enabled by setting the following variable to true.
+   */
+  var enableWritingDataObjects = false
```
Geheiner, line 48: OK now I'm a bit confused on how you use this new feature.
I thought the idea was to generate the class files with the CLI.
What is the use case that you satisfy when you set this property to true? You may want to add the answer as documentation in the code

zzeekk: This setting is used at Runtime, not for generating the Catalog class file.
We discussed today in the team at a customer. There was a concern that by making it ease to write to a DataObject interactively (e.g. from Polynote), we will loose on governance, writing new data should be done by SDLB, e.g. executing Actions. Nevertheless we agreed that it should not be removed completely, but disabled by default...

PR #631, opened 2023-03-09
131 create an action to apply a ml model

### LINKED ISSUE #131: Create an Action to apply a ML model
As a data scientist i would like to have an Action to apply a ML model inside my data pipeline.

New MLFlowAction which applies a MLFLow model to the data. The model to apply is defined by the hadoop path were it is stored/deployed.

**Additional context**
One of the more popular generic ML model registries is MLFlow. MLFlow has a Python interface to create a Spark user-defined-function, which can be used in PySpark to apply a model to data:
```
predict = mlflow.pyfunc.spark_udf(spark, "/my/local/model")
df.withColumn("prediction", predict("name", "age")).show()
```
Preconditions: Custom transformations in Python are implemented (#130)

https://www.mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#mlflow.pyfunc.spark_udf

hiltpold:
* The framework was extended with a MLflowTrainAction and a MLflowPredictAction. Provided with the correct python code to train a sklearn model and the labelled data, the MLflowTrainAction will apply the python code and log automatically to a running MLflow instance. If a model has already been trained, the MLflowPredictAction applies the model to input data and outputs the inferred data.
*  For this to work a PythonCodeSparkDfsTransformer was added. This allows for n:m transformations using python.
* The MLflowPythonUtil is a proposal how to interact with the MLflow instance using the python API

Here a config snippet:

```
  train {
    type = MLflowTrainAction
    mlflowId = mlflow
    runInfoId = mlflow-run-info
    trainIdSelector = "labels"
    inputIds = [labels]
    pythonModelFile = "./models/fancy_model.py"
    registerModel: true
    metadata.feed = ml-train
  }

  predict {
    type = MLflowPredictAction
    mlflowId = mlflow
    outputId = prediction
    resultType = "double"
    predictIdSelector = "predict"
    inputIds = [data2predict]
    metadata.feed = ml-predict
  }
```

Great idea to combine a data engineering framework with some ML features.


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/MLflowDataObject.scala
```diff
@@ -0,0 +1,100 @@
...
+                             modelName: Option[String] = None,
+                             modelDescription: Option[String] = None,
+                             override val metadata: Option[DataObjectMetadata] = None
+                           )(@transient implicit val instanceRegistry: InstanceRegistry)
+  extends DataObject
+    with CanCreateSparkDataFrame
+    with SmartDataLakeLogger {
+
+  // entry point for accessing dynamically Java objects living inside the JVM
+  var entryPoint: Option[MLflowPythonSparkEntryPoint] = None
```
zzeekk: When accessing this DataObject from several Actions we might become problems with entryPoint and pythonMLFlowApi variables. It would be more logical to move this "state" to the action. We just let the MLFlowDataObject create pythonMLflowApi (or better pythonMLflowClient?) by a method get MLFlowClient, and keep the object in the the MLFlow*Action

sdl-core/src/main/scala/io/smartdatalake/workflow/action/MLflowTrainAction.scala
```diff
@@ -0,0 +1,258 @@
...
+      .getOrElse(throw ConfigurationException(s"ML model must either be provided as python file or code"))
+  }
+
+  // handle MLflow in and output data object
+  val mlflow = getInputDataObject[MLflowDataObject](mlflowId)
+  val mlflowRunInfo = getOutputDataObject[DataObject with CanWriteDataFrame with CanCreateDataFrame](runInfoId)
+
+  // handle data in- and outputs
+  override val inputs: Seq[DataObject with CanCreateDataFrame] =
+    inputIds.map(getInputDataObject[DataObject with CanCreateDataFrame]) :+ mlflow
```
zzeekk, line 95: As we are writing an MLFLow model, i suggest that mlflow DataObject should be an output, not input. But what DataFrame could we write there?
Maybe we should implement CanReceiveScriptNotification instead of CanWriteDataFrame, but then it could no longer extend DataFrameActionImpl.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/MLflowPredictAction.scala
```diff
@@ -0,0 +1,212 @@
...
+             |
+             |mlflow.set_tracking_uri("${mlflow.trackingURI}")
+             |# set experiment name
+             |mlflow.set_experiment("${mlflow.experimentName}")
+             |# get predict dataframe from SDL
+             |predict_id = get_predict_id("$predictIdSelector", list(inputDfs.keys()))
+             |print(f"The following DataObjectId will be used for training: {predict_id}")
+             |df = inputDfs[f"{predict_id}"]
+             |# add predictions; TODO: get correct type according to input
+             |df_predict = df.withColumn("predictions", lit(None).cast(StringType()))
```
zzeekk: To allow proper validation in init phase, the model should already be loaded there. I expect that the downloaded model is cached, as there is a SparkModelCache in mlflow.
We could make a switch to disable loading the model in init phase for performance reasons.

zzeekk (commented):
Further notes:
- All logs and exceptions should be prefixed with "(DataObject~<name>)..." or "(Action~<name>)..."
- Can we create a switch to disable autolog? This could help reducing funtionality in case of problems.
- There is a WARNING mlflow.pyfunc: Calling `spark_udf()` with `env_manager="local"`. Should we make env_manager configurable?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/MLflowDataObject.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  override def getSparkDataFrame(
+                                  partitionValues: Seq[PartitionValues] = Seq()
+                                )(implicit context: ActionPipelineContext): DataFrame = {
+    implicit val session = context.sparkSession
+    import session.implicits._
+
+    // TODO: fetch latest run information and return as DataFrame
+    val df: DataFrame = Seq.empty[MLflowExperiment].toDF()
```
zzeekk, line 91: If we return run information here, can we use this input DataFrame in MLFlowPredictAction to choose the run version to be used for prediction? We could then have some Spark expression configured for MLFlowPredictAction to choose one run from the DataFrame...

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/MLflowDataObject.scala
hiltpold: Good point. Renamed the variable and moved the `entryPoint` to the Action.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/MLflowPredictAction.scala
hiltpold: done.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/MLflowTrainAction.scala
hiltpold: Good catch. In addition, if we set the `MLflowDataObejct` as output, it does not even necessarily have to implement the `CanWriteparkDataFrame` trait. It could only write a Notification that the training is finished. But somehow this switch between Script and DF is not well defined yet.
Btw. the run information is written to `runInfo` DataObject.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/MLflowDataObject.scala
hiltpold: This contradicts a bit the previous comment. Run information is written to the 'runInfo' DataObject. So yes, this could be used for input in the `MLFlowPredictAction`.

pgruetter:
As discussed, let's keep the branch but since it's a draft implementation, we will abandon the PR for now.

PR #637, opened 2023-03-24
Improve JDBC transaction handling

### LINKED ISSUE #635: Disable JDBC autoCommit per default
Many JDBC drivers use auto-commit mode per default. This is not recommended in combination with explicit `commit()` calls which SDLB uses for transactional behaviour. For some JDBC drivers (e.g. Oracle), this even leads to exceptions. It is possible to disable auto-commit mode by setting `JdbcTableConnection.autoCommit` to `false`, but the default value for `autoCommit` is `None` for backwards compatibility, in which case the JDBC default commit mode is used.

We should set `autoCommit` to `false` per default to ensure that the transactional behaviour of SDLB is correct. However, we should check if this requires additional `commit()` calls in SDLB. There might be JDBC SQL statements which have only been committed because of the auto-commit mode. This might be especially problematic if `JdbcTableConnection.maxParallelConnections > 1`.

ddeuber:
See #635 

This PR includes several improvements for the JDBC transaction handling:

* A new class `JdbcTransaction` inside `JdbcConnection` which reserves a database connection for the active transaction and releases it on commit or rollback. This makes sure we always use the same connection even if the connection pool contains multiple ones and that no other part of the application interferes with the transaction.
* `JdbcConnection.execJdbcStatement` and `JdbcConnection.execJdbcStatement` always commit after the execution. Because Spark writes to the database using separate connections, it is hard to guarantee end-to-end transactional behaviour in `JdbcTableDataObject`. In particular, depending on the isolation level of the database, changes from Spark connections might not be visible inside the transaction if the transaction is started before the Spark job. Therefore, we commit all statements or queries immediately, except those managed by `JdbcTransaction`. This was also the default behaviour before when the JDBC driver was set to `autoCommit = true` per default.
* The property `JdbcConnection.autoCommit` is now `false` per default and it should not be necessary for a user to set it to `true`. I think it requires too much knowledge of how SDLB works for a user to set `autoCommit` and now that the transaction handling has been improved it is obsolete. Thus, it is now deprecated and should be removed in a future release.
* The property `JdbcConnection.enableCommit` has been removed. Its behaviour has been replaced by `!autoCommit`. Since `enableCommit` was only introduced in version 2.4.2 and most people probably use `autoCommit` instead, I feel it is ok to remove it although it is not backwards compatible.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
@@ -36,6 +36,7 @@ import org.apache.spark.sql.jdbc.JdbcDialects
 import org.apache.spark.sql.types.StructType
 
 import java.sql.{DriverManager, ResultSet, Statement, Connection => SqlConnection}
+import java.time.Duration
 
 /**
  * Connection information for jdbc tables.
@@ -50,11 +51,9 @@ import java.sql.{DriverManager, ResultSet, Statement, Connection => SqlConnectio
  *                               Note that Spark manages JDBC Connections on its own. This setting only applies to JDBC connection
  *                               used by SDL for validating metadata or pre/postSQL.
  * @param connectionPoolMaxIdleTimeSec timeout to close unused connections in the pool
- * @param enableCommit whether to enable commits for transactional behaviour in SDL
- *                     Note that most JDBC drivers use auto-commit per default which is not recommended when [[enableCommit]] is true.
- *                     Default is enableCommit = true. To disable auto-commit see [[autoCommit]].
- * @param autoCommit flag to enable or disable the auto-commit behaviour of the JDBC driver
- *                   If not set, the default auto-commit mode of the JDBC driver is used.
+ * @param connectionPoolMaxWaitTimeSec timeout when waiting for connection in pool to become available. Default is 60 seconds.
+ * @param autoCommit flag to enable or disable the auto-commit behaviour. When autoCommit is enabled, each database request is executed in its own transaction.
+ *                   Default is autoCommit = false. It is not recommended to enable autoCommit as it will deactivate any transactional behaviour.
  * @param connectionInitSql SQL statement to be executed every time a new connection is created, for example to set session parameters
  */
 case class JdbcTableConnection(override val id: ConnectionId,
@@ -64,9 +63,9 @@ case class JdbcTableConnection(override val id: ConnectionId,
                                db: Option[String] = None,
                                maxParallelConnections: Int = 1,
                                connectionPoolMaxIdleTimeSec: Int = 3,
+                               connectionPoolMaxWaitTimeSec: Int = 60,
                                override val metadata: Option[ConnectionMetadata] = None,
-                               enableCommit: Boolean = true,
-                               autoCommit: Option[Boolean] = None,
+                               @Deprecated @deprecated("Enabling autoCommit is no longer recommended.", "2.5.0") autoCommit: Boolean = false,
                                connectionInitSql: Option[String] = None
                                ) extends Connection with SmartDataLakeLogger {
 
@@ -86,32 +85,26 @@ case class JdbcTableConnection(override val id: ConnectionId,
     }
   }
 
-  /**
-   * Get a JDBC connection from the pool, create a JDBC statement and execute an arbitrary function
-   */
-  def execWithJdbcStatement[A](doCommit: Boolean = false)(func: Statement => A ): A = {
-    execWithJdbcConnection { conn =>
-      var stmt: Statement = null
-      try {
-        stmt = conn.createStatement
-        val result = func(stmt)
-        if (doCommit && enableCommit) conn.commit()
-        result
-      } finally {
-        if (stmt != null) stmt.close()
-      }
+  private def execWithJdbcStatement[A](conn: SqlConnection, doCommit: Boolean)(func: Statement => A): A = {
+    var stmt: Statement = null
+    try {
+      stmt = conn.createStatement
+      val result = func(stmt)
+      if (doCommit && !autoCommit) conn.commit()
+      result
+    } finally {
+      if (stmt != null) stmt.close()
     }
   }
 
   /**
-   * Execute an SQL statement
-   * @return true if the first result is a ResultSet object; false if it is an update count or there are no results
+   * Get a JDBC connection from the pool, create a JDBC statement and execute an arbitrary function
    */
-  def execJdbcStatement(sql:String, logging: Boolean = true, doCommit: Boolean = false) : Boolean = {
-    execWithJdbcStatement(doCommit) { stmt =>
+  def execJdbcStatement(sql:String, logging: Boolean = true) : Boolean = {
+    execWithJdbcConnection(execWithJdbcStatement(_, doCommit = true) { stmt =>
       if (logging) logger.info(s"execJdbcStatement: $sql")
       stmt.execute(sql)
-    }
+    })
   }
 
   /**
@@ -121,7 +114,7 @@ case class JdbcTableConnection(override val id: ConnectionId,
    * @return the evaluated result
    */
   def execJdbcQuery[A](sql:String, evalResultSet: ResultSet => A ) : A = {
-    execWithJdbcStatement() { stmt =>
+    execWithJdbcConnection(execWithJdbcStatement(_, doCommit = true) { stmt =>
       var rs: ResultSet = null
       try {
         logger.info(s"execJdbcQuery: $sql")
@@ -130,7 +123,7 @@ case class JdbcTableConnection(override val id: ConnectionId,
       } finally {
         if (rs != null) rs.close()
       }
-    }
+    })
   }
 
   def test(): Unit = {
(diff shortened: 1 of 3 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -50,11 +51,9 @@ import java.sql.{DriverManager, ResultSet, Statement, Connection => SqlConnectio
  *                               Note that Spark manages JDBC Connections on its own. This setting only applies to JDBC connection
  *                               used by SDL for validating metadata or pre/postSQL.
  * @param connectionPoolMaxIdleTimeSec timeout to close unused connections in the pool
- * @param enableCommit whether to enable commits for transactional behaviour in SDL
- *                     Note that most JDBC drivers use auto-commit per default which is not recommended when [[enableCommit]] is true.
- *                     Default is enableCommit = true. To disable auto-commit see [[autoCommit]].
- * @param autoCommit flag to enable or disable the auto-commit behaviour of the JDBC driver
- *                   If not set, the default auto-commit mode of the JDBC driver is used.
+ * @param connectionPoolMaxWaitTimeSec timeout when waiting for connection in pool to become available. Default is 60 seconds.
```
zzeekk: This means that after 60s waiting for a connection, an exception is thrown. I'm not sure about this behaviour, because we use maxParallelConnections also to throttle the number of connections created when accessing a database to limit the load on the database. In this case we want to wait until a connection gets available, and not abort when there is none within 60s. Maybe we could increase the default to 10m?

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -199,7 +192,9 @@ case class JdbcTableConnection(override val id: ConnectionId,
   val pool = new GenericObjectPool[SqlConnection](new JdbcClientPoolFactory)
   pool.setMaxTotal(maxParallelConnections)
   pool.setMaxIdle(1) // keep max one idle jdbc connection
```
zzeekk: I studied setMaxIdle for SFtpFileRefConnection pool, and came to the conclusion that we should remove setMaxIdle, as connections will already be terminated by setMinEvictableIdle after some amount of time. If we additionally setMaxIdle, it means that if there is already an idle connection and another one arrives, it will be terminated immediately. In my opinion the second connection could also be kept open until setMinEvictableIdle is reached

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -64,9 +63,9 @@ case class JdbcTableConnection(override val id: ConnectionId,
                                db: Option[String] = None,
                                maxParallelConnections: Int = 1,
```
zzeekk: I think we should increase the default number of parallel connections maybe 10. If we have only 1 parallel connection there are chances for dead locks when reading & writing 2 DataObjects that use the same Connection.

sdl-core/src/test/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnectionTest.scala
```diff
@@ -21,21 +21,86 @@ package io.smartdatalake.workflow.connection.jdbc
...
+    // prepare
+    val jdbcConnection = JdbcTableConnection("jdbcCon1", "jdbc:hsqldb:file:target/JdbcTableDataObjectTest/hsqldb",
+      "org.hsqldb.jdbcDriver", maxParallelConnections = 2)
+
+    // run
+    val transaction1 = jdbcConnection.beginTransaction()
+    val transaction2 = jdbcConnection.beginTransaction()
+
+    // check
+    // no exception
```
zzeekk, line 82: If we dont close these transactions, are there no side-effects by hsqldb in the unit tests? We have connections from several test cases open on the same file.
I suggest to rollback/commit these transactions.

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -216,14 +211,66 @@ case class JdbcTableConnection(override val id: ConnectionId,
...
+   */
+  def beginTransaction(): JdbcTransaction = {
+    new JdbcTransaction(this)
+  }
+
+  /**
+   * Class for handling database transactions. If all operations succeeded call [[commit]], otherwise [[rollback]].
+   */
+  class JdbcTransaction(connection: JdbcTableConnection) extends SmartDataLakeLogger {
+    logger.info(s"begin transaction [$id]")
```
zzeekk:
```suggestion
    logger.info(s"($id) begin transaction")
```

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -216,14 +211,66 @@ case class JdbcTableConnection(override val id: ConnectionId,
...
+    private val jdbcConnection: SqlConnection = connection.pool.borrowObject()
+
+    def execJdbcStatement(sql:String, logging: Boolean = true) : Boolean = {
+      connection.execWithJdbcStatement(jdbcConnection, doCommit = false) { stmt =>
+        if (logging) logger.info(s"execJdbcStatement in transaction [$id]: $sql")
```
zzeekk:
```suggestion
        if (logging) logger.info(s"($id) execJdbcStatement in transaction: $sql")
```

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -216,14 +211,66 @@ case class JdbcTableConnection(override val id: ConnectionId,
...
+        if (autoCommit) logger.warn("autoCommit is enabled, so statement will be committed immediately")
+        stmt.execute(sql)
+      }
+    }
+
+    def commit(): Unit = {
+      if (!connection.autoCommit) {
+        logger.info(s"commit transaction [$id]")
```
zzeekk:
```suggestion
        logger.info(s"($id) commit transaction")
```

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
```diff
@@ -216,14 +211,66 @@ case class JdbcTableConnection(override val id: ConnectionId,
...
+        jdbcConnection.commit()
+      };
+      close()
+    }
+
+    def rollback(): Unit = {
+      try {
+        if (!connection.autoCommit) {
+          logger.info(s"roll back transaction [$id]")
```
zzeekk:
```suggestion
          logger.info(s"($id) roll back transaction")
```

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
zzeekk: Had another thought on this - actionally JdbcTableConnections handles only queries initiated from SDLB. But the connections transferring the data are created by Spark and parallelization is configured by Spark properties. In this case i think that deadlocks should not happen also if we have only 1 connection, as we dont have simultaneous read/write. But we could nevertheless increase the default a little to maybe 3? WDYT?

zzeekk (commented):
Good improvement of transaction handling!
I have some other remarks to discuss and minor suggestions for formatting log entries according to SDLB standards.

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
ddeuber: I added `connectionPoolMaxWaitTimeSec` because otherwise it is hard to detect that the application hangs because it waits for a connection and that `maxParallelConnections` should be increased. In addition, if due to an error in the code the connection is not returned to the pool correctly, the application might hang indefinitely without logging why. We can increase the default to 10m because waiting a few minutes more or less is usually not a problem for SDLB.

sdl-core/src/test/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnectionTest.scala
ddeuber: I suspect that the connection is garbage collected after the test and therefore closed, but rollback/commit is definitely better.

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
ddeuber: Increasing to 3 should not do any harm because it is only a max value and the connections are created on demand. In most cases only one connection is used as you say, in this case no additional connection will be initialized. I will increase it to 3, it is a bit unintuitive to have a connection pool with capacity 1.

sdl-core/src/main/scala/io/smartdatalake/workflow/connection/jdbc/JdbcTableConnection.scala
ddeuber: Makes sense, I will remove it.

zzeekk (approved):

PR #642, opened 2023-04-12
Feature/571 stringorsecret fixes

ddeuber:
Some fixes/improvements for `StringOrSecret`

1. Maps which do not contain `StringOrSecret` are no longer marked as containing secretes in the JSON schema
2. sparkOptions can now also contain secrets

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
@@ -98,30 +98,19 @@ extends SmartDataLakeLogger {
     SecretsUtil.registerProvider(id, providerConfig.provider)
   }
 
-  val resolvedSparkOptions: Option[Map[String, String]] = sparkOptions.map(_.mapValues(_.resolve()))
-
   /**
    * Get Hadoop configuration as Spark would see it.
    * This is using potential hadoop properties defined in sparkOptions.
    */
   def getHadoopConfiguration: Configuration = {
-    PrivateAccessor.getHadoopConfiguration(resolvedSparkOptions.getOrElse(Map()))
+    PrivateAccessor.getHadoopConfiguration(sparkOptions.map(_.mapValues(_.resolve())).getOrElse(Map()))
   }
 
   /**
    * Create a spark session using settings from this global config
    */
   private def createSparkSession(appName: String, master: Option[String], deployMode: Option[String] = None): SparkSession = {
-    // prepare additional spark options
-    // enable MemoryLoggerExecutorPlugin if memoryLogTimer is enabled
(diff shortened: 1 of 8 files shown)


sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
```diff
@@ -112,14 +114,14 @@ extends SmartDataLakeLogger {
...
+    val executorPlugins = resolvedSparkOptions.flatMap(_.get("spark.plugins")).toSeq ++ (if (memoryLogTimer.isDefined) Seq(classOf[MemoryLoggerExecutorPlugin].getName) else Seq())
     val executorPluginOptions = if (executorPlugins.nonEmpty) Map("spark.executor.plugins" -> executorPlugins.mkString(",")) else Map[String, String]()
     // config for MemoryLoggerExecutorPlugin can only be transferred to Executor by spark-options
     val memoryLogOptions = memoryLogTimer.map(_.getAsMap).getOrElse(Map())
     // get additional options from modules
     val moduleOptions = ModulePlugin.modules.map(_.additionalSparkProperties()).reduceOption(mergeSparkOptions).getOrElse(Map())
     // combine all options: custom options will override framework options
-    val sparkOptionsExtended = Seq(moduleOptions, memoryLogOptions, executorPluginOptions).reduceOption(mergeSparkOptions).getOrElse(Map()) ++ sparkOptions.getOrElse(Map())
+    val sparkOptionsExtended = Seq(moduleOptions, memoryLogOptions, executorPluginOptions).reduceOption(mergeSparkOptions).getOrElse(Map()) ++ resolvedSparkOptions.getOrElse(Map())
```
zzeekk: Resolving spark options here is too early, as they will be logged in AppUtil.createSparkSession / optionalConfigs:
https://github.com/smart-data-lake/smart-data-lake/blob/dc6707e06deb8286452672cdef647768e440bd44/sdl-core/src/main/scala/io/smartdatalake/app/AppUtil.scala#L129

I suggest to resolve them as late as possible, and convert options which are basic Strings to StringOrSecret. WDYT?

ddeuber: Yes, you are right. 
When resolving the secrets "on demand" they will be resolved multiple times because they are also used in GlobalConfig.getHadoopConfiguration. Do you think this might be a problem? We could also adapt `StringOrSecret` so that resolved secrets are cached.

zzeekk: Good point, but i think this is ok for now. Some secret providers might also already include a cache, and getting a secret from an Environment variable is cheap. Lets keep it as is.

zzeekk (approved):

PR #643, opened 2023-04-14
Case sensitivity

### LINKED ISSUE #619: Possibility to write case sensitive column name through JDBC
Currently when you write data to a JdbcTableDataObject, columns are converted to lowercase:
https://github.com/smart-data-lake/smart-data-lake/blob/10d8c9e2cb5f7ad35d6be2a25df015923389e689/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/JdbcTableDataObject.scala#L195
This is not always desirable, i.e. when doing 1:1 data migrations and you want to keep the source names. 

I propose to use the setting of `spark.sql.caseSensitive` that we can already get through `SchemaUtil` to decide whether to use case sensitive column names or not:
`if (SchemaUtil.isSparkCaseSensitive) df else df.colNamesLowercase`

**Additional context**
- Maybe implicitly using the spark setting is not the best option and we should make it configurable for each data object / every connection
- Another problem is using column names with Dots.

ddeuber:
See #619 

Breaking changes:
* Instead of the Spark property `spark.sql.caseSensitive`, the new SDL environment property `caseSensitive` (can be set e.g. in `GlobalConfig.environment`) is used to determine whether SDL is case sensitive. If  `spark.sql.caseSensitive` is not explicitely set in a configuration file, it will be set according to `caseSensitive`.
* The data objects `JdbcTableDataObject`, `ParquetFileDataObject` and `IcebergTableDataObject` no longer convert the column names to lowercase when writing. Instead, `StandardizeColNamesTransformer` should be used if lower case column names are required.
* `BlacklistTransformer` and `WhitelistTransformer` are now case sensitive if SDL is case sensitive


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/ParquetFileDataObject.scala
```diff
@@ -93,12 +93,6 @@ case class ParquetFileDataObject( override val id: DataObjectId,
 
   override val options: Map[String, String] = parquetOptions.getOrElse(Map())
 
-  // when writing parquet files, schema column names are forced to lower,
-  // because they can also be accessed by Hive which is case insensitive.
-  // See: https://medium.com/@an_chee/why-using-mixed-case-field-names-in-hive-spark-sql-is-a-bad-idea-95da8b6ec1e0
```
ddeuber, line 98: This issue was fixed in Spark 2.4, see https://spark.apache.org/docs/latest/sql-migration-guide.html#:~:text=when%20spark.sql.caseSensitive%20is%20set%20to%20false%2C%20Spark%20does%20case%20insensitive%20column%20name%20resolution%20between%20Hive%20metastore%20schema%20and%20Parquet%20schema

ddeuber:
It is hard to test the case sensitivity of columns in automated tests because HSQLDB, which we use for these tests, changes the column names to uppercase. I manually tested moving data from one MariaDB table to another with the following config

```
global {
    enableHive = false
}

actions = {
   load = {
     type = CopyAction
     inputId = ext
     outputId = stage
     metadata {
       feed = columnnames
     }
   }
   write = {
     type = CopyAction
     inputId = stage
     outputId = int
     metadata {
       feed = columnnames
     }
   }
}

connections = {
  mariadb {
    type = JdbcTableConnection
    driver = org.mariadb.jdbc.Driver
    url = "jdbc:mariadb://localhost:3306"
    db = test
    authMode {
      type = BasicAuthMode
      user = "root"
      password = "rootpw"
    }
  }
}

dataObjects {
  ext {
    type = JdbcTableDataObject
    connectionId = mariadb
    table = {
     name = SourceTable
    }
  }
  stage {
    type = ParquetFileDataObject
    path = "target/stage/columnnames"
  }
  int {
    type = JdbcTableDataObject
    connectionId = mariadb
    table = {
     name = SinkTable
    }
  }
}
```

sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
```diff
@@ -134,7 +135,19 @@ extends SmartDataLakeLogger {
...
     // get additional options from modules
     val moduleOptions = ModulePlugin.modules.map(_.additionalSparkProperties()).reduceOption(mergeSparkOptions).getOrElse(Map())
-    Seq(moduleOptions, memoryLogOptions, executorPluginOptions).reduceOption(mergeSparkOptions).map(_.mapValues(StringOrSecret)).getOrElse(Map())
+    // if SDL is case sensitive then Spark should be as well
+    val caseSensitivityOptions = Map("spark.sql.caseSensitive" -> Environment.caseSensitive.toString)
+    Seq(moduleOptions, memoryLogOptions, executorPluginOptions, caseSensitivityOptions).reduceOption(mergeSparkOptions).map(_.mapValues(StringOrSecret)).getOrElse(Map())
+  }
+
+  private def checkCaseSensitivityIsConsistent(options: Map[String, StringOrSecret]): Unit = {
+    options.get("spark.sql.caseSensitive")
```
zzeekk:
```suggestion
    options.get(SQLConf.CASE_SENSITIVE.key)
```

zzeekk: Can we replace this in the other places also with the "constant"?

zzeekk (approved):
Nice improvement, lets do this.
Just one minor comment about replacing fixed coded string...

PR #645, opened 2023-04-21
add an entry point for Glue jobs, filtering out all Glue specific args

mand35:
In an Glue job various Glue specific command line arguments are passed to the application. This fails due to unrecognized arguments. This adds a new entry point filtering out all Glue specific arguments. 

Glue job fails, due to unrecognized arguments.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
@@ -116,8 +116,6 @@ class GlueSmartDataLakeBuilder extends SmartDataLakeBuilder {
     val options: Seq[String] = rmArg(List[String](), args.toList)
     logger.info(s"Filtered command line args: +${options.mkString(";")}")
 
-    val scoptVersion = classOf[OParser[_, SmartDataLakeBuilderConfig]].getPackage.getImplementationVersion
-
     OParser.parse(parser, options, SmartDataLakeBuilderConfig()) match {
       case Some(config) =>
         assert(config.overrideJars.isEmpty || ignoreOverrideJars, "Option override-jars is not supported by DefaultSmartDataLakeBuilder. Use DatabricksSmartDataLakeBuilder for this option.")
--- a/sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
@@ -27,12 +27,16 @@ import scopt.OParser
  *
  * Glue passes all job parameters (command line arguments) down to the application. This application filters out the
  *   Glue specific ones.
- * SCOPT issue:
- * This was tested with Glue 4.0. Glue 4.0 uses an older version of scopt, which is specified in the XBOOTPATH.
- * To fix the issue:
- *  - place the newer scopt jar an s3 bucket,
- *  - specify the path to it in the job property `Dependent JARs path` and
- *  - Job parameter: `--user-jars-first` with value `true`
+ * SCOPT issue with Glue 4.0:
+ * Since SmartDataLakeBuilder Version 2.5.0 we use scopt Version 4.0.1 in order to parse command line arguments.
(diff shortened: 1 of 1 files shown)


sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * Glue passes all job parameters (command line arguments) down to the application. This application filters out the
+ *   Glue specific ones.
+ * SCOPT issue:
+ * This was tested with Glue 4.0. Glue 4.0 uses an older version of scopt, which is specified in the XBOOTPATH.
+ * To fix the issue:
+ *  - place the newer scopt jar an s3 bucket,
+ *  - specify the path to it in the job property `Dependent JARs path` and
+ *  - Job parameter: `--user-jars-first` with value `true`
+ */
```
Geheiner:
```suggestion
 * Glue passes all job parameters (command line arguments) down to the application. This application filters out the
 *   Glue specific ones.
 * SCOPT issue with Glue 4.0:
 * Since SmartDataLakeBuilder Version 2.5.0 we use scopt Version 4.0.1 in order to parse command line arguments.
 Unfortunately, Glue 4.0 overwrites the library scopt with an older version by changing the classpath of the boot classloader using a Vm Option that looks like -Xbootclasspath/a:"/path/to/scopt_2.12-3.7.1.jar"
 You are affected if you get an error message like this:
 Exception in User Class: java.lang.NoSuchMethodError : scopt.OptionDef.<init>(Lscopt/OptionDefKind;Ljava/lang/String;Lscopt/Read;)V
 
 * To fix the issue, you basically need to make sure scopt 4 is loaded, effectively undoing the -Xbootclasspath/a option:
 *  - place the newer scopt jar an s3 bucket. You can get it here https://mvnrepository.com/artifact/com.github.scopt/scopt_2.12/4.0.1,
 *  - specify the path to it in the job property `Dependent JARs path` and 
 *  - set Job parameter: `--user-jars-first` with value `true`
 */
```

sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
```diff
@@ -0,0 +1,142 @@
...
+class GlueSmartDataLakeBuilder extends SmartDataLakeBuilder {
```
Geheiner, line 40: Do you know if these arguments are specific to a certain version of glue or if they are expeted to remain stable?

mand35: @Geheiner Do you target the `--user-jars-first` argument or the Glue arguments in general? 
I guess there is no guarantee that these stay constant. 
Do you suggest to turn around and filtering all SDLB arguments instead of filtering out Glue ones?

Geheiner: I'm talking about the Glue arguments.
The ones with --user-jars-first you can always change however you want.
Let's see how stable it is.
If glue keeps creating new arguments that we have to filter you then it may be worth extracting the arguments to ignore into some conf file. But for now it's not worth it, let's keep il like this

Geheiner (approved):

sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
Geheiner: @mand35 
How about letting scopt accept unexpected arguments?
I did this https://github.com/smart-data-lake/smart-data-lake/pull/645/commits/e5b1d10678e8df6af87e031b065344dc91d8b30d
Tested it locally and it has the desired effect

Geheiner (approved):
I approve mysself :P

sdl-core/src/main/scala/io/smartdatalake/app/GlueSmartDataLakeBuilder.scala
Geheiner: the setting actually only causes scopt to issue warnings instead of error and abort as soon as something unexpected is passed, so the user still sees if we mistyped something

zzeekk (approved):

PR #661, opened 2023-05-12
Feature/652 reading schema files from classpath location

### LINKED ISSUE #652: Reading schema files from classpath location
It would be nice to version schema files with the code, and load them from the classpath as resource.
Currently the URI given the reference schema files is strictly treated as Hadoop URI.
On the otherside configuration files can already be loaded from classpath, see also ConfigLoader:95.

Create a new method that generalizes reading files from URI's. If there is a "cp:" prefix the file will be loaded from classpath.
Replace usage of HdfsUtil.readHadoopFile with the new method (where appropriate).

dnLeiting:
The code change implements the feature requested in #652 

The method "parseInputPath" parses files from URI's. If there is a "cp:" prefix the file will be loaded from classpath.
This function is a replacement for using HdfsUtil.readHadoopFile.

Tests for the new "parseInputPath" method have been added.
In addition test for reading a DDL File are added.

It would be nice to version schema files with the code, and load them from the classpath as resource.
Currently the URI given the reference schema files is strictly treated as Hadoop URI.


sdl-core/src/main/scala/io/smartdatalake/util/misc/FileUtil.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package io.smartdatalake.util.misc
+
+import io.smartdatalake.config.ConfigurationException
+import org.apache.hadoop.fs.Path
+
+import java.io.{InputStream, InputStreamReader}
+
+object FileUtil {
```
zzeekk: The name is misleading in my opinion. The methods dont handles files, but classpath ressources. Could we name it ResourceUtil?

sdl-core/src/main/scala/io/smartdatalake/util/misc/FileUtil.scala
```diff
@@ -0,0 +1,50 @@
...
+   */
+  def canHandleScheme(path: Path): Boolean = path.toUri.getScheme == "cp"
+
+  /**
+   * Creates InputStream out of provided file path
+   *
+   * @param path [[Path]] pointing to a directory or file
+   * @return [[InputStream]] from the provided path
+   */
+  def readUriFromPath(path: Path): InputStream ={
```
zzeekk: Also here i would suggest a better name, e.g. readResource.
And should we add an assert statement to validate that canHandleSchema is true?

sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
```diff
@@ -145,6 +147,19 @@ object SchemaUtil {
     StructType.fromDDL(ddl)
   }
 
+  private def parseInputPath(inputPath: String): String = {
```
zzeekk: And again a naming discussion :-)
The method doesn't parse, but just reads the content.
Suggestion: readFromPath?

sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
```diff
@@ -157,7 +172,9 @@ object SchemaUtil {
       case DDL =>
         SparkSchema(getSchemaFromDdl(value))
       case DDLFile =>
-        val content = HdfsUtil.readHadoopFile(value)
+        val valueElements = value.split(";")
+        assert(valueElements.size == 1, s"XSD schema provider configuration error. Configuration format is '<path-to-ddl-file>', but received $value.")
```
zzeekk:
```suggestion
        assert(valueElements.size == 1, s"DDL schema provider configuration error. Configuration format is '<path-to-ddl-file>', but received $value.")
```

sdl-core/src/main/scala/io/smartdatalake/util/misc/FileUtil.scala
dnLeiting: Sounds good to me. :)

zzeekk (approved):

zzeekk:
Looks good, thanks!

dnLeiting:
@zkull you updated the method getSchemaFromXsd in SchemaUtil from taking a String to taking a Path as input. This results in a conflict with my changes. I would leave your implementation as it is and not use the method readFromPath in case a Xsd file is provided as schema. Is that ok?

zzeekk:
You're right, the merge is indeed tricker... if we dont use readFromPath for Xsd, it cannot read files from classpath. Unfortunately Xsd is even the case that triggered this feature request...
I did a rebase and merge in a separate branch and PR #669, as rebase would need a "force commit" if i would have continued on this branch.
You find my changes in https://github.com/smart-data-lake/smart-data-lake/pull/669/commits/cdf81ef3a90708c1da54e0291055d374796ad759

zzeekk (changes requested):
@dnLeiting: Can we close this one in favor of #669?

dnLeiting:
Can be closed with respect to #669

Issue #666, opened 2023-05-23
Add hostname and executorId when logging to Azure LogAnalytics


zzeekk:
**Is your feature request related to a problem? Please describe.**
If multiple SDLB jobs log into the same LogAnalytics table, it's hard to discern the logs of the different jobs and hosts.
Partly this is also due to the context not being filled for every LogEvent.

**Describe the solution you'd like**
Add hostname, and executorId to context, and make sure appName & runId are present in context of all LogEvents.

PR #670, opened 2023-05-24
Feature/666 improve loganalytics logging

zzeekk:
see #666

### DIFF
--- a/sdl-azure/src/main/scala/io/smartdatalake/util/azure/LogAnalyticsAppender.scala
+++ b/sdl-azure/src/main/scala/io/smartdatalake/util/azure/LogAnalyticsAppender.scala
@@ -19,12 +19,22 @@
 
 package io.smartdatalake.util.azure
 
+import io.smartdatalake.app.AppUtil
 import org.apache.logging.log4j.core.Filter.Result
 import org.apache.logging.log4j.core._
 import org.apache.logging.log4j.core.appender.AbstractAppender
+import org.apache.logging.log4j.core.impl.ThrowableProxy
+import org.apache.logging.log4j.core.time.Instant
 import org.apache.logging.log4j.layout.template.json.JsonTemplateLayout
+import org.apache.logging.log4j.message.Message
+import org.apache.logging.log4j.spi.DefaultThreadContextMap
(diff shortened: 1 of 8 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/misc/SmartDataLakeLogger.scala
```diff
@@ -38,9 +39,11 @@ trait SmartDataLakeLogger {
     e
   }
 
-  private[smartdatalake] def logWithSeverity(severity: Level, msg: String): Unit = {
+  private[smartdatalake] def logWithSeverity(severity: Level, msg: String, cause: Throwable): Unit = {
     severity match {
+      case Level.ERROR if Environment.includeDAGResultExceptionInLog => logger.error(msg, cause)
       case Level.ERROR => logger.error(msg)
```
pgruetter, line 45: Isn't log.error called twice if includeDAGResultExceptionInLog is true, once with cause, once without?

sdl-azure/src/main/scala/io/smartdatalake/util/azure/LogAnalyticsAppender.scala
```diff
@@ -81,13 +91,20 @@ class LogAnalyticsAppender(name: String, var backend: Option[LogAnalyticsBackend
...
   override def stop(): Unit = {
     flush()
     super.stop()
-    println("stopped")
+    println("LogAnalyticsAppender stopped")
+  }
+
+  private def enrichSparkLocalProperties(event: LogEvent) = {
+    // If we are on an executor, we have a TaskContext
+    // On executors we need to enricht LogEvent with SDLB informations from spark local properties
```
pgruetter:
```suggestion
    // On executors we need to enrich LogEvent with SDLB information from spark local properties
```

pgruetter (approved):
Got two small comments, please check.

sdl-core/src/main/scala/io/smartdatalake/util/misc/SmartDataLakeLogger.scala
zzeekk: No, pattern matching always executes only one case statement and then stops.

Issue #656, opened 2023-05-03
Modernize Schema Viewer


ddeuber:
The SDL schema viewer suffers from multiple issues:

- It uses obsolete Javascript libraries (e.g. jQuery mobile, tv4).
- The build chain has too many manual steps.
- There are performance issues (see e.g. #644).

Therefore, the schema viewer should be rewritten as a React component. We should also analyze whether the d3 tree structure is still the best way to visualize the schema.

ddeuber:
The definitions section of the SDLB JSON schema is, because of the nested class hierarchy, not conformant with the JSON schema specification: https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-9:~:text=This%20keyword%27s%20value%20MUST%20be%20an%20object.%20%20Each%20member%20value%20of%20this%0A%20%20%20object%20MUST%20be%20a%20valid%20JSON%20Schema.

We should be consistent with the JSON schema specification because it helps when using JSON schema types in Typescript for the schema viewer. In addition, there might be validation issues when people use the schema for other purposes (e.g. in IDEs).

zzeekk:
Hi @ddeuber, you mean because there are nested definitions?
I validated the schema in different tools, e.g. https://json-schema.hyperjump.io/ and also IntelliJ (JSON Schema Version 7). It seems valid. Do you have a validator that mentions errors?

ddeuber:
Yes, because of the nesting. I noticed it because it is not consistent with the Typescript type JSONSchema7 in @types/json-schema. Apparently, not all schema validators are so strict with the JSON schema. Nevertheless, I think we should follow the specification to be on the safe side.

zzeekk:
I see, you would like to use @types/json-schema. But the latest version of json-schema spec is 2020-12. And they have nested $ref even in their documentation: https://json-schema.org/draft/2020-12/json-schema-core.html#name-dereferencing.
They list other javascript libraries supporting 2020-12 under https://json-schema.org/implementations#validator-javascript.
Do you think it is possible to adapt sdl-schema and what would be the effort for it?

ddeuber:
In 2020-12, the definitions resp. $defs can also not be nested: https://json-schema.org/draft/2020-12/json-schema-core.html#name-schema-re-use-with-defs:~:text=Each%20member%20value%20of%20this%20object%20MUST%20be%20a%20valid%20JSON%20Schema.

[...]

PR #674, opened 2023-06-02
Integrate new schema viewer (#656)

ddeuber:
This PR can be used to provide feedback about the new schema viewer.


pgruetter:
Thanks a lot for your work, looks very clean and works smoothly. 
Only thing I found was an overlap that happens sometimes (i.e. when two main branches are open at the same time):
image

If you open the same branches on the old schema viewer, you can see that it automatically moved the entries to make room. 
Not sure how difficult that is to fix.
As alternative, you'd have to close other main branches as soon as you open a second one which is not optimal.

ddeuber:
You are right, this is a bug. I created an issue https://github.com/smart-data-lake/schema-viewer/issues/8.

pgruetter (approved):
Awesome! Thanks a lot for your work 👍

Issue #672, opened 2023-05-24
State file: add inputIds and outputIds


zzeekk:
**Is your feature request related to a problem? Please describe.**
State files for now does not include informations about the lineage of actions, e.g. input/outputIds for every action.
In the Visualizer UI we would like to show the lineage of a run. We could get the information from the configuration, but this might not reflect the version with which the state file as been created.

**Describe the solution you'd like**
Add input/outputIds to RuntimeInfo for every Action.

Issue #671, opened 2023-05-24
State file: add startTstmp and duration for Init and Prepare phases


zzeekk:
**Is your feature request related to a problem? Please describe.**
State files for now only include informations about the execution phase, but not prepare and init phase.
In the Visualizer UI we would like to include prepare and init-phase in the timeline view.

**Describe the solution you'd like**
Add start- and end-time for prepare and init-phase to RuntimeInfo (per Action).
Currently there are startTstmp + duration for exec-phase in RuntimeInfo.
I suggest to add startTstmpPrepare + durationPrepare and startTstmpInit + durationInit attributes.

As we begin to change the state file, we should also add a version number. I suggest to put the sdlbVersion (BuildVersionInfo.readBuildVersionInfo) and the appVersion (AppUtil.getManifestVersion) in ActionDAGRunState.

zzeekk:
@pgruetter : please review suggestion.

pgruetter:
Thanks, looks good! 
Should we also include the addition of a version number to the state file? Can also create a separate issue for that.

zzeekk:
Right, lets add version with this issue as well, i already extended the description.

PR #675, opened 2023-06-04
672 - Insert inputIds and outputIds to state file

pgruetter:
Add version number to state file and only accept recovery for state files of same version

See #672 
Add inputIds and outputIds to state file. 
Also see comment from #671: version number is added in this PR already.

Needed by UI to properly show lineage at runtime.

### DIFF
--- a/sdl-core/ext-state/state-test/succeeded/test.1.1.json
+++ b/sdl-core/ext-state/state-test/succeeded/test.1.1.json
@@ -36,7 +36,13 @@
           "bytes_written" : 20
         }
       } ],
-      "dataObjectsState" : [ ]
+      "dataObjectsState" : [ ],
+      "inputIds" : [ {
+        "id" : "oneInput"
+      } ],
+      "outputIds" : [ {
+        "id" : "oneOutput"
+      } ]
     }
   },
   "isFinal" : true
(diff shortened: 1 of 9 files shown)


pgruetter:
@zzeekk Not sure if Environment is the right place to store the hardcoded value for run state format version.

zzeekk:
Hi @pgruetter, great that you do some coding again 👍 
I have some general discussion points:
1. is it a good idea to introduce another version number that we need to manage? Couldn't we reuse SDLB version number? We should put SDLB version and application jar version there in any case for traceability...
2. Shouldn't state file versions be backward compatible, e.g. a newer SDLB version should be able to read/interpret schema files of older versions?

pgruetter:
I thought about these points as well:
1. I would not link it to the SDLB version. The state run format should not change very often (if at all). If we increment the version with each SDLB version (and with that also use SNAPSHOT versions), handling gets more difficult.
2. Generally speaking: yes, they should be. But note that the assertion statement I've inserted is only checked during recoveries. I would argue that you should not have a failed run, then upgrade SDL and try to recover the run. This could lead to all sorts of strange behavior. 
New attributes (like the version number), should be optional so we can read old files though.

Happy to further discuss offline.

pgruetter:
@zzeekk As discussed: RunStateFormatVersion changed to Int and moved from Environment to Object. Will insert SDLB version info in a separate PR.

zzeekk (approved):
LGTM

PR #676, opened 2023-06-05
Feature/671 start tstmp for init and prepare in statefile

### LINKED ISSUE #671: State file: add startTstmp and duration for Init and Prepare phases

pgruetter:
Closes #671 
Add startTstmp and duration for prepare and init phase. 

See #671

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
@@ -75,7 +75,7 @@ case class DataObjectState(dataObjectId: DataObjectId, state: String) {
 
 private[smartdatalake] object ActionDAGRunState {
 
-  val runStateFormatVersion: Int = 1
+  val runStateFormatVersion: Int = 2
 
   private val durationSerializer = Json4sCompat.getCustomSerializer[Duration](formats => (
     {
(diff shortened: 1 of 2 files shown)


zzeekk:
Hi @pgruetter, here i have another remark we discussed last friday with @Entouanes:
Parsing the ISO duration format in javascript is not easy, and we need to do it in a MongoDB stored procedure to create the workflow summary! We found a way using a complex regexp and no additional library.
But considering that for the workflow summary we need to calculate endTime by max(startTime+duration), which includes parsing the duration, i think it would be better to replace prepare/init/execDuration by prepare/init/execTstmp. WDYT?
Or keep both for simple backward compatibility.

zzeekk (approved):

pgruetter:
@zzeekk Thanks. I also increased runStateFormatVersion to 2 as the format has now changed.

PR #683, opened 2023-06-09
Delete snapshot schema after release

ddeuber:
After a new version has been released, the schema of the according snapshot version will be deleted.


.github/workflows/release.yml
```diff
@@ -92,6 +92,7 @@ jobs:
         # note that sdl-schema.json should be available in the location below from the previous maven build.
         mv sdl-lang/target/classes/sdl-schema.json static/json-schema-viewer/schemas/sdl-schema-${MVN_RELEASE_VERSION}.json
         git add static/json-schema-viewer/schemas/sdl-schema-${MVN_RELEASE_VERSION}.json
+        git rm static/json-schema-viewer/schemas/sdl-schema-${MVN_RELEASE_VERSION}-SNAPSHOT.json
```
pgruetter: What happens if the SNAPSHOT doesn't exist? Wouldn't this step fail and abort the whole pipeline? Maybe we should add `--ignore-unmatch` to `git rm` ?

ddeuber: Usually it should exist, but yes, better be safe.

pgruetter (approved):

PR #686, opened 2023-06-11
fix handling skipped SubFeeds when writing to the same DataObject

zzeekk:
Fix handling skipped SubFeeds when writing to the same DataObject from multiple Actions.


zzeekk:
closed in favor of #688

PR #688, opened 2023-06-12
#687: fix handling skipped SubFeeds when writing to the same DataObject

### LINKED ISSUE #687: Skipped not handled correctly if multiple Actions writing to the same DataObject
**Describe the bug**
When writing to the same DataObject from multiple Actions (which is actually an anti-pattern), skipped SubFeeds are not handled correctly, e.g. if one SubFeed is skipped, the union of all SubFeeds writing to this DataObject is also skipped. In consequence the subsequent Action will be skipped.

**Expected behavior**
The union of all SubFeeds writing to a DataObject is only marked as skipped, if all SubFeeds are skipped.

zzeekk:
closes #687 

Fix handling skipped SubFeeds when writing to the same DataObject from multiple Actions.


mand35 (approved):
look good and works as expected.

PR #699, opened 2023-06-19
Housekeeping mode

pgruetter:
## What changes are included in the pull request?
Contains a new blog article about housekeeping mode.

### DIFF
--- a/blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
+++ b/blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
@@ -66,7 +66,7 @@ stage_json {
 
 Each time we start to process uploaded data, we use the `run_id` to keep track of all batch jobs and version of files delivered.
 If you use a state path (see [commandLine](../../docs/reference/commandLine)), 
-your runs automatically use `run_id`s and you can use it by extending your DataObject:
+your runs automatically generate a `run_id` to identify the run, and you can use it by extending your DataObject:
 
 ```
 stage_json {
--- a/blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
+++ b/blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
@@ -161,7 +161,7 @@ This is the else condition and simply returns the existing partition values if t
 
 :::info
 The housekeeping mode is applied after writing a DataObject.
-Keep in mind, that it is always executed.
+Keep in mind, that it is executed with every run.
 :::
 
 #### Compaction
--- a/blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
+++ b/blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
@@ -97,17 +97,17 @@ This partition layout has many advantages in our case as we know exactly
 during which run a particular file was processed and which reporting unit uploaded it.
 In further stages we can clearly work with files that were processed in the current run and not touch any old `run_id`s. 
 
-But the drawbacks of this partition scheme becomes apparent after running for a longer time:
+For this use case, a few things are important to note:
   - Some reporting units don't upload data for days. You end up with only a few reporting_unit partitions per run_id.
   - File sizes are rather small (< 1 MiB), partition sizes end up very small too.
   - If you use hourly runs and run 24/7, you end up with 168 partitions per week, plus sub-partitions for reporting units.
   - Once files are correctly processed, we don't read the uploaded files anymore. 
     We still keep them as raw files should we ever need to re-process them.
 
-If you have actions working with all partitions, they will become very slow.
+The drawback becomes apparent when you have actions working with all partitions, they will become very slow.
 Spark doesn't like a lot of small partitions.
 
-To mitigate these drawbacks, we use SDLB's Housekeeping Feature.
+To mitigate that, we use SDLB's Housekeeping Feature.
 
 ## HousekeepingMode
 If you take a look at DataObject's parameters, you will see a `housekeepingMode`. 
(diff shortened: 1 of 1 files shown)


iliassoto (approved):

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
```diff
@@ -0,0 +1,197 @@
...
+      fileX.json
+```
+This partition layout has many advantages in our case as we know exactly 
+during which run a particular file was processed and which reporting unit uploaded it.
+In further stages we can clearly work with files that were processed in the current run and not touch any old `run_id`s. 
+
+But the drawbacks of this partition scheme becomes apparent after running for a longer time:
+  - Some reporting units don't upload data for days. You end up with only a few reporting_unit partitions per run_id.
+  - File sizes are rather small (< 1 MiB), partition sizes end up very small too.
+  - If you use hourly runs and run 24/7, you end up with 168 per week, plus sub-partitions for reporting units.
```
iliassoto: 168 files?

pgruetter: Partitions :-) Thanks, fixed it.

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
```diff
@@ -0,0 +1,197 @@
...
+    inputId = import_json
+    outputId = stage_json
+    executionMode = { type = FileIncrementalMoveMode }
+    metadata.feed = stage_json
+}
+```
+
+Each time we start to process uploaded data, we use the `run_id` to keep track of all batch jobs and version of files delivered.
+If you use a state path (see [commandLine](../../docs/reference/commandLine)), 
+your runs automatically use `run_id`s and you can use it by extending your DataObject:
```
zzeekk:
```suggestion
your runs automatically generate a `run_id` to identify the run, and you can use it by extending your DataObject:
```

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
```diff
@@ -0,0 +1,197 @@
...
+```
+stage_json {
+    type = JsonFileDataObject
+    path = processedFolder
+    partitions = [run_id,reporting_unit]
+    schema = """reporting_unit string, run_id string, ...."""
+}
+```
+Note how we just use run_id as part of the schema without any further declaration.
+Since we use the state path, SDLB uses a `run_id` internally, and we can use it too.
```
zzeekk:
```suggestion
Since we use the state path, SDLB uses a `run_id` internally, and if it's referenced as partition column in a DataObject, processed data get automatically assigned the id of the current run.
```

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
```diff
@@ -0,0 +1,197 @@
...
+      file1.json
+    reporting_unit=rp03/
+      fileX.json
+```
+This partition layout has many advantages in our case as we know exactly 
+during which run a particular file was processed and which reporting unit uploaded it.
+In further stages we can clearly work with files that were processed in the current run and not touch any old `run_id`s. 
+
+But the drawbacks of this partition scheme becomes apparent after running for a longer time:
+  - Some reporting units don't upload data for days. You end up with only a few reporting_unit partitions per run_id.
```
zzeekk, line 101: This is actionally not yet a drawback, is it?
The list rather explains details of how partitions look like. But the drawback is the number of partitions (incl. subpartitions) created in an year. Also if you dont have to read these files, there might be a list partitions operation somewhere, which can already take a lot of time. And as you mention later, if you have to read all the partitions with small files it gets really slow...

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
```diff
@@ -0,0 +1,197 @@
...
+The run_id is floored to the next 500 value, so as example, the new value of run_id 1984 will be 1500 (because integer 1984/500=3, 3*500=1500).  
+Remember that we need to return all partition values in the map, also the ones we don't want to alter.
+For `reporting_unit` we simply return the existing value `elements.reporting_unit`.
+
+`..., elements)`  
+This is the else condition and simply returns the existing partition values if there is nothing to archive.
+
+:::info
+The housekeeping mode is applied after writing a DataObject.
+Keep in mind, that it is always executed.
```
zzeekk: Maybe more precise, wdyt?
```suggestion
Keep in mind, that it is executed with every run.
```

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
```diff
@@ -0,0 +1,197 @@
...
+:::
+
+#### Compaction
+We don't want compact files in our case.
```
zzeekk:
```suggestion
We don't want to compact files in our case. 
```

zzeekk (approved):
Nice article. I made some small suggestions.
Maybe at the end we could mention that we are thinking about extending it for DeltaLake / Iceberg optimize and vacuum, and that they should let us now if they have ideas how this should work, or even have completly different ideas for housekeeping (we might also create a custom housekeeping mode to create own logic in a project...).

blog/2023-06-08-SDL-Housekeeping/2023-06-08-housekeeping.md
pgruetter: Good point. I rewrote that in https://github.com/smart-data-lake/smart-data-lake/pull/699/commits/48a5799cd74298feeb6e67e8330c4433ec898018

Issue #700, opened 2023-06-27
Uplad final state to API


zzeekk:
**Is your feature request related to a problem? Please describe.**
The final state of a SDLB run should be uploaded to an API for showing it in the UI.

**Describe the solution you'd like**
Extend existing FinalStateWriter with an option to configure an http endpoint, where it uploads the state.

zzeekk:
I finally found it better to create a separate State Listener. It can be configured as follows:
```
global {
  stateListeners = [{
    className = "io.smartdatalake.util.misc.FinalStateUploader"
    options {
      uploadUrl = "https://<api-hostname>/api/v1/run"
      #uploadStagePath = "upload-stage-path"
    }
  }]
}
```
uploadStagePath is optional, if set failed uploads are temporarily stored in this folder and the upload retried on the initialization of the next run.

pgruetter:
Done in #701

PR #701, opened 2023-06-29
Feature/700 upload state

zzeekk:
see #700 

Should we create a Snapshot Release from this branch for testing?


pgruetter (approved):
Tested, works correctly including upload of previously failed state files. 
We do need to think about authentication and potentially other settings like timeout etc.

PR #705, opened 2023-07-18
Feature/minor improvements

zzeekk:
Small improvements and fixes:
- sort partition values in log output when writing to DataObject
- fix and improve SQLDf(s)Transformer description
- PartitionDiffMode: apply selectExpression before nbOfPartitionValuesPerRun
- LabSparkDataObjectWrapper: add dropPartitions, infos and partitionModDates methods
- HiveTableDataObject: avoid wrong warning about different path of existing table by using getAbsolutePath
. This fix uses the same logic as already present in DeltaLakeTableDataObject.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
@@ -187,11 +187,11 @@ private[smartdatalake] trait HadoopFileDataObject extends FileRefDataObject with
    * List partitions on data object's root path
    */
   override def listPartitions(implicit context: ActionPipelineContext): Seq[PartitionValues] = {
-    getPartitionPathStatis
+    getPartitionPathsStatus
       .map(path => extractPartitionValuesFromDirPath(path.getPath.toString))
   }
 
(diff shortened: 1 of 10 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
```diff
@@ -187,14 +187,18 @@ private[smartdatalake] trait HadoopFileDataObject extends FileRefDataObject with
    * List partitions on data object's root path
    */
   override def listPartitions(implicit context: ActionPipelineContext): Seq[PartitionValues] = {
+    getPartitionPathStatis
```
Geheiner:
```suggestion
    getPartitionPathStatuses
```
picky detail, but in english that s a bit different than in german:
https://www.grammarbook.com/blog/singular-vs-plural/plural-of-status/#:~:text=For%20those%20in%20search%20of,with%20an%20%E2%80%9Ces%E2%80%9D).

Geheiner (approved):
LGTM

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
zzeekk: You're right, changed it to getPartitionPathsStatus

PR #727, opened 2023-09-12
update ExcelFileDataObject Docu with all schema types

mand35:
In the ExcelFileDataObject  descriptions there was a typo for the caseClase schema type as well as some missing. 
This was corrected the be shown properly in the schema viewer.


Geheiner (approved):
Thanks!

PR #728, opened 2023-09-12
add option to replace nonstandardSQL with _ in col names

mand35:
When using removeNonStandardSQLNameChars so far chars are just removed. Thus a column name "sum of values" will be "sumofvalues". I suggest to change it to replace all non SQL standard characters with "_" thus we get "sum_of_values".

A new option replaceNonStandardSQLNameCharsWithUnderscores is introduced for the StandardizeColNamesTransformer which first (in function order) replace everything which is not a-zA-Z0-9_ with "_". Multiple consecutive instances are replaced by single "_"

I as a user want readable column names.


mand35:
@zkull @patrickgruetter changed the scope, description and PR to replace all non-standard SQL chars.

zzeekk (approved):
LGTM.

PR #732, opened 2023-10-03
update timestamps of webservice to be newer, preventing timeouts

Geheiner:
Every now and then the default timestamps that we use for querying the web api become too old, resulting in timeouts.
The previous timestamps resulted in this url https://opensky-network.org/api/flights/departure?airport=LSZB&begin=1630310979&end=1630656579 when trying to run with state files, which results in a timeout on the server side.
Solution: use newer timstamps


pgruetter (approved):
LGTM

PR #733, opened 2023-10-04
Feature/documentation corrections

moskast:
What changes are included in the pull request?
Added instructions for DeltaLakeTableDataObject definition changes for `btl-departures-arrivals-airports` and `btl-distances`.

Why are the changes needed?
As these were only in the solution file but not mentioned in the tutorial itself.


pgruetter (approved):
LGTM

PR #735, opened 2023-10-05
Feature/fix typos

Geheiner:
We currently have a problem in the getting-started guide that the requests to https://opensky-network.org do not work with older timestamps.
I now updated the timestamps to the most recent ones (begin=09-10-2023, end=11-10-2023).
Moreover, I explained this issue in the getting-started and told the user for part 1 to replace older timestamps in case of problems.
For part3 where we use a Custom Data Object, I implemented a default mechanism in the code that always gets recent timestamps. At least the problem will be definetly solved there. This mechanism is also included in this pr for the getting-started code base: https://github.com/smart-data-lake/getting-started/pull/19
Finally, I fixed https://github.com/smart-data-lake/smart-data-lake/pull/735#issuecomment-1750469848


Geheiner:
By the way I also noticed that we have a couple of dead links towards the schema viewer.
Looks like it's URL used to be [json-schema-viewer](http://smartdatalake.ch/json-schema-viewer/index.html) but now it's https://smartdatalake.ch/JsonSchemaViewer
Based on our usual conventions, we should probably change the url of the schema viewer to json-schema-viewer?

pgruetter (approved):

pgruetter:
> we should probably change the url of the schema viewer to json-schema-viewer?
Can you create a separate issue / PR for that?

docs/getting-started/part-1/get-departures.md
```diff
@@ -157,8 +164,8 @@ podman run --rm -v ${PWD}/data:/mnt/data -v ${PWD}/target:/mnt/lib -v ${PWD}/con
 After executing it, you will see the file *data/stg_departures/result.json* has been replaced with the output of your pipeline.
 
 :::caution
-Since both web servers are freely available on the internet, they might be overloaded by traffic.
-If the download fails because of a timeout, wait a couple of minutes and try again.
+Since both web servers are freely available on the internet, **a rate limiting applies**. https://opensky-network.org/ will stop responding if you make too many calls.
+If the download fails because of a timeout, wait a couple of minutes and try again. In the worst case, it should work again on the next day.
```
pgruetter: Sounds a bit strange IMHO, "in the worst case, it should work again".
```suggestion
If the download fails because of a timeout, wait a couple of minutes and try again. In the worst case, it will only work again the following day.
```

pgruetter (approved):
Just one suggestion. Feel free to ignore it ;-)

Issue #529, opened 2022-05-18
Collect Delta Lake query stats as metrics


zzeekk:
**Is your feature request related to a problem? Please describe.**
Delta Lake collects detailed statistics about each executed transaction in a "history". These statistics should be logged as metrics by SDLB.

**Describe the solution you'd like**
After writing to Delta Lake in DeltaLakeTableObject writeSparkDataFrame the "history" should be queried and metrics returned by writeSparkDataFrame. To get the right entry from "history" the "userMetadata" feature should be used.
Currently writeSubFeed supports to return metrics, but not writeSparkDataFrame. This has to be refactored.
It might be a good point to refactor also how Spark metrics are collected, as this happens currently asynchronous, but could be handled synchronous as with FilesObservation.

**Additional context**
https://docs.delta.io/latest/delta-utility.html#history-schema
https://books.japila.pl/delta-lake-internals/demo/user-metadata-for-labelling-commits/
https://docs.databricks.com/delta/delta-utility.html#retrieve-delta-table-history

PR #736, opened 2023-10-07
implement custom metrics per DataObject (#529)

zzeekk:
See #529 

This is a larger refactoring of how metrics are handled.
Changes include:
- subfeeds are enriched with metrics
- spark metrics are collected synchronously
- specific metrics for jdbc, delta, iceberg and snowflake
- standardize naming for inserted/updated/deleted rows
- add metrics to SubFeedsExpressionData so it can be used in executionCondition
- ActionPipelineContext.phase is now implemented immutable
- ActionPipelineContext.currentAction added

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
@@ -43,7 +43,7 @@ private[smartdatalake] class SparkStageMetricsListener(actionId: ActionId, dataO
   /**
    * Stores jobID and jobDescription indexed by stage ids.
    */
-  val jobInfoLookupTable: mutable.Map[Int, JobInfo] = mutable.Map.empty
+  val stageIdToJobInfo: mutable.Map[Int, JobInfo] = mutable.Map.empty
 
   // parses job group
   private val jobGroupRegex = (s"${Regex.quote(context.appConfig.appName)} $actionId runId=([0-9]+) attemptId=([0-9]+)").r.unanchored
@@ -66,7 +66,7 @@ private[smartdatalake] class SparkStageMetricsListener(actionId: ActionId, dataO
     jobInfo.foreach { i =>
       if (dataObjectId.isEmpty || i.dataObjectId.contains(dataObjectId.get)) {
         runningJobs(i.id) = i
-        jobStart.stageIds.foreach(stageId => jobInfoLookupTable(stageId) = i)
+        jobStart.stageIds.foreach(stageId => stageIdToJobInfo(stageId) = i)
       }
     }
   }
@@ -84,7 +84,7 @@ private[smartdatalake] class SparkStageMetricsListener(actionId: ActionId, dataO
   override def onStageCompleted(stageCompleted: SparkListenerStageCompleted): Unit = {
     // extract basic job information
     val stageId = stageCompleted.stageInfo.stageId
-    jobInfoLookupTable.get(stageId).foreach { jobInfo =>
+    stageIdToJobInfo.get(stageId).foreach { jobInfo =>
       // extract useful information/metrics
       val taskMetrics = stageCompleted.stageInfo.taskMetrics
       val shuffleReadMetrics = taskMetrics.shuffleReadMetrics
@@ -136,7 +136,7 @@ private[smartdatalake] class SparkStageMetricsListener(actionId: ActionId, dataO
   /**
    * Wait for Spark metrics and return main metrics from latest job.
    */
-  def waitForMetrics(timeoutSec: Int = 10): MetricsMap = {
+  def waitForLastMetrics(timeoutSec: Int = 10): MetricsMap = {
     waitForSparkMetrics(timeoutSec).sortBy(_.jobInfo.id).lastOption.map(_.getMainInfos).getOrElse(Map())
   }
 }
@@ -152,7 +152,7 @@ object SparkStageMetricsListener {
     // exec spark code
     sparkCode
     // return metrics
-    stageMetricsListener.map(_.waitForMetrics()).getOrElse(Map())
+    stageMetricsListener.map(_.waitForLastMetrics()).getOrElse(Map())
   }
 }
 
--- a/sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
@@ -19,20 +19,26 @@
 
 package io.smartdatalake.metrics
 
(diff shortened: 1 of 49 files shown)


sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
```diff
@@ -19,30 +19,38 @@
...
+  /**
+   * Keeps track of running jobs, e.g. jobs that started but did not yet end.
+   * Map key is jobId.
+   */
+  val runningJobs: mutable.Map[Int, JobInfo] = mutable.Map.empty
 
   /**
    * Stores jobID and jobDescription indexed by stage ids.
    */
   val jobInfoLookupTable: mutable.Map[Int, JobInfo] = mutable.Map.empty
```
Geheiner:
```suggestion
  val stageIdToJobInfo: mutable.Map[Int, JobInfo] = mutable.Map.empty
```
A name like that makes it clear that it's just a map. I found the word "Table" confusing, I needed to open IntelliJ to understand what it does

sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
```diff
@@ -84,13 +104,56 @@ private[smartdatalake] class SparkStageMetricsListener(action: Action)(implicit
...
+    metrics
+  } finally {
+    context.sparkSession.sparkContext.removeSparkListener(this)
+  }
+
+  /**
+   * Wait for Spark metrics and return main metrics from latest job.
+   */
+  def waitForMetrics(timeoutSec: Int = 10): MetricsMap = {
+    waitForSparkMetrics(timeoutSec).sortBy(_.jobInfo.id).lastOption.map(_.getMainInfos).getOrElse(Map())
```
Geheiner, line 140: what about the metrics for the other jobs that are not the last one?
Are we only interested in the metrics of the last one?
Then why are the others even stored if we are just interested in the last one?

sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
```diff
@@ -84,13 +104,56 @@ private[smartdatalake] class SparkStageMetricsListener(action: Action)(implicit
...
     }
+    metrics
+  } finally {
+    context.sparkSession.sparkContext.removeSparkListener(this)
+  }
+
+  /**
+   * Wait for Spark metrics and return main metrics from latest job.
+   */
+  def waitForMetrics(timeoutSec: Int = 10): MetricsMap = {
```
Geheiner:
```suggestion
  def waitForMetricsOfLastJob(timeoutSec: Int = 10): MetricsMap = {
```

sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRun.scala
```diff
@@ -112,34 +108,36 @@ private[smartdatalake] case class ActionDAGRun(dag: DAG[Action], executionId: SD
...
+        node.prepare(actionContext)
         node.outputs.map(outputDO => DummyDAGResult(outputDO.id.id))
       case x => throw new IllegalStateException(s"Unmatched case $x")
     }
   }
 
-  def init(implicit context: ActionPipelineContext): Seq[SubFeed] = {
-    context.phase = ExecutionPhase.Init
+  def init(context: ActionPipelineContext): Seq[SubFeed] = {
+    implicit val phaseContext: ActionPipelineContext = context.copy(phase = ExecutionPhase.Init)
```
Geheiner, line 126: how can it happen that context.phase is not Init in this situation? seems strange

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
```diff
@@ -231,7 +231,7 @@ trait GenericSchema extends GenericTypedObject {
     SchemaConverter.convert(this, toSubFeedType)
   }
   def equalsSchema(schema: GenericSchema): Boolean = {
-    diffSchema(schema).nonEmpty || schema.diffSchema(this).nonEmpty
+    diffSchema(schema).isEmpty && schema.diffSchema(this).isEmpty
```
Geheiner, line 234: looks like a massive bugfix! How come it seemded to have no big impact until now?

Geheiner (commented):
Let's discuss this together.
I'm not sure about the impact of the phase now being immutable and on the syncronous waiting for metrics

sdl-core/src/main/scala/io/smartdatalake/metrics/SparkStageMetricsListener.scala
zzeekk: For debugging and maybe future detailled Spark job statistics it is interesting to keep all metrics.

sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRun.scala
zzeekk: Well, this is the beginning of the Init-Phase and we have to initialize the context correctly.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
zzeekk: You're right.... for now this method is only used in IcebergTableDataObject. I did some tests and found this bug.

zzeekk:
> Let's discuss this together. I'm not sure about the impact of the phase now being immutable and on the syncronous waiting for metrics

Well, the flow gets easier like that. I discovered this possibility when implementing expectations, and think we should apply it here as well, as it also allows to better control the metrics created per DataObject, e.g. every DataObject can decide it's own metrics.

sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRun.scala
Geheiner: And why not directly change the original context variable when you call the respective function ?, eg
actionDAGRun.init(context.copy(phase = ExecutionPhase.Init))
Or do we have it in too many places?

Geheiner (approved):
It looks good to me

sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRun.scala
zzeekk: Yes, it's called in many places. Like this you don't have to remember to pass a context with phase = ExecutionPhase.Init to this function.

PR #745, opened 2023-10-25
739 custom housekeepingmode implementation

### LINKED ISSUE #739: Custom HousekeepingMode implementation
I would like to extend the functionality of PartitionRetentionMode housekeeping with custom logging.

Currently HousekeepingMode is a sealed trait. This configuration pattern does not allow to implement custom housekeeping implementations. HousekeepingMode should be changed to a normal trait, and the corresponding configuration parsing should dynamically lookup and instantiate the class name given in type (e.g. extending ParsableFromConfig[...]), instead of using the mechanism to instantiate children's of a sealed trait.
The new mechanism is already implemented in different places like ExecutionMode. It should be changed for HousekeepingMode as well.

moskast:
See #739


zzeekk (approved):
Thanks!

PR #758, opened 2023-11-10
755 create a generic transformer for null handling

### LINKED ISSUE #755: Create a Generic Transformer for null handling
One common transformation is handling null values in various columns. We could provide a generic transformer to handle these.

Create a new transformer called `ConvertNullValuesTransformer`.

Parameters: 
`whitelist` (default=None), `blacklist` (default=None), `valueForString` (default="na"), `valueForNumber` (default=-1)

Function:
Replaces all null values in columns of String type with `valueForString` and for number types with `valueForNumber`. 
If `whitelist` is set, only applies it to the given list. If `blacklist` is set, these columns are excluded. If neither is set, applies the function to all columns.

**Additional context**
- I don't think setting a boolean value makes sense. Both true and false as default are dangerous.
- Check if the given columns actually exist in the Data Object to alert the user of any typos.

ThierryGirod:
 
Added a new transformer called `ConvertNullValuesTransformer` to handle null values in various columns according to issue #755 .
Refactoring of the `WhitelistTransformer` and `BlacklistTransformer` because the filtering functionality is also needed in  `ConvertNullValuesTransformer` 

See  #755

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/misc/GenericSchemaUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/misc/GenericSchemaUtil.scala
@@ -30,6 +30,14 @@ object GenericSchemaUtil {
     else schema.columns.contains(colName.toLowerCase)
   }
 
+  /**
+   * Filters columns based on the provided criteria.
+   *
+   * @param schema            The generic schema containing columns.
+   * @param columnsFilterList The list of columns to include or exclude.
+   * @param includeColumns    If true, includes specified columns; if false, excludes them.
+   * @return A sequence of filtered column names.
+   */
   def filterColumns(schema: GenericSchema, columnsFilterList: Seq[String], includeColumns: Boolean = true): Seq[String] = {
     val columns = schema.columns
     val filteredColumns =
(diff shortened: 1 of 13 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/misc/GenericSchemaUtil.scala
```diff
@@ -0,0 +1,70 @@
...
+import org.slf4j.LoggerFactory
+
+object GenericSchemaUtil {
+
+  def columnExists(schema: GenericSchema, colName: String): Boolean = {
+    if (Environment.caseSensitive) schema.columns.contains(colName)
+    else schema.columns.contains(colName.toLowerCase)
+  }
+
+  def filterColumns(schema: GenericSchema, columnsFilterList: Seq[String], includeColumns: Boolean = true): Seq[String] = {
```
zzeekk, line 41: Would be nice to add some scaladoc for this function, explaining parameter includeColumns.

zzeekk (approved):
LGTM! just a minor comment about adding some scala doc.
Make sure to use "Squash and merge" to merge the PR, otherwise we will get lots of commits on the develop branch.
And add the Issue number to the commit message with convention "some text (#755)".

sdl-core/src/main/scala/io/smartdatalake/util/misc/GenericSchemaUtil.scala
ThierryGirod: Fixed in commit c92d31081886560b062c25d383f1d61eaeb02d90

PR #763, opened 2023-11-23
improve scope for spark provided dependencies

zzeekk:
Improve scope for spark provided dependencies

Reducing size of the fat-jar (without spark)
Reducing possibility for library version conflict with spark distribution.

### DIFF
--- a/sdl-core/pom.xml
+++ b/sdl-core/pom.xml
@@ -111,19 +111,17 @@
             <scope>${spark.deps.scope}</scope>
         </dependency>
 
-        <!-- this needs to have compile scope, if it's runtime it doesnt compile anymore -->
         <!-- this is an unused declared dependency, but it's needed to override scope -->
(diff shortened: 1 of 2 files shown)


sdl-core/pom.xml
```diff
@@ -111,19 +111,17 @@
             <scope>${spark.deps.scope}</scope>
         </dependency>
 
-        <!-- this needs to have compile scope, if it's runtime it doesnt compile anymore -->
```
Geheiner, line 114: hmm this comment is not true anymore?

Geheiner (approved):
LGTM if my comment can be ignored...

sdl-core/pom.xml
zzeekk: now scope is compile or provided, both are ok. The comment is no longer needed in my opinion.

PR #768, opened 2023-11-30
Bugfix: reset skipped subfeeds

zzeekk:
Fix for #770 

Additionally i added a DebugTransformer to simplify logging schema, data and explain plan through a slim configuration change.


pgruetter (approved):
Looks good and I don't see any side effects.

Can't test the DebugTransformer at the moment since my local setup seems to be broken with SDL 2.6.X, need to check that. But very nice addition!

One note: Would be good to create issues for bugs like that, makes it easier to track / link in the release.

zzeekk:
> One note: Would be good to create issues for bugs like that, makes it easier to track / link in the release.

Thanks, you're right. I created #770 and updated the commit message.

Issue #764, opened 2023-11-26
Migrate state files from older versions


zzeekk:
**Is your feature request related to a problem? Please describe.**
With new SDLB versions state file format changes. There is yet no automated process to migrate old state files to the current version.

For instance with SDLB version 2.6.0-SNAPSHOT and runStateFormatVersion=4 the following changes have been applied:
- Instead of having `results.subFeed`, the attributes of `subFeed` have been moved directly into `result`.
- `actionState.inputIds[].id` was cleaned up as `inputIds[]`, and the the same applies for `actionState.outputIds`

State files from version 2.5.x can currently not be read with version 2.6.0-SNAPSHOT

**Describe the solution you'd like**
Old versions should be migrated to current version automatically on read.
The state files version can be read from the top-level attribute runStateFormatVersion.
A list of state migrators should be applied to convert old versions automated and version by version to the current version.
Further a command line tool should be created to migrate all files in a folder to the current version.

PR #769, opened 2023-12-02
Feature/764 state migrator

zzeekk:
see #764

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
@@ -151,6 +151,7 @@ private[smartdatalake] object ActionDAGRunState extends SmartDataLakeLogger {
     val attemptId = jValue \ "attemptId" match {
       case JInt(i) => i.toInt
     }
+    assert(formatVersion <= runStateFormatVersion, s"Cannot read state file with formatVersion=${formatVersion} newer than the version of this build (${runStateFormatVersion}). Check state file app=$appName runId=$runId attemptId=$attemptId and that your SDLB version is up-to-date!")
     val migrators = stateMigrators.dropWhile(m => m.versionFrom <= formatVersion)
     if (migrators.nonEmpty) {
       logger.info(s"Applying state migrators ${migrators.mkString(", ")} to state json for app=$appName runId=$runId attemptId=$attemptId")
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
@@ -27,7 +27,8 @@ import io.smartdatalake.workflow.action.{ExecutionId, RuntimeEventState, Runtime
 import org.apache.spark.util.Json4sCompat
 import org.json4s._
 import org.json4s.ext.EnumNameSerializer
-import org.json4s.jackson.Serialization.{read, write, writePretty}
+import org.json4s.jackson.JsonMethods
+import org.json4s.jackson.Serialization.{write, writePretty}
 import org.reflections.Reflections
 
 import java.time.{Duration, LocalDateTime}
@@ -74,8 +75,9 @@ case class DataObjectState(dataObjectId: DataObjectId, state: String) {
   def toStringTuple: (String,String) = (dataObjectId.id, state)
 }
 
-private[smartdatalake] object ActionDAGRunState {
+private[smartdatalake] object ActionDAGRunState extends SmartDataLakeLogger {
 
+  // Note: if increasing this version, please check if a StateMigrator is needed to read files of older versions. See also stateMigrators below.
   val runStateFormatVersion: Int = 4
 
   private val durationSerializer = Json4sCompat.getCustomSerializer[Duration](formats => (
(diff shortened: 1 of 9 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/ActionDAGRunState.scala
```diff
@@ -125,11 +133,47 @@ private[smartdatalake] object ActionDAGRunState {
...
+      logger.info(s"Applying state migrators ${migrators.mkString(", ")} to state json for app=$appName runId=$runId attemptId=$attemptId")
+      Some(migrators.foldLeft(jValue)((v, m) => m.migrate(v)))
+    } else None
+  }
+
+  // list of state migrators, sorted in ascending order
+  private val stateMigrators: Seq[StateMigratorDef] = Seq(
+    new StateMigratorDef3To4()
+  ).sortBy(_.versionFrom) // force ordering
+  assert(stateMigrators.groupBy(_.versionFrom).forall(_._2.size == 1)) // check that versionFrom is unigue
```
pgruetter:
```suggestion
  assert(stateMigrators.groupBy(_.versionFrom).forall(_._2.size == 1)) // check that versionFrom is unique
```

sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
```diff
@@ -0,0 +1,81 @@
...
+  def versionTo: Int
+  def migrate(json: JValue): JValue
+  override def toString: String = s"$versionFrom -> $versionTo"
+}
+
+/**
+ * Migrate state from format version 3 to 4:
+ * - the structure of `actionState.results` was cleaned up.
+ *   instead of having `results.subFeed`, the attributes of `subFeed` have been moved directly into `result`.
+ * - Additionally `actionState.inputIds[].id` was cleaned up as `inputIds[]`, and the the same applies for `actionState.outputIds`.
```
pgruetter:
```suggestion
 * - Additionally `actionState.inputIds[].id` was cleaned up as `inputIds[]`, and the same applies for `actionState.outputIds`.
```

sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
```diff
@@ -0,0 +1,81 @@
...
+ */
+class StateMigratorDef3To4 extends StateMigratorDef with SmartDataLakeLogger {
+  override val versionFrom = 3
+  override val versionTo = 4
+  override def migrate(json: JValue): JValue = {
+    assert(json \ "runStateFormatVersion" match { case JInt(version) => version <= versionFrom}, s"Version should be equals or less than $versionFrom")
```
pgruetter: Shouldn't we assume that version = versionFrom ? The upgrade procedure is probably strictly working for one version only.

sdl-lang/src/main/scala/io/smartdatalake/meta/state/StateMigrator.scala
```diff
@@ -0,0 +1,86 @@
...
+import io.smartdatalake.util.hdfs.HdfsUtil
+import io.smartdatalake.util.hdfs.HdfsUtil.RemoteIteratorWrapper
+import io.smartdatalake.util.misc.SmartDataLakeLogger
+import io.smartdatalake.workflow.ActionDAGRunState.checkStateFormatVersionAndMigrate
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.json4s.jackson.{JsonMethods, prettyJson}
+import scopt.OptionParser
+
+case class StateMigratorConfig(statePath: Option[String] = None)
```
pgruetter, line 38: Can you please add an overall ScalaDoc ?

pgruetter (commented):
Great feature, just added a few comments. 
The assumption should be, that we strictly update one version to the next IMHO.

sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
zzeekk: Then we could not migrate from version 1/2 to version 4. As there are no changes needed to migrate from version 1to2 or 2to3, no StateMigratorDef exists for these steps.

sdl-lang/src/main/scala/io/smartdatalake/meta/state/StateMigrator.scala
zzeekk: Yes, sure. Done.

sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
pgruetter: OK, makes sense. Thanks!

pgruetter (approved):

iliassoto:
I did some testing for this feature. The mentioned files are attached:

1. I ran "myApp.conf" and with SDLB 2.3.0., resulting in the state "myAppName.1.1.json"
2. I then created a new project with the SDLB built from this branch, and tried to run the pipeline again using the old state file, where an error was thrown (see file). 

The JSON itself is able to be parsed: 

```scala
  // read state from json
def fromJson(stateJson: String): ActionDAGRunState = {
      val jValue = JsonMethods.parse(stateJson) //WORKS
      val migratedJValue = checkStateFormatVersionAndMigrate(jValue).getOrElse(jValue) //ERROR HAPPENS HERE
      // extract into class structures
      migratedJValue.extract[ActionDAGRunState]
    } catch {
      case ex: Exception => throw new IllegalStateException(s"Unable to parse state from json: ${ex.getMessage}", ex)
    }
  }
```

[error.log](https://github.com/smart-data-lake/smart-data-lake/files/14001698/error.log)
[myApp.hocon.txt](https://github.com/smart-data-lake/smart-data-lake/files/14001699/myApp.hocon.txt)
[myAppName.1.1.json](https://github.com/smart-data-lake/smart-data-lake/files/14001700/myAppName.1.1.json)

The operation _json \ "runStateFormatVersion"_ results in a _JNothing_-Type, which throws the error during the assert.

```scala
  override def migrate(json: JValue): JValue = {
    assert(json \ "runStateFormatVersion" match { case JInt(version) => version <= versionFrom}, s"Version should be equals or less than $versionFrom")
    json.transformField {
      case (name, actionsState) if name == "actionsState" =>
        (name, actionsState.transformField {
          case (name, results) if name == "results" =>
            (name, JArray(results.children.map {
              result =>
```

I guess we can just add another case that also handles JNothing for versions that don't have that node in their JSON files?

zzeekk:
> ```scala
> I guess we can just add another case that also handles JNothing for versions that don't have that node in their JSON files?
> ```
You're right, early state files without version attribute produce an error. I added the JNothing case, and fixed adding the version attribute afterwards.

iliassoto (approved):

PR #772, opened 2023-12-06
column encrypt: add ECB algorithm

mand35:
In the column encryption transformer another algorithm is implemented: AES/ECB/PKCS5Padding
It is of minor cryptographic complexity, but it allows to encrypt same items with same ciphertext. 

When encrypting columns sometimes we need to be able to aggregate over encrypted values. Therefore, we need same values to be encrypted with the same ciphertext.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
@@ -32,63 +32,83 @@ import java.security.SecureRandom
 import java.util.Base64
 import javax.crypto.spec.{GCMParameterSpec, SecretKeySpec}
 import javax.crypto.{Cipher, SecretKey}
-import org.apache.spark.sql.functions.{col, udf}
+import org.apache.spark.sql.functions.{col, lit, udf}
 
-trait EncryptDecrypt {
-  def keyAsBytes: Array[Byte]
+trait EncryptDecrypt extends Serializable {
+  protected def keyAsBytes: Array[Byte]
+  protected def encryptUDF: UserDefinedFunction
+  protected def decryptUDF: UserDefinedFunction
 
-  val cryptUDF: UserDefinedFunction = udf(encryptGCM _)
-  private val GCM_ALGORITHM_STRING: String = "AES/GCM/NoPadding"
-  private val ECB_ALGORITHM_STRING: String = "AES/ECB/PKCS5Padding"
-  private val IV_SIZE = 128
-  private val TAG_BIT_LENGTH = 128
-
-  private val secureRandom = new SecureRandom()
-
-
-  def process(df: DataFrame, encryptColumns: Seq[String]): DataFrame = {
+  def encrypt(df: DataFrame, encryptColumns: Seq[String]): DataFrame = {
     encryptColumns.foldLeft(df) {
-      case (dfTemp, colName) => dfTemp.withColumn(colName, cryptUDF(col(colName)))
+      case (dfTemp, colName) => dfTemp.withColumn(colName, encryptUDF(col(colName)))
     }
   }
 
-  private def generateAesKey(keyBytes: Array[Byte]): SecretKey = {
+  def decrypt(df: DataFrame, encryptColumns: Seq[String]): DataFrame = {
+    encryptColumns.foldLeft(df) {
+      case (dfTemp, colName) => dfTemp.withColumn(colName, decryptUDF(col(colName)))
+    }
+  }
+
+  def generateAesKey(keyBytes: Array[Byte]): SecretKey = {
     new SecretKeySpec(keyBytes, "AES")
   }
 
-  def encryptECB(message: String): String = {
-    val cipher: Cipher = Cipher.getInstance(ECB_ALGORITHM_STRING)
-    val secretKey = generateAesKey(keyAsBytes)
+}
+
+class EncryptDecryptECB(keyBytes: Array[Byte]) extends EncryptDecrypt {
+  override protected val keyAsBytes: Array[Byte] = keyBytes
+  private val ALGORITHM_STRING: String = "AES/ECB/PKCS5Padding"
+  override def encryptUDF: UserDefinedFunction = udf(encryptFunc _)
+  override def decryptUDF: UserDefinedFunction = udf(decryptFunc _)
+
+
+  private def encryptFunc(message: String): String = {
+    val cipher: Cipher = Cipher.getInstance(ALGORITHM_STRING)
+    val secretKey: SecretKey = generateAesKey(keyAsBytes)
     cipher.init(Cipher.ENCRYPT_MODE, secretKey)
     val data = cipher.doFinal(message.getBytes())
     Base64.getEncoder.encodeToString(data)
   }
 
-  def decryptECB(encryptedDataString: String): String = {
+  private def decryptFunc(encryptedDataString: String): String = {
     val data = Base64.getDecoder.decode(encryptedDataString)
-    val cipher: Cipher = Cipher.getInstance(ECB_ALGORITHM_STRING)
-    val secretKey = generateAesKey(keyAsBytes)
+    val cipher: Cipher = Cipher.getInstance(ALGORITHM_STRING)
+    val secretKey: SecretKey = generateAesKey(keyAsBytes)
     cipher.init(Cipher.DECRYPT_MODE, secretKey)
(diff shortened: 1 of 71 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
```diff
@@ -106,19 +123,27 @@ trait EncryptDecrypt {
...
   extends SparkDfTransformer with EncryptDecrypt {
   private val cur_key: StringOrSecret = key.getOrElse(SecretsUtil.convertSecretVariableToStringOrSecret(keyVariable.get))
 
   override def keyAsBytes: Array[Byte] = cur_key.resolve().getBytes
 
-  override val cryptUDF: UserDefinedFunction = udf(encrypt _)
+  override val cryptUDF: UserDefinedFunction = algorithm match {
+    case "GCM" => udf(encryptGCM _)
+    case "ECB" => udf(encryptECB _)
+    case _ => {assert(true, s"unsupported en/decryption algorithm ${algorithm}, using GCM")
```
Geheiner: So basically you could just add a warning instead of making an assert

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
```diff
@@ -141,20 +166,28 @@ object EncryptColumnsTransformer extends FromConfigFactory[GenericDfTransformer]
...
   private val cur_key: StringOrSecret = key.getOrElse(SecretsUtil.convertSecretVariableToStringOrSecret(keyVariable.get))
 
   override def keyAsBytes: Array[Byte] = cur_key.resolve().getBytes
 
-  override val cryptUDF: UserDefinedFunction = udf(decrypt _)
+  override val cryptUDF: UserDefinedFunction = algorithm match {
+    case "GCM" => udf(decryptGCM _)
+    case "ECB" => udf(decryptECB _)
+    case _ => {
+      assert(true, s"unsupported en/decryption algorithm ${algorithm}, using GCM")
```
Geheiner: Same as above

Geheiner (approved):
Lgtm, just remove the assert

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
```diff
@@ -56,22 +56,39 @@ trait EncryptDecrypt {
     new SecretKeySpec(keyBytes, "AES")
   }
 
-  def encrypt(message: String): String = {
+  def encryptECB(message: String): String = {
+    val cipher: Cipher = Cipher.getInstance(ECB_ALGORITHM_STRING)
+    val secretKey = generateAesKey(keyAsBytes)
+    cipher.init(Cipher.ENCRYPT_MODE, secretKey)
```
zzeekk: For performance reasons i suggest to initialize the cipher outside this function, otherwise this is done for every row.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
```diff
@@ -37,16 +37,16 @@ import org.apache.spark.sql.functions.{col, udf}
 trait EncryptDecrypt {
   def keyAsBytes: Array[Byte]
 
-  val cryptUDF: UserDefinedFunction = udf(encrypt _)
-  private val ALGORITHM_STRING: String = "AES/GCM/PKCS5Padding"
+  val cryptUDF: UserDefinedFunction = udf(encryptGCM _)
```
zzeekk: Why is this statically initialized to GCM?

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
```diff
@@ -56,22 +56,39 @@ trait EncryptDecrypt {
...
-  def encrypt(message: String): String = {
+  def encryptECB(message: String): String = {
+    val cipher: Cipher = Cipher.getInstance(ECB_ALGORITHM_STRING)
+    val secretKey = generateAesKey(keyAsBytes)
+    cipher.init(Cipher.ENCRYPT_MODE, secretKey)
+    val data = cipher.doFinal(message.getBytes())
+    Base64.getEncoder.encodeToString(data)
+  }
+
+  def decryptECB(encryptedDataString: String): String = {
```
zzeekk: To apply object oriented pattern you could:
Instead of mixing GCM and ECB in the same "trait", have a class implementing GCM and another one implementing ECB, both extending a common base trait EncryptDecrypt.
Then you could choose the implementation class in EncryptColumnsTransformer.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
AlexandreDelezeElca: Can be kind of agree for performance reason. But then, according to the [documentation ](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/javax/crypto/Cipher.html#doFinal()), you need to handle any exceptions of the method doFinal. Otherwise it will make crash your implementation.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
```diff
@@ -106,19 +123,27 @@ trait EncryptDecrypt {
...
  * @param keyVariable    contains the id of the provider and the name of the secret with format <PROVIDERID>#<SECRETNAME>,
  *                       e.g. ENV#<ENV_VARIABLE_NAME> to get a secret from an environment variable OR CLEAR#mYsEcReTkeY
+ * @param algorithm      There are GCM (AES/GCM/NoPadding) and ECB (AES/ECB/PKCS5Padding) implemented. DEFAULT: GCM
  */
 case class EncryptColumnsTransformer(override val name: String = "encryptColumns",
                                      override val description: Option[String] = None,
                                      encryptColumns: Seq[String],
                                      @Deprecated @deprecated("Use `key` instead", "2.5.0") private val keyVariable: Option[String] = None,
                                      private val key: Option[StringOrSecret],
+                                     algorithm: String = "GCM"
```
AlexandreDelezeElca, line 48: Here should be ECB in your case

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
AlexandreDelezeElca: Here is probably more an exception than an assert. Assertion should not be used for production code and used instead of exceptions. Assertion can be disabled depending of the configuration to have performance benefits.

AlexandreDelezeElca (commented):
Consider to remove GCM if not used in the future

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
mand35: thanks for the hints, indeed this needs to be an exception, I corrected it.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
mand35: that was meant to be overriden, but with an default value, but now its handled differently. 👍

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/EncryptColumnsTransformer.scala
mand35: done.

mand35:
> Consider to remove GCM if not used in the future

No, GCM should be still the default encryption method, since it is the strongest. Nevertheless, we need to provide both methods.

mand35:
due to merge issues this PR will be closed, a proper new one is opened (PR#777)

PR #775, opened 2023-12-18
Improve and unify dataobject schema scaladoc

zzeekk:
Improve scaladoc description of schema providers available to define DataObject schemas (DataObject.schema attribute).
Cleanup using super-method Scaladoc, instead of rewriting attribute descriptions in every DataObject (Schema export is searching scaladoc on super-method if not present on case class directly).

Fix marking CustomDataFrameAction.transformer as deprecated in Schema Viewer.

Schema provider options where not described previously.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/JsonFileDataObject.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/JsonFileDataObject.scala
@@ -39,20 +39,9 @@ import org.apache.spark.sql.DataFrame
  * Reading and writing details are delegated to Apache Spark [[org.apache.spark.sql.DataFrameReader]]
  * and [[org.apache.spark.sql.DataFrameWriter]] respectively.
  *
- * @param stringify Set the data type for all values to string.
+ * @param stringify Set the data type for all values to string. Use action/transformers instead.
  * @param jsonOptions Settings for the underlying [[org.apache.spark.sql.DataFrameReader]] and
  *                    [[org.apache.spark.sql.DataFrameWriter]].
- * @param schema An optional data object schema. If defined, any automatic schema inference is avoided.
- *               As this corresponds to the schema on write, it must not include the optional filenameColumn on read.
- *               Define the schema by using one of the schema providers DDL, jsonSchemaFile, xsdFile or caseClassName.
- *               The schema provider and its configuration value must be provided in the format <PROVIDERID>#<VALUE>.
- *               A DDL-formatted string is a comma separated list of field definitions, e.g., a INT, b STRING.
- * @param sparkRepartition Optional definition of repartition operation before writing DataFrame with Spark to Hadoop.
- * @param expectedPartitionsCondition Optional definition of partitions expected to exist.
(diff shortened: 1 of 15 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/JsonFileDataObject.scala
```diff
@@ -67,7 +56,7 @@ case class JsonFileDataObject( override val id: DataObjectId,
                                override val schemaMin: Option[GenericSchema] = None,
                                override val saveMode: SDLSaveMode = SDLSaveMode.Overwrite,
                                override val sparkRepartition: Option[SparkRepartitionDef] = None,
-                               stringify: Boolean = false,
+                               @Deprecated @deprecated("Use action/transformers instead", "2.6.0") stringify: Boolean = false,
```
pgruetter, line 59: Maybe a stupid question, but is the double deprecated necessary?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/RawFileDataObject.scala
```diff
@@ -32,17 +32,13 @@ import io.smartdatalake.workflow.dataframe.spark.SparkSchema
...
- * @param customFormat Custom Spark data source format, e.g. binaryFile or text. Only needed if you want to read/write this DataObject with Spark.
- * @param options Options for custom Spark data source format. Only of use if you want to read/write this DataObject with Spark.
- * @param fileName Definition of fileName. This is concatenated with path and partition layout to search for files. Default is an asterix to match everything.
- * @param saveMode Overwrite or Append new data.
- * @param expectedPartitionsCondition Optional definition of partitions expected to exist.
- *                                    Define a Spark SQL expression that is evaluated against a [[PartitionValues]] instance and returns true or false
- *                                    Default is to expect all partitions to exist.
- * @param housekeepingMode Optional definition of a housekeeping mode applied after every write. E.g. it can be used to cleanup, archive and compact partitions.
- *                         See HousekeepingMode for available implementations. Default is None.
+ * By specifying format binary or text files can used with Spark.
```
pgruetter: Not clear to me, maybe:
```suggestion
 * By specifying a customFormat binaryFile or Text, files can be used with Spark.
```
?

pgruetter (approved):
See comments

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/JsonFileDataObject.scala
zzeekk: Uppercase "@Deprecated" is the java annotation, and lowercase is Scala.
Normally we just use the scala variant, but in order to mark the attribute in the schema export / schema viewer as deprecated we need also the Java variant. This is due to the Scala variant being only available at compile time, but not for reflection in the schema export which runs as separate cmd after compilation...

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/RawFileDataObject.scala
zzeekk: True, not so clear.
I improved similar to your suggestion.

PR #777, opened 2023-12-20
Column encrypt ECB and Databricks UDF preparation

mand35:
# What changes are included in the pull request?
In the column encryption transformer another algorithm is implemented: AES/ECB/PKCS5Padding
It is of minor cryptographic complexity, but it allows to encrypt same items with same ciphertext.
Further a package for deploying on Databricks as function is provided. 

# Why are the changes needed?
When encrypting columns sometimes we need to be able to aggregate over encrypted values. Therefore, we need same values to be encrypted with the same ciphertext.

This is kind of duplicate of PR #772 , but unfortunately the other one got messed up with a unwanted merge.

### DIFF
--- a/sdl-core/pom.xml
+++ b/sdl-core/pom.xml
@@ -469,6 +469,22 @@
                             </excludes>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>crypt-jar</id>
+                        <phase>package</phase>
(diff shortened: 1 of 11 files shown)


zzeekk (approved):
Nice!

Geheiner (approved):

Issue #707, opened 2023-07-24
Release for Scala 2.13


zzeekk:
Spark is supporting Scala 2.13 for a while now.
We should release SDLB for Scala 2.13 as well.
To be checked if all dependencies support Scala 2.13 already.

**Additional context**
A spark-extensions Release for Spark 3.4 / Scala 2.13 is already released.
See the following commit for changes needed: https://github.com/smart-data-lake/spark-extensions/commit/449309a9142eab8bc83050ff164641b290b3357a#diff-ef8e27cd4720dfae3562d9c560e82c4623a726b30899c23a90af9f6f371aab92
- import scala.collection.JavaConverters._ -> import scala.jdk.CollectionConverters._: needs scala-collection-compat_${scala.minor.version} to work in Scala 2.12 as well
- methods with an empty parameter list should be call including it, e.g. method(), otherwise there is a warning, and in Scala 3 this will result in a compile error

PR #778, opened 2024-01-02
Compile for Scala 2.13 (#707)

zzeekk:
see #707

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,15 +63,14 @@ jobs:
         maven_profiles: scala-2.12,release-sonatype
         maven_args: -B -f pom.xml
 
-    - name: Maven deploy to sonatype for Scala 2.11, only if on master-spark2* branch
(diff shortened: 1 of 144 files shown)


pgruetter (approved):
Works for me! Tested with profile 2.13 with JDK 11. Ran some tests with Delta lake and it works correctly. 
Only issue I had are errors about scoverage plugin dependency that it can't find. The error is gone if I activate the test-sonar profile but I didn't see that before.

pgruetter:
Ah, you can also see the problem with scoverage on this page: the coverage seems to be 0.0% (see SonarCloud analysis). It looks like this is not a SonarCloud problem. The coverage generated locally also produces 0.0% in all classes.

zzeekk:
> Ah, you can also see the problem with scoverage on this page: the coverage seems to be 0.0%
Maybe a problem because i disabled scoverage:report for scala-2.13, but did not remove sonarcube upload? I try to remove this as well, let's see.

Issue #711, opened 2023-08-09
Implement ScalaClassSparkDfNTo1Transformer


Geheiner:
**Is your feature request related to a problem? Please describe.**
DF=DataFrame 
DO=DataObject

When using a CustomDfsTransformer you currently need to provide a Map to tell SDLB which output DF belongs to which DO.
It's best practice to have one Output DO on an Action, so that's the most common case that we encounter.
And when you only have one output DF, the mapping to the DO can be done automatically.

**Describe the solution you'd like**
This is why I propose a new kind of Transformer that does just that:  it takes the same map of dfs to resolve the inputs, but directly accepts a DF as an output.
This will be the signature of transform:
def transform(session: SparkSession, options: Map[String, String], dfs: Map[String, DataFrame]): DataFrame 

This makes the Action writing less error prone.

**Additional context**
Similar efforts were already done with https://github.com/smart-data-lake/smart-data-lake/issues/489 for Datasets.
There we were able to go one step further and also not need the map for the inputs because the DF to DO Mapping could be inferred thanks to the provided case class. For DFs we don't have a case class and just relying on DF/DO Ordering is too risky. That's why I propose to restrict this feature to just automatize the output DF/DO Mapping

zzeekk:
Good idea. I have further suggestions to simplify implementation of a CustomDfsTransformer.
We could provide a default implementation of `CustomDfsTransformer.transform(session: SparkSession, options: Map[String,String], dfs: Map[String,DataFrame]) : Map[String,DataFrame]`, which dynamically searchs for another transform method in the Transformer implementation and dynamically maps paramters.

Example:
The implementation of CustomDfsTransformer could have a method:
`transform(session: SparkSession, dfDataObject1: DataFrame, dfDataObject2: DataFrame, isExec: Boolean, nbOfDts: Option[Int]): DataFrame`
The default implementation of the main CustomDfsTransformer.transform method would map:
- dfs('data-object-1') to dfDataObject1 parameter
- dfs('data-object-2') to dfDataObject2 parameter
- options('isExec').toBoolean to isExec parameter
- options.get('nbOfDts').map(_.toInt) to nbOfDts parameter
- as the function returns only one DataFrame, a Map(outputIds.head, df) can be created as final return value.

The mapping is possible through reflection of parameter types and names. For options we should support primitive types like String, Boolean, Short, Int, Long, Float, Double.

zzeekk:
[...]

PR #779, opened 2024-01-14
SparkDfsTransformer: Implement dynamic custom transform function (#711)

zzeekk:
see #711

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/misc/CustomCodeUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/misc/CustomCodeUtil.scala
@@ -20,8 +20,13 @@ package io.smartdatalake.util.misc
 
 import io.smartdatalake.config.ConfigurationException
 import io.smartdatalake.definitions.Environment
+import io.smartdatalake.workflow.action.spark.customlogic.NotFoundError
+import io.smartdatalake.workflow.action.spark.transformer.ScalaClassSparkDsNTo1Transformer.tolerantGet
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+import org.apache.spark.sql.functions.col
 
 import java.io.{FileNotFoundException, InputStream}
+import scala.reflect.runtime.universe
 import scala.tools.reflect.ToolBox
 import scala.util.{Failure, Success, Try}
 
@@ -31,8 +36,10 @@ import scala.util.{Failure, Success, Try}
   */
 private[smartdatalake] object CustomCodeUtil {
 
+  private val runtimeMirror = scala.reflect.runtime.currentMirror
+
   // get Scala Toolbox to compile code at runtime
-  private lazy val tb = scala.reflect.runtime.currentMirror.mkToolBox()
+  private lazy val tb = runtimeMirror.mkToolBox()
 
   /**
    * Compiling Scala Source Code into Object of Type T
@@ -70,4 +77,61 @@ private[smartdatalake] object CustomCodeUtil {
     // return value
     content
   }
+
+  /**
+   * Get method symbol with given name from class definition
+   */
+  def getClassMethodsByName(cls: Class[_], methodName: String): scala.Seq[universe.MethodSymbol] = {
+    val mirror = scala.reflect.runtime.currentMirror
+    val classType = mirror.classSymbol(cls).toType
+    val t = classType.members.filter(_.isMethod).filter(_.name.toString == methodName)
+    classType.members.filter(_.isMethod).filter(_.name.toString == methodName).map(_.asMethod).toSeq
+  }
+
+  /**
+   * Extract default values for parameters of a method signature.
+   * @param instance class instance of object the method belongs to
+   * @param method method symbol to read signature from
+   * @return a Map with parameter names and their default values.
+   */
+  def getMethodParameterDefaultValues(instance: Any, method: universe.MethodSymbol): Map[String, Any] = {
+    val instanceMirror = runtimeMirror.reflect(instance)
+    val classType = instanceMirror.symbol.toType
+    method.paramLists.head.zipWithIndex.flatMap {
+      case (p,i) =>
+        val parameterName = p.name.toString
+        // There is a special method for getting the default value for a given parameter.
+        // The method name for default values is by convention, see also https://stackoverflow.com/questions/13812172/how-can-i-create-an-instance-of-a-case-class-with-constructor-arguments-with-no
+        val defaultMethod = classType.member(universe.TermName(s"${method.name.toString}$$default$$${i+1}"))
+        if (defaultMethod != universe.NoSymbol) {
+          val defaultMethodMirror = instanceMirror.reflectMethod(defaultMethod.asMethod)
+          Some((parameterName, defaultMethodMirror.apply()))
+        } else None
+    }.toMap
+  }
+
+  /**
+   * Dynamically call method on class instance.
+   */
+  def callMethod[R](instance: Any, methodSymbol: universe.MethodSymbol, args: Seq[Any]): R = {
+    val instanceMirror = runtimeMirror.reflect(instance)
+    instanceMirror.reflectMethod(methodSymbol).apply(args:_*).asInstanceOf[R]
+  }
+
+  /**
+   * Extract method parameters with default values through reflection.
+   * @param instance: class instance for method to inspect
+   * @param method: method symbol to inspect
+   */
+  def analyzeMethodParameters(instance: Any, method: universe.MethodSymbol): Seq[MethodParameterInfo] = {
+    val parameters = method.paramLists.head
+    val defaultValues = getMethodParameterDefaultValues(instance, method)
+    parameters.map { p =>
+      MethodParameterInfo(p.name.toString, p.typeSignature, defaultValues.get(p.name.toString))
+    }
+  }
+}
+
+case class MethodParameterInfo(name: String, tpe: universe.Type, defaultValue: Option[Any]) {
+  def toMap: Map[String, Any] = Seq(Some("name" -> name), Some("type" -> tpe.toString), defaultValue.map("default" -> _)).flatten.toMap
 }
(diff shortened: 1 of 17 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/misc/CustomCodeUtil.scala
```diff
@@ -70,4 +77,61 @@ private[smartdatalake] object CustomCodeUtil {
...
     content
   }
+
+  /**
+   * Get method symbol with given name from class definition
+   */
+  def getClassMethodsByName(cls: Class[_], methodName: String): scala.Seq[universe.MethodSymbol] = {
+    val mirror = scala.reflect.runtime.currentMirror
+    val classType = mirror.classSymbol(cls).toType
+    val t = classType.members.filter(_.isMethod).filter(_.name.toString == methodName)
```
Geheiner: unused variable, you can remove it

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDfsTransformer.scala
```diff
@@ -19,27 +19,108 @@
...
    * @param session Spark Session
    * @param options Options specified in the configuration for this transformation
    * @param dfs DataFrames to be transformed
-   * @return Transformed DataFrame
+   * @return Map of transformed DataFrames
    */
-  def transform(session: SparkSession, options: Map[String,String], dfs: Map[String,DataFrame]) : Map[String,DataFrame]
+  def transform(session: SparkSession, options: Map[String,String], dfs: Map[String,DataFrame]) : Map[String,DataFrame] = {
+    val transformMethod = customTransformMethod.get
+    assert(transformMethod.returnType =:= typeOf[Map[String, DataFrame]], s"The return type of the transform method should be Map[String,DataFrame], but is ${transformMethod.returnType}")
```
Geheiner: it's a bit picky but from a scala point of view you should use require here as you are validating the input parameters of yours method. I let you the choice if you want to leave it like this as I have seen this in several places but it always looks strange to me to find asserts in the code, do we sometimes run sdlb with the asserts diabled? Probably not for this one!

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDfsTransformer.scala
```diff
@@ -19,27 +19,108 @@
...
+        val targetException = e.getTargetException
+        targetException.setStackTrace(e.getTargetException.getStackTrace ++ e.getStackTrace)
+        throw targetException
+    }
+
+    // distinguish between returning single and multiple DataFrames.
+    if (transformMethod.returnType =:= typeOf[Map[String, DataFrame]]) {
+      transformResult.asInstanceOf[Map[String, DataFrame]]
+    } else if (transformMethod.returnType <:< typeOf[Map[String, Dataset[_]]]) {
+      transformResult.asInstanceOf[Map[String, Dataset[_]]].mapValues(_.toDF)
```
Geheiner, line 115: It's nice that you extract the dataset type of the return dataset, but it gets converted right back to a dataframe, so basically we could also not make the distinction between dataset and dataframe in that part of the code. Or do you forsee a possible extension that you want to prepare with this?

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDfsTransformer.scala
```diff
@@ -54,6 +135,22 @@ trait CustomDfsTransformer extends Serializable {
    */
   def transformPartitionValues(options: Map[String, String], partitionValues: Seq[PartitionValues]): Option[Map[PartitionValues,PartitionValues]] = None
 
+  // lookup custom transform method
+  override private[smartdatalake] lazy val customTransformMethod = {
+    val transformMethods = CustomCodeUtil.getClassMethodsByName(getClass, "transform")
+    val (defaultTransformMethods,subClassTransformMethods) = transformMethods
+      .partition(_.owner == typeOf[CustomDfsTransformer].typeSymbol)
+    assert(defaultTransformMethods.size == 1) // this is the default implementation of CustomDfsTransformer
```
Geheiner: again for me this would be by the scala book a require

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDfsTransformer.scala
```diff
@@ -32,8 +32,22 @@ import org.apache.spark.sql.DataFrame
...
+ * To define the transformation a class implementing the trait [[CustomDfsTransformer]] has to be created.
+ * There are two methods to define of defining transformation:
+ *
+ * 1) Overwrite the generic transform function of CustomDfsTransformer: Define a transform function which receives a map of input DataObjectIds with DataFrames and a map of options and has
+ * to return a map of output DataObjectIds with DataFrames. The exact signature is `transform(session: SparkSession, options: Map[String,String], dfs: Map[String,DataFrame]): Map[String,DataFrame]`.
+ *
+ * 2) Implement any transform method using parameters of type SparkSession, Map[String,String], DataFrame, Dataset[<Product>] and any primitive data type (String, Boolean, Int, ...).
+ * Primitive data types might also use default values or be enclosed in an Option[...] to mark it as non required.
+ * The transform method is then called dynamically by looking for the parameter values in the input DataFrames and Options.
+ * Using this method you can avoid writing code to prepre DataFrames and Options from the corresponding Map parameters.
```
Geheiner:
```suggestion
 * Using this method you can avoid writing code to prepare DataFrames and Options from the corresponding Map parameters.
```

sdl-core/src/test/scala/io/smartdatalake/workflow/action/spark/transformer/CustomDfsTransformerTest.scala
```diff
@@ -0,0 +1,76 @@
...
+      .filter(_.owner != typeOf[CustomDfsTransformer].typeSymbol)
+    val transformMethod = transformMethodsOfSubclass.head
+    val params = CustomCodeUtil.analyzeMethodParameters(transformer, transformMethod)
+    assert(params.find(_.name == "defaultBoolean").get.defaultValue.nonEmpty)
+  }
+
+  test("CustomDfsTransformer can dynamically map parameters with default values or optional") {
+    val transformer = new DynamicTransformer(true, None, true, 1L, None)
+    val options = Map("isExec" -> "true", "long" -> "1")
+    val df1 = Seq(("x",1)).toDF("a","b")
```
Geheiner: please run the formatter on this file

sdl-lang/src/test/scala/io/smartdatalake/meta/configexporter/DynamicTestTransformer.scala
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.meta.configexporter
+
+import io.smartdatalake.workflow.action.spark.customlogic.CustomDfsTransformer
+import io.smartdatalake.workflow.action.spark.transformer.Test
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+
+class DynamicTestTransformer extends CustomDfsTransformer {
```
Geheiner, line 30: Can you add a comment what this code is testing? I'm not sure I understand. Is it an end 2 end test for your feature? If yes, please indicate it. It can be confusing that you test your new feature in the tests of an unrelated feature.

In generel in my opinion it would be nice to have a complete example of a Test using your new feature so that it can serve as an implementation example for developpers. With complete example I mean one package where you clearly see the class used, the config file and sdlb being called end 2 end to show you the complete picture looks like.

Geheiner (approved):
Implementation Looks really good, I just have minor suggestions, especially a end 2 end example with a complete test with config

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDfsTransformer.scala
zzeekk: You're right, i'll change it to a require.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDfsTransformer.scala
zzeekk: Well, the DataFrame (Dataset[Row]) case is handled above. 
With the typeOf[Map[String, Dataset[_]]] we then check for return values of a Dataset[XYZ]. Like that the transform function can return a Dataset[XYZ] instead of DataFrame, and CustomDfsTransformer converts it to a DataFrame for further use by SDLB.
I agree that it would be easy to use .toDF in the transform function, but using a Dataset[XYZ] as return type is a more precise definition of the return value. This can be interesting to show in the UI documentation of the transformer.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/customlogic/CustomDfsTransformer.scala
zzeekk: changed all asserts -> require

sdl-lang/src/test/scala/io/smartdatalake/meta/configexporter/DynamicTestTransformer.scala
zzeekk: This is just a CustomDfsTransformer implementation to check that parameters are exported into the documentation with ConfigJsonExorter. It is used in ConfigJsonExporterTest.
I'll add some scala doc.
I will also add an end2end config test, thats a good idea.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/ScalaClassSparkDfsTransformer.scala
zzeekk: done

sdl-core/src/test/scala/io/smartdatalake/workflow/action/spark/transformer/CustomDfsTransformerTest.scala
zzeekk: done :-)

PR #789, opened 2024-02-16
784 implement flattendftransformer

### LINKED ISSUE #784: Implement FlattenDfTransformer
I want to be able to transform a DataFrame with a nested schema into one with a flat schema. Child columns should inherit the names of the parents in order to be uniquely identifiable. I.e. the following schema...

root
 |-- name: string (nullable = true)
 |-- age: string (nullable = true)
 |-- jsonData: struct (nullable = true)
 |    |-- jobs: array (nullable = true)
 |    |    |-- element: struct (containsNull = true)
 |    |    |    |-- years: string (nullable = true)
 |    |    |    |-- job_name: string (nullable = true)

should be converted to...

root
 |-- name: string (nullable = true)
 |-- age: string (nullable = true)
 |-- jsonData_jobs_years: string (nullable = true)
 |-- jsonData_jobs_job_name: string (nullable = true)

iliassoto:
Implementation of a Transformer that flattens Spark Dataframes with a nested schema. Child columns inherit the names of the parents in order to be uniquely identifiable.  For example, the following schema...

root
|-- name: string (nullable = true)
|-- age: string (nullable = true)
|-- jsonData: struct (nullable = true)
| |-- jobs: array (nullable = true)
| | |-- element: struct (containsNull = true)
| | | |-- years: string (nullable = true)
| | | |-- job_name: string (nullable = true)

is converted to...

root
|-- name: string (nullable = true)
|-- age: string (nullable = true)
|-- jsonData_jobs_years: string (nullable = true)
|-- jsonData_jobs_job_name: string (nullable = true)

The Transformer can also explode cells that are defined as an array. This can be enabled in the configuration using the option "enableExplode = true".

The Transformer is only implemented for Spark as the necessary functionality is not yet available for Snowflake.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
@@ -64,17 +64,17 @@ case class SparkFlattenDfTransformer(override val name: String = "sparkFlattenDa
   @tailrec
   private def flattenDf(df: DataFrame): DataFrame = {
     val complexFields = getComplexFields(df)
-    if (complexFields.isEmpty) return df;
-    val (colName, colType) = complexFields.head
-    val newDf = colType match {
-      case StructType(fields) => {
-        val inner = for (n <- fields) yield n.name
-        val expanded = for (k <- inner) yield col(colName + '.' + k).alias(colName + '_' + k)
-        df.select(col("*") +: expanded: _*).drop(colName)
-      }
-      case a: ArrayType => df.withColumn(colName, explode_outer(col(colName)))
+    complexFields.foldLeft(df){
+      case (df, (colName, colType)) =>
+        val newDf = colType match {
+          case StructType(fields) =>
+            fields.map(_.name).foldLeft(df) {
+              case (df, name) => df.withColumn(colName + '_' + name, col(colName + '.' + name))
+            }.drop(colName)
+          case a: ArrayType => df.withColumn(colName, explode_outer(col(colName)))
+        }
+        flattenDf(newDf)
     }
-    flattenDf(newDf)
   }
   override def transform(actionId: ActionId, partitionValues: Seq[PartitionValues], df: DataFrame, dataObjectId: DataObjectId)(implicit context: ActionPipelineContext): DataFrame =
     flattenDf(df)
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
@@ -61,7 +61,6 @@ case class SparkFlattenDfTransformer(override val name: String = "sparkFlattenDa
     }.toSeq
   }
 
-  @tailrec
   private def flattenDf(df: DataFrame): DataFrame = {
     val complexFields = getComplexFields(df)
     complexFields.foldLeft(df){
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
@@ -1,0 +1,81 @@
+/*
+ * Smart Data Lake - Build your data lake the smart way.
+ *
+ * Copyright © 2019-2024 ELCA Informatique SA (<https://www.elca.ch>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.workflow.action.spark.transformer
+
+import com.typesafe.config.Config
+import io.smartdatalake.config.SdlConfigObject.{ActionId, DataObjectId}
+import io.smartdatalake.config.{FromConfigFactory, InstanceRegistry}
+import io.smartdatalake.util.hdfs.PartitionValues
+import io.smartdatalake.workflow.ActionPipelineContext
+import io.smartdatalake.workflow.action.generic.transformer.{GenericDfTransformer, SparkDfTransformer}
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.functions.{col, explode_outer}
+import org.apache.spark.sql.types.{ArrayType, DataType, StructType}
+import scala.annotation.tailrec
+
+/**
+ * Returns a flattened Dataframe from another DataFrame
+ * with a nested schema. Schemas in the form:
+ *
+ * |---parent
+ *
+ * |------|child1
+ *
+ * |------|child2
+ *
+ * are transformed to the following flat columns:
+ *
+ * |---parent_child1
+ *
+ * |---parent_child2
+ * @param name         name of the transformer
+ * @param description  Optional description of the transformer
+ */
+case class SparkFlattenDfTransformer(override val name: String = "standardizeSparkDatatypes", override val description: Option[String] = None) extends SparkDfTransformer {
+
+  private def getComplexFields(df: DataFrame): List[(String, DataType)] = {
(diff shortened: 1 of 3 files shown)


sdl-core/src/test/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDFTransformerTest.scala
```diff
@@ -0,0 +1,102 @@
...
+      .add("name", StringType, true)
+      .add("hobbies", ArrayType(StringType), true)
+
+    val arrayData = new ArrayList[Row]()
+    arrayData.add(Row("Michael", Seq("football", "programming")))
+    arrayData.add(Row("Bob", Seq("playing piano", "reading")))
+
+    val array_df = session.createDataFrame(arrayData, arraySchema)
+    val exploded_df = flattenDfTransformer.transform(ActionId("ActionId"), Seq(), array_df, DataObjectId("dataObjectId"))
+    val non_exploded_df = flattenDfTransformerNoExplode.transform(ActionId("ActionId2"), Seq(), array_df, DataObjectId("dataObjectId2"))
```
ThierryGirod, line 113: You could eventually write the negative test to another test for better separation, so you have one test per case.

sdl-core/src/test/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDFTransformerTest.scala
```diff
@@ -0,0 +1,102 @@
...
+    unNestedData.add(Row("Bob", 25))
+    val unNested_df = session.createDataFrame(unNestedData, normalSchema)
+    val new_df = flattenDfTransformer.transform(ActionId("ActionId"), Seq(), unNested_df, DataObjectId("dataObjectId"))
+    assert(new_df.schema.equals(normalSchema))
+  }
+
+
+  test("Array explode works") {
+    val flattenDfTransformer = SparkFlattenDfTransformer()
+    val flattenDfTransformerNoExplode = SparkFlattenDfTransformer(enableExplode = false)
```
ThierryGirod: Eventually extract the negative test case to a new test.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
```diff
@@ -0,0 +1,95 @@
...
+ * |---parent_child1
+ *
+ * |---parent_child2
+ * @param name         name of the transformer
+ * @param description  Optional description of the transformer
+ * @param enableExplode Enables exploding Arrays in the Dataframe, therefore potentially increasing the number of rows. Default is set to "true"
+ */
+case class SparkFlattenDfTransformer(override val name: String = "sparkFlattenDataFrame",
+                                     override val description: Option[String] = None,
+                                     enableExplode: Boolean = true) extends SparkDfTransformer {
```
ThierryGirod: Eventually disable explode by default. Because from my perspective when I would use the FlattenDfTransfomer as a user I would not expect from the name of the transfomer that he also explodes my data. Instead giving the possiblity to enable it, would need a conscious decision by the user.

ThierryGirod (commented):
Looks good overall, just see my comments.

ThierryGirod:
@iliassoto  You can also fix the 2 issues from sonarcloud regarding the naming of the variables. Variables should be camel-cased not snake-cased.

ThierryGirod (approved):
LGTM!

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
```diff
@@ -0,0 +1,95 @@
...
+              if (field.dataType.isInstanceOf[StructType] || field.dataType.isInstanceOf[ArrayType]))
+        yield (field.name, field.dataType)).toList
+      }
+      case false => {
+        (for (field <- df.schema.fields
+              if (field.dataType.isInstanceOf[StructType]))
+        yield (field.name, field.dataType)).toList
+      }
+    }
+  }
```
zzeekk: I have some scala remarks about this piece of code:
- matching a boolean (enableExplode) to true/false is a little bit verbose, an if / else statement would be more appropriate
- Checking the type of an instance can be done by isInstanceOf, but here pattern matching is more elegant.
- At the end you want to map and filter a list based on types. The second you can easily do with collect (collects matching records).

So i would suggest something like this:
```suggestion
  private def getComplexFields(df: DataFrame): Seq[(String, DataType)] = {
    df.schema.fields.map(f => (f.name, f.dataType)).collect {
      case (name, dataType: StructType) => (name,dataType)
      case (name, dataType: ArrayType) if enableExplode => (name,dataType)
    }.toSeq
  }
```

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
```diff
@@ -0,0 +1,95 @@
...
+    val newDf = colType match {
+      case StructType(fields) => {
+        val inner = for (n <- fields) yield n.name
+        val expanded = for (k <- inner) yield col(colName + '.' + k).alias(colName + '_' + k)
+        df.select(col("*") +: expanded: _*).drop(colName)
+      }
+      case a: ArrayType => df.withColumn(colName, explode_outer(col(colName)))
+    }
+    flattenDf(newDf)
+  }
```
zzeekk: And some Scala/Spark remarks here
- this gets all complex fields, but processes only one field per call of flattenDf. Performancewise it would be better to handle all fields of a level at the same time.
- it creates many select statements, which is not optimal for Sparks logical plan. It would be better to add new columns using withColumn together with foldLeft(df){...}

Here a suggestion:
```suggestion
  private def flattenDf(df: DataFrame): DataFrame = {
    val complexFields = getComplexFields(df)
    complexFields.foldLeft(df){
      case (df, (colName, colType)) =>
        val newDf = colType match {
          case StructType(fields) =>
            fields.map(_.name).foldLeft(df) {
              case (df, name) => df.withColumn(colName + '_' + name, col(colName + '.' + name))
            }.drop(colName)
          case a: ArrayType => df.withColumn(colName, explode_outer(col(colName)))
        }
        flattenDf(newDf)
    }
  }
```

sdl-core/src/test/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDFTransformerTest.scala
```diff
@@ -0,0 +1,115 @@
...
+    nestedData.add(Row("Bob", Row(25, "456 Elm St")))
+
+    val nested_df = session.createDataFrame(nestedData, nestedSchema)
+    val new_df = flattenDfTransformer.transform(ActionId("ActionId"), Seq(), nested_df, DataObjectId("dataObjectId"))
+    val expectedSchema = new StructType()
+      .add("name", StringType, true)
+      .add("info_age", IntegerType, true)
+      .add("info_address", StringType, true)
+    assert(new_df.schema.equals(expectedSchema))
+  }
```
zzeekk: I suggest testing 2 hierarchy levels:
```suggestion
  test("Nested DataFrame returns a flat Dataframe") {
    val flattenDfTransformer = SparkFlattenDfTransformer()
    val nestedSchema: StructType = new StructType()
      .add("name", StringType, true)
      .add("info", new StructType()
        .add("age", IntegerType, true)
        .add("address", new StructType()
          .add("street", StringType)
          .add("number", StringType)
        ))

    //createDataFrame requires Java Lists
    val nestedData = new ArrayList[Row]()
    nestedData.add(Row("Michael", Row(30, Row("Main St","123"))))
    nestedData.add(Row("Bob", Row(25, Row("Elm St","456"))))

    val nested_df = session.createDataFrame(nestedData, nestedSchema)
    val new_df = flattenDfTransformer.transform(ActionId("ActionId"), Seq(), nested_df, DataObjectId("dataObjectId"))
    val expectedSchema = new StructType()
      .add("name", StringType, true)
      .add("info_age", IntegerType, true)
      .add("info_address_street", StringType)
      .add("info_address_number", StringType)
    assert(new_df.schema.equals(expectedSchema))
  }
```

zzeekk (approved):
Nice new feature!
I have some suggestions for Scala/Spark improvements.
Looks good otherwise.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
iliassoto: Thanks @zzeekk I like the idea, but this would remove the tail-recursive structure and calls the function flattenDf for each column. Do you think the tradeoff is still worth it?

sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/SparkFlattenDfTransformer.scala
iliassoto: Yes way better :) !

PR #802, opened 2024-03-06
Fix using catalog for check

### LINKED ISSUE #801: DeltaLakeTableDataObject: dbExisting doesn't use configured catalog
The _isDbExisting_ method of the DeltaLakeTableDataObject checks that the DB configured exists in the catalog, but it only uses the DB name and not the provided catalog. This can result in an error if the configured catalog is not the same as the default catalog of the target metastore.

**To Reproduce**
Configure a DeltaLakeTableDataObject with a different catalog as the default catalog in the target metastore. The DB in the configured DataObject should exist in the configured catalog, but not in the default catalog of the metastore. 

**Expected behavior**
The method should check in the configured catalog and not in the default one.

iliassoto:
See #801


zzeekk (approved):
Very good.

PR #819, opened 2024-04-14
Try checkPermission method with two signatures

### LINKED ISSUE #818: Fix Databricks method checkPermission
**Describe the bug**
Some Databricks runtimes have replaced the signature of the internal method com.databricks.backend.daemon.dbutils.FSUtils.checkPermission. The SDLB tries to set the method as accessible using the "old" signature, which throws a NoSuchMethodException.

**To Reproduce**
Run SDLB in clusters hosted in Switzerland North, DRT 13.3. 

**Additional context**
Old / existing signature is: checkPermission(pathAndActions: Seq[PathAndActions], isUCEnabled: Boolean)(thunk: => T)
New signature is: checkPermission(pathAndActions: Seq[PathAndActions], isUCEnabled: Boolean, credentialId: Option[java.util.UUID)(thunk: => T)

A bugfix has been implemented in 2.6.1-SNAPSHOT. The same changes must be written in 2.7.0

iliassoto:
See https://github.com/smart-data-lake/smart-data-lake/issues/818.

Also, removed *checkPermissionAccess* as it isn't used.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/hdfs/UCFileSystemFactory.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/hdfs/UCFileSystemFactory.scala
@@ -72,10 +72,13 @@ object UCFileSystemFactory {
  */
 private[smartdatalake] class DbUtilsInterface(fsUtilsInst: Any, credentialScopeHelperClass: Class[_]) extends SmartDataLakeLogger {
 
-  def checkPermissionAccess[X](pathAndActions: Seq[_], withUnityCatalog: java.lang.Boolean, code : => X): X = {
-    checkPermissionMethod.invoke(fsUtilsInst, pathAndActions, withUnityCatalog, () => code).asInstanceOf[X]
-  }
-  private lazy val checkPermissionMethod = getMethod(fsUtilsInst.getClass, "checkPermission", Seq(classOf[Seq[_]], classOf[Boolean], classOf[Function0[_]]))
+  //As of 14.04.2024, Databricks provides the checkPermission method with two different signatures depenending on where the clusters are hosted.
+  private lazy val checkPermissionMethod = try {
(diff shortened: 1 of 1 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/hdfs/UCFileSystemFactory.scala
```diff
@@ -72,10 +72,13 @@ object UCFileSystemFactory {
...
-  }
-  private lazy val checkPermissionMethod = getMethod(fsUtilsInst.getClass, "checkPermission", Seq(classOf[Seq[_]], classOf[Boolean], classOf[Function0[_]]))
+  //As of 14.04.2024, Databricks provides the checkPermission method with two different signatures depenending on where the clusters are hosted.
+  private lazy val checkPermissionMethod = try {
+      getMethod(fsUtilsInst.getClass, "checkPermission", Seq(classOf[Seq[_]], classOf[Boolean], classOf[Function0[_]]))
+    }
+    catch {
+      case nsm: NoSuchMethodException => getMethod(fsUtilsInst.getClass, "checkPermission", Seq(classOf[Seq[_]], classOf[Boolean], classOf[Option[_]], classOf[Function0[_]]));
+    }
   checkPermissionMethod.setAccessible(true)
```
zzeekk: I think we could remove this completely, as it is not used currently. I added checkPermissionAccess here intially, as it was also in the stacktrace of accessing hadoop paths, but was not used in the end. registerPathAccess is sufficient, and removing checkPermissionAccess would make it more stable ;-)

iliassoto: Method entirely removed

zzeekk (approved):

PR #824, opened 2024-04-19
791 pre post sql for delta tables

### LINKED ISSUE #791: Pre-/Post SQL for delta tables
For JdbcTableDataObjects, we support preRead/postRead and preWrite/postWrite SQL Statements which are very handy. 
The same could be implemented for DeltaLakeTableDataObject.

Implement the same 4 pre/post read/write SQL Statements on DeltaLakeTableDataObject.

iliassoto:
See #791 .

The methods that were part of JdbcTableDataObject were passed to trait TransactionTableDataObject.
Logic implemented for IcebergTableDataObject, DeltaLakeTableDataObject, SnowflakeTableDataObject and TickTockHiveTableDataObject.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
@@ -34,14 +34,13 @@ object SparkQueryUtil extends SmartDataLakeLogger {
    */
   def executeSqlStatementBasedOnTable(session: SparkSession, stmt: String, table: Table): Unit = {
     try {
-      val newStmt: String = (table.catalog, table.db) match {
-        case (None, None) => stmt;
-        case (Some(cat), None) => f"USE CATALOG $cat;$stmt";
-        case (None, Some(db)) => f"USE SCHEMA $db;$stmt";
-        case (Some(cat), Some(db)) => f"USE CATALOG $cat;USE SCHEMA $db;$stmt";
-      }
-      logger.info(s"Executing SQL statement: $newStmt")
-      newStmt.split(";").foreach(session.sql(_))
+      val newStmt = Seq(
+        table.catalog.map(cat => s"USE CATALOG $cat"),
+        table.db.map(db => s"USE SCHEMA $db"),
+        Some(stmt)
+      ).flatten
+      logger.info(s"Executing SQL statements: ${newStmt.mkString(";")}")
+      newStmt.foreach(session.sql)
     } catch {
       case e: Exception =>
         logger.warn(s"Error in SQL statement '$stmt':\n${e.getMessage}")
--- a/sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
@@ -1,0 +1,43 @@
+/*
+ * Smart Data Lake - Build your data lake the smart way.
+ *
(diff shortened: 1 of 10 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
```diff
@@ -0,0 +1,51 @@
...
+   * the catalogs and schemas of the given DataObject are set as default.
+   * @param session Spark Session
+   * @param stmt Desired SQL statement to be executed.
+   * @param table DataObject in which the SQLStatement is configured
+   */
+  def executeSqlStatementBasedOnTable(session: SparkSession, stmt: String, table: Table): Unit = {
+    try {
+      val newStmt: String = (table.catalog, table.db) match {
+        case (None, None) => stmt;
+        case (Some(cat), None) => f"USE CATALOG $cat; $stmt";
```
pgruetter: Unfortunately, running two SQL commands in one go doesn't work in Spark. See i.e. [here](http://mark.thegrovers.ca/tech-blog/spark-sql-gotchas). 
So you need to execute two statements on the session, the first (optional) one for USE, and the second one executing stmt.

pgruetter (changes requested):

sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
iliassoto: Thank you @pgruetter! I just updated this

pgruetter: Great! Now it works correctly. Tested with Delta Tables. :+1:

pgruetter (approved):

sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
```diff
@@ -0,0 +1,51 @@
...
+  def executeSqlStatementBasedOnTable(session: SparkSession, stmt: String, table: Table): Unit = {
+    try {
+      val newStmt: String = (table.catalog, table.db) match {
+        case (None, None) => stmt;
+        case (Some(cat), None) => f"USE CATALOG $cat;$stmt";
+        case (None, Some(db)) => f"USE SCHEMA $db;$stmt";
+        case (Some(cat), Some(db)) => f"USE CATALOG $cat;USE SCHEMA $db;$stmt";
+      }
+      logger.info(s"Executing SQL statement: $newStmt")
+      newStmt.split(";").foreach(session.sql(_))
```
zzeekk: Instead of creating and splitting statements in a string, we could use a Seq directly:
```suggestion
      val newStmt = Seq(
        table.catalog.map(cat => s"USE CATALOG $cat"), 
        table.db.map(db => s"USE SCHEMA $db"),
        Some(stmt)
      ).flatten
      logger.info(s"Executing SQL statements: ${newStmt.mkString(";")}")
      newStmt.foreach(session.sql)
```

zzeekk (approved):
Looks good, just a minor suggestion to simplify some code.

sdl-core/src/main/scala/io/smartdatalake/util/spark/SparkQueryUtil.scala
iliassoto: Thank you @zzeekk for the suggestion. I included your changes :)

PR #828, opened 2024-04-22
539 activate merge mode on existing tables

### LINKED ISSUE #539: Activate Merge Mode on existing tables
When using merge mode on a historization action, a column dl_hash is added to your DataObject which is used to detect changes. 
This works fine when merge mode is activated for a new DataObject (that has never been written before). 
When activating it for a DataObject with existing data, schema evolution takes care to add the column dl_hash, but it's empty for all existing records. 

First, check if the current solution works. It's likely, that all previous records (with empty dl_hash) get closed and a new record with dl_hash is created. If that's the case, we can discuss to leave it like that.
The cleaner solution would be to calculate dl_hash for all previous records on activation.

ThierryGirod:
 
Fix when using merge mode on a historization action the added column dl_hash was NULL.

View issue #539

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/evolution/SchemaEvolution.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/evolution/SchemaEvolution.scala
@@ -57,17 +57,20 @@ object SchemaEvolution extends SmartDataLakeLogger {
     * @param cols
     * @return
     */
-  def sortColumns(df: DataFrame, cols: Seq[String]): DataFrame = {
-    val dfCols = df.columns
-    val colsToSelect = cols.filter( c => dfCols.contains(c)).map(col)
+  def sortColumns(df: DataFrame, cols: Seq[String], caseSensitive: Boolean = false): DataFrame = {
+    val dfCols = if (caseSensitive) df.columns else df.columns.map(_.toLowerCase)
+    val colsToSelect = if (caseSensitive) cols.filter( c => dfCols.contains(c)).map(col)  else cols.filter(c => dfCols.contains(c.toLowerCase)).map(col)
     df.select(colsToSelect: _*)
   }
 
-  def getFieldTuples(dataFrame: DataFrame): Set[(String, String)] = {
-    getFieldTuples(dataFrame.schema)
+  def getFieldTuples(dataFrame: DataFrame, caseSensitive: Boolean = false): Set[(String, String)] = {
+    getFieldTuples(dataFrame.schema, caseSensitive)
   }
-  def getFieldTuples(schema: StructType): Set[(String, String)] = {
-    schema.fields.map(f => (f.name, f.dataType.simpleString)).toSet
+  def getFieldTuples(schema: StructType, caseSensitive: Boolean): Set[(String, String)] = {
+    if(caseSensitive)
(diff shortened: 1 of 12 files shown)


zzeekk (approved):
LGTM.
Should we do this for Iceberg as well?
Is the PR still "draft" by intention?

ThierryGirod:
@zzeekk yes its intentional, because I am currently working on it for iceberg and jdbc.

sdl-core/src/main/scala/io/smartdatalake/util/evolution/SchemaEvolution.scala
```diff
@@ -173,17 +186,19 @@ object SchemaEvolution extends SmartDataLakeLogger {
       // this defines the ordering of the resulting DataFrame's
       val tgtCols = if (Environment.schemaEvolutionNewColumnsLast) {
         // new columns last
-        oldColsWithoutTechCols ++ newColumns(oldDf, newDf) ++ colsToIgnore
+        if (caseSensitiveComparison) oldColsWithoutTechCols ++ newColumns(oldDf, newDf) ++ colsToIgnore
+        else oldColsWithoutTechCols ++ newColumns(oldDf, newDf) ++ colsToIgnore.map(_.toLowerCase)
```
zzeekk:
```suggestion
        if (caseSensitiveComparison) oldColsWithoutTechCols ++ newColumns(oldDf, newDf) ++ (if (caseSensitiveComparison) colsToIgnore else colsToIgnore.map(_.toLowerCase))
```

sdl-core/src/main/scala/io/smartdatalake/util/evolution/SchemaEvolution.scala
```diff
@@ -173,17 +186,19 @@ object SchemaEvolution extends SmartDataLakeLogger {
...
       } else {
         // deleted columns last
-        newColsWithoutTechCols ++ deletedColumns(oldDf, newDf) ++ colsToIgnore
+        if (caseSensitiveComparison) newColsWithoutTechCols ++ deletedColumns(oldDf, newDf) ++ colsToIgnore
+        else newColsWithoutTechCols ++ deletedColumns(oldDf, newDf) ++ colsToIgnore.map(_.toLowerCase)
```
zzeekk:
```suggestion
        if (caseSensitiveComparison) newColsWithoutTechCols ++ deletedColumns(oldDf, newDf) ++ (if (caseSensitiveComparison) colsToIgnore else colsToIgnore.map(_.toLowerCase))
```

zzeekk (approved):
Well done!
Please use "squash and merge" to merge the PR...

sdl-core/src/main/scala/io/smartdatalake/util/evolution/SchemaEvolution.scala
ThierryGirod: Fixed in Commit cb32221f47394f53c4b06c5d6b2c277df14e2d91

sdl-core/src/main/scala/io/smartdatalake/util/evolution/SchemaEvolution.scala
ThierryGirod: Fixed in commit cb32221f47394f53c4b06c5d6b2c277df14e2d91

Issue #813, opened 2024-03-30
Better support for SDLB in Notebooks


zzeekk:
**Is your feature request related to a problem? Please describe.**
SDLB creates stable and maintainable production data pipelines.
But development in Notebook is not supported very well yet.
Its possible to use DataObject definitions in Notebooks, see LabCatalogGenerator, but you cannot yet work with or extend transformations.

**Describe the solution you'd like**
Extend LabCatalogGenerator with Actions and transformations.

PR #829, opened 2024-04-24
SDLB Interface for working in Notebooks / Labs

zzeekk:
see #813

Documentation and example in Databricks see #830


pgruetter (approved):
Just skimmed through this one, but looks good to me.

PR #830, opened 2024-04-29
Notebook catalog documentation

zzeekk:
Documentation for #813 

Should we add this rather as blog post, WDYT?


docs/reference/notebookCatalog.md
```diff
@@ -0,0 +1,166 @@
+---
+id: notebookCatalog
+title: Use in Notebooks
+---
+
+SDLB can also be used in Notebooks. On one side SDLB can provide a high-quality catalog of data objects, and on the other hand it helps to easily prototype transformations in an interactive Notebook environment, and then transfer them to repository for production use.
```
pgruetter:
```suggestion
SDLB can also be used in Notebooks. On one side SDLB can provide a high-quality catalog of data objects, and on the other side it helps to easily prototype transformations in an interactive Notebook environment, and then transfer them to repository for production use.
```

pgruetter:
@zzeekk To be honest, I would separate the _Usage_ of the feature (in the Reference doc) and the _technical details_  (as a blog post). 
Example:
Can we provide the Maven profile by default? Then we could simply explain in the doc, that the profile needs to be activated, without the code block.
Or explanations like "The functions of a DataObject have been adapted for interactive use in notebooks via a corresponding wrapper." could also be omitted in the doc.
Let's discuss offline maybe, have a few additional questions / comments.

docs/reference/notebookCatalog.md
```diff
@@ -0,0 +1,166 @@
...
+        </plugins>
+    </build>
+</profile>
+```
+
+The LabCatalogGenerator creates source code for class `ActionCatalog` and `DataObjectCatalog` in package `io.smartdatalake.generated`. These names can be customized by command line parameters. The directory for the generated source code has to be provided as `srcDirectory` parameter.
+
+Once generated, the fat-jar (created with maven profile `fat-jar`) and SDLB configuration files have to be uploaded into a Databricks Volume. Then create a Databricks Cluster with the fat-jar as Library and Environment Variable `JNAME=zulu17-ca-amd64` to use Java 17.
+
+Make your SDLB data pipeline did run at least once, so that tables are existing. A simple Databricks Job configuration can look as follows:
```
pgruetter:
```suggestion
Make sure your SDLB data pipeline runs at least once to ensure that tables exist. A simple Databricks Job configuration can look as follows:
```

pgruetter (approved):

Issue #794, opened 2024-02-26
Question: how to configure Hikari settings?


AdiWehrli:
Hello together,

Does one know, how we can set Hikari properties for JDBC connection like `connectionTimeout`, `idleTimeout`, `connectionTestQuery` etc.?

Thanks very much in advance.

Kind regards,
Adi

pgruetter:
Hi @AdiWehrli 
Don't know Hikari. But did you try setting the jdbcOptions on the data object?
See https://smartdatalake.ch/json-schema-viewer?schema=sdl-schema-2.6.0-SNAPSHOT.json&path=%5B2%2C12%2C14%5D
Should be possible to set any JDBC options with this parameter.

pgruetter:
.. which gets me thinking: it should be possible to set jdbcOptions on JDBC connections. Any options set on the connection, should be applied each time the connection is used. 
@zzeekk What do you think?

ThierryGirod:
@pgruetter I think we have to pay attention. The point that comes to my mind is that you cannot generalize all jdbc options to the connection. You also have options that are data object related ex. if you specify the query or the dbtable parameter.

pgruetter:
True, there's also considerations like which options you can overwrite in your DataObject that are also defined on the connection level. But I mean in general, would be maybe worth analyzing.

AdiWehrli:
Thanks for the replies. To provide more details: we want to ensure, that a `JdbcTableConnection` can switch between multiple configured DB hosts, if one would suddenly be unavailable. That's the main point here. And we mostly use Hikari in our other applications to ensure this.

As I looked at the [SDL schema](https://smartdatalake.ch/json-schema-viewer/?schema=sdl-schema-2.5.2.json) I recognized, that such properties would not be valid. We have the following valid properties for pooling/initalization in `JdbcTableConnection`:
`maxParallelConnections`
`connectionPoolMaxIdleTimeSec`
`connectionPoolMaxWaitTimeSec`
`connectionInitSql`

But there are no properties to ensure valid connections, as you can see in [HikariCP - Configuration (knobs, baby!)](https://github.com/brettwooldridge/HikariCP#gear-configuration-knobs-baby), are there? 

Maybe the Hikari property `connectionTestQuery` could mean almost the same as SDL property `connectionInitSql`? But I do not suppose so, because Hikari also has a separate `connectionInitSql`:

[...]

PR #831, opened 2024-05-08
Update JdbcTableConnection.scala: added properties to pool.

AdiWehrli:
As the `testOnBurrow` property of `BaseGenericObjectPool` is set to `false` as default, we should be able to set this. Additionally I added some more "testOn/While" properties.

Also see issue https://github.com/smart-data-lake/smart-data-lake/issues/794.


AdiWehrli:
I tested it locally with some log output to check whether the flags were correctly set.

AdiWehrli:
So, now this pull request is ready.

zzeekk:
Great finding! GenericObjectPool supports tests/validations indeed.
We would also need to implement the validation method in JdbcClientPoolFactory, inherited default implementation just returns true:
`override def validateObject(p: PooledObject[SqlConnection]): Boolean`

And question is what query to execute as test. I think we need another configuration property for this query, as it is probably not the same for every database product...

Regarding the new configuration properties - i would suggest to move them into a substructure by creating a `case class ConnectionPoolConfig(maxIdleTimeSec: Int = 3, maxWaitTimeSec: Int = 600, testOnBorrow: Boolean = false, testOnCreate: Boolean = false, testOnReturn: Boolean = false, testWhileIdle: Boolean = false, testQuery: String = <what is a  good default?>)` and a property in JdbcTableConnection `connectionPool: ConnectionPool = ConnectionPool()`
WDYT?

Unfortunately we just merge some changes into develop-spark3 and you need to pull some updates and resolve minor conflicts.

AdiWehrli:
Hello @zzeekk 

Thanks for your answer. That's a good idea to move those properties to a seperate config class. A good default for `testQuery` would be `SELECT 1`, I suppose?

I will pull the changes and resolve the conflicts maybe on Sunday.

Kind regards, 
Adi

AdiWehrli:
Aha, I saw, that the connection properties now are set in `JdbcUtil`. So, I will create a new pull-request, maybe tomorrow.

Issue #608, opened 2022-12-17
Handle writing no data to DataObject more intelligently


zzeekk:
**Is your feature request related to a problem? Please describe.**
Currently when writing to a DataObject it might collect metrics saying that there was no data to process. This is logged, but not used to inform subsequent actions. 

**Describe the solution you'd like**
Inform subsequent actions that input is empty by setting a corresponding attribute on the SubFeed, maybe isSkipped, or something similar.

iliassoto:
I think the solution should separate two cases: 
1. There is no data to process in the first place (the input data object is empty)--> Action should be skipped in my opinion. Currently, reading from an empty ParquetFile causes problems.
2. There is no data to be written in the output DataObject --> Action should be carried out but it attaches a flag isSkipped.

zzeekk:
Also Expectations dont get an answer from the Spark Observation if there is no data... the Error message is confusing.

zzeekk:
We can check if there is Data before writing to DataFrame to the Spark DataSource by adding a custom operation to the Spark plan. Like that we can check if there are any records when executing the Spark plan.

For 1:1 Actions the behaviour then is simple: if there is no data, a NoDataToProcessWarning is thrown, the action is marked as skipped. Nothing is written to the output DataObject.

For 1:n Actions (or also n:m...) it is more complicated. I suggest that a NoDataToProcessWarning is only thrown, if the mainOutput has no data, but not when writing to further outputs. And i also suggest that if the mainOutput has no data, further outputs are processed nevertheless, meaning a mainOutput without data does not stop processing further outputs of the Action. This means the action might be marked as skipped, but has written data to non-main Outputs.

PR #840, opened 2024-06-04
Feature/608: add SparkPlan rule to fail if there is no data

zzeekk:
see #608

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/spark/SDLSparkExtension.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/spark/SDLSparkExtension.scala
@@ -1,0 +1,64 @@
+/*
+ * Smart Data Lake - Build your data lake the smart way.
+ *
+ * Copyright © 2019-2024 ELCA Informatique SA (<https://www.elca.ch>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.util.spark
+
+import io.smartdatalake.definitions.Environment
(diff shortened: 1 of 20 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/spark/SDLSparkExtension.scala
```diff
@@ -0,0 +1,78 @@
...
+}
+
+case class AssertNotEmptyExec(child: SparkPlan) extends UnaryExecNode {
+
+  // Output has the same attributes as input
+  override def output: Seq[Attribute] = child.output
+
+  // Examine child's data
+  override protected def doExecute() = {
+    val rdd = child.execute()
```
Geheiner: is Spark smart enough to not completely execute the whole query plan a second time for this? It does not look like it

sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
```diff
@@ -352,7 +352,7 @@ abstract class DataFrameActionImpl extends ActionSubFeedsImpl[DataFrameSubFeed]
         subFeed.withMetrics(streamingMetrics.getMainInfos ++ sparkMetrics).asInstanceOf[DataFrameSubFeed]
       case _ =>
         // Auto persist if dataFrame is reused later
-        val preparedSubFeed = if (context.dataFrameReuseStatistics.contains((output.id, subFeed.partitionValues))) {
+        var preparedSubFeed = if (context.dataFrameReuseStatistics.contains((output.id, subFeed.partitionValues))) {
```
Geheiner: why does this need to be a var now? I don t see any reassignement

Geheiner (approved):

sdl-core/src/main/scala/io/smartdatalake/util/spark/SDLSparkExtension.scala
zzeekk: Thanks, you're right. This triggers a separate spark job (within the same Query), which is not optimal.
I now found different ways to evaluate the logical plan, without using rdd count or isEmpty.

sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
zzeekk: fixed

sdl-core/src/main/scala/io/smartdatalake/util/spark/SDLSparkExtension.scala
Geheiner: I don t get what you changed compared to before, let's have a look together

PR #843, opened 2024-06-11
Fix creating no sample file for SparkFileDataObject if schema is fixed

zzeekk:
SparkFileDataObjects dont need a sample file for schema inference if the schema is fixed.

Implemented logic is not correct / can be improved.


zzeekk:
replaced by #845

Issue #849, opened 2024-06-25
FileIncrementalMoveMode should allow to archive files relative to parent path, not dataObject-root-path


zzeekk:
**Is your feature request related to a problem? Please describe.**
FileIncrementalMoveMode.archiveDir moves files for partitioned DataObjects as follows:
`<dataObject-root>/<archiveDir>/<partitionDirs>/<filename>`
This is good because the partition does no longer exist afterwards.

But sometimes there is a need to archive files inside the partition directory:
`<dataObject-root>/<partitionDirs>/<archiveDir>/<filename>`

**Describe the solution you'd like**
Add a flag FileIncrementalMoveMode.archiveInsidePartition (boolean). If enabled, files are archived inside the partitions directory in the `<archiveDir>`.

PR #850, opened 2024-06-25
implement FileIncrementalMoveMode.archiveInsidePartition (#849)

zzeekk:
see #849


sdl-core/src/main/scala/io/smartdatalake/workflow/action/executionMode/FileIncrementalMoveMode.scala
```diff
@@ -41,12 +41,20 @@ import org.apache.spark.sql.Row
  * A better implementation would be to observe files by a custom metric. Unfortunately there is a problem in Spark with that, see also [[CollectSetDeterministic]]
  * - Partition values preserved.
  *
- * @param archivePath if an archive directory is configured, files are moved into that directory instead of deleted, preserving partition layout.
+ * @param archivePath if an archive directory is configured, files are moved into that directory instead of deleted, preserving partition layout. See also `àrchiveInsidePartition` option.
```
pgruetter:
```suggestion
 * @param archivePath if an archive directory is configured, files are moved into that directory instead of deleted, preserving partition layout. See also `archiveInsidePartition` option.
```

pgruetter (approved):

Issue #851, opened 2024-06-25
Add partition columns to user defined schema of SparkFileDataObjects when they are missing


zzeekk:
**Is your feature request related to a problem? Please describe.**
For now partition columns must be included in user defined schemas of SparkFileDataObjects.
This limits direct use of existing schema definitions (XSD, Json Schema) if the files are read from partitioned sources.
It would be nice if missing partition columns are automatically added to user defined schemas if they are missing.

**Describe the solution you'd like**
Add partition columns to user defined schema of SparkFileDataObjects when they are missing.

PR #852, opened 2024-06-25
Add missing partition columns to schema of SparkFileDataObjects (#851)

zzeekk:
see #851

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
@@ -253,6 +253,9 @@ trait GenericSchema extends GenericTypedObject {
   def sql: String
   def add(colName: String, dataType: GenericDataType): GenericSchema
   def add(field: GenericField): GenericSchema
+  def addIfNotExists(colName: String, dataType: GenericDataType): GenericSchema = {
+    if (!columnExists(colName)) add(colName, dataType) else this
+  }
   def remove(colName: String): GenericSchema
   def filter(func: GenericField => Boolean): GenericSchema
   def getEmptyDataFrame(dataObjectId: DataObjectId)(implicit context: ActionPipelineContext): GenericDataFrame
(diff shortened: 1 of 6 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
```diff
@@ -253,6 +253,9 @@ trait GenericSchema extends GenericTypedObject {
   def sql: String
   def add(colName: String, dataType: GenericDataType): GenericSchema
   def add(field: GenericField): GenericSchema
+  def addIfNotExists(colName: String, dataType: GenericDataType): GenericSchema = {
+    if (!columnExists(colName)) add(colName, dataType) else this
```
pgruetter: We could log a info message here if the column is automatically added. Otherwise it's a bit hidden and hard to trace.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/SparkFileDataObject.scala
```diff
@@ -109,12 +109,16 @@ trait SparkFileDataObject extends HadoopFileDataObject
...
+   * @return The consolidated schema to validate the schema on read and write, and for the data frame reader when reading from the source.
    */
   def getSchema(implicit context: ActionPipelineContext): Option[SparkSchema] = {
     _schemaHolder = _schemaHolder.orElse(
-        // get defined schema
-        schema.map(_.convert(typeOf[SparkSubFeed]).asInstanceOf[SparkSchema])
+        // get defined schema, add potentially missing partition columns as type string
+        schema.map { s =>
+          partitions.foldLeft(s.convert(typeOf[SparkSubFeed])) {
+              case (schema,col) => schema.addIfNotExists(col, SparkSimpleDataType(StringType))
```
pgruetter: Should we find a way to get the actual datatype instead of String, or is it not possible at this point?

zzeekk: Default for hadoop partition layout is string, other datatypes can cause strange problems.
-> It is best to leave it as string.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/GenericDataFrame.scala
zzeekk: yes, done.

pgruetter (approved):

Issue #853, opened 2024-07-05
Improve logging sdlb version, add app version


zzeekk:
**Is your feature request related to a problem? Please describe.**
SDLB githash should be written to log, not only pom version, date, creator.
Additionally there should be the possibility to log also the version of the application using SDLB

**Describe the solution you'd like**
Read git hash from repository during build and add it to sdlb-version-info.properties.

Create a cmd line tool to write an app-version-info.properties file during maven build of the application using SDLB. Additionally log these versions on startup if the file is available.

PR #855, opened 2024-07-06
implement logging sdlb build githash and app version info (#853)

zzeekk:
see #853

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
@@ -20,7 +20,7 @@
 package io.smartdatalake.workflow
 
 import io.smartdatalake.util.misc.SmartDataLakeLogger
-import org.json4s.JsonAST.{JInt, JNothing, JObject}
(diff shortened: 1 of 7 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
```diff
@@ -101,6 +101,35 @@ class StateMigratorDef3To4 extends StateMigratorDef with SmartDataLakeLogger {
...
     }.asInstanceOf[JObject]
 
+    // update version and return
+    updateVersion(migratedJson, versionTo)
+  }
+}
+
+/**
+ * Migrate state from format version 4 to 5:
+ * - buildVersionInfo is renamed to sdlbVersionInfo
```
pgruetter, line 111: Does that imply a change in the UI ? Then we should maybe directly create a Issue for it.

pgruetter (approved):

sdl-core/src/main/scala/io/smartdatalake/workflow/StateMigratorDef.scala
zzeekk: Yes, created https://github.com/smart-data-lake/sdl-visualization/issues/102

Issue #856, opened 2024-07-13
Comparison Column in DataFrameIncrementalMode has become caseSensitive by default


Geheiner:
**Describe the bug**
The Column that is used to compare the Input DataObject and the Output DataObject in the DataFrameIncrementalMode has become caseSensitive by default in recent versions. This is an unwanted change of behavior and can cause a the following problem:
If the comparison column in the InputDO is case sensitive, but if the comparison column is case insensitive in the outputDO, DataFrmaeIncrementalMode will treat them as 2 different columns and fail.

**To Reproduce**
Use on InputDataObject with a column named lastUpdate and an OutputDataObject with a column lastupdate and use DataframeIncrementalMode with comparisonColumn lastupdate.

**Expected behavior**
Task runs through.
Currently, it generates an error telling you that lastupdate cannot be found in the source.

**Additional context**

PR #858, opened 2024-07-14
856 fix bug, add test cases

Geheiner:
Fixes Bug https://github.com/smart-data-lake/smart-data-lake/issues/856

### DIFF
--- a/sdl-core/src/test/scala/io/smartdatalake/testutils/TestUtil.scala
+++ b/sdl-core/src/test/scala/io/smartdatalake/testutils/TestUtil.scala
@@ -120,7 +120,7 @@ object TestUtil extends SmartDataLakeLogger {
                      )(implicit instanceRegistry: InstanceRegistry, context: ActionPipelineContext): HiveTableDataObject = {
     val table = Table(db=db,name=tableName,primaryKey=primaryKeyColumns)
     val path = dirPath+s"$tableName"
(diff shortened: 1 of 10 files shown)


sdl-core/src/test/scala/io/smartdatalake/testutils/TestUtil.scala
```diff
@@ -120,7 +120,7 @@ object TestUtil extends SmartDataLakeLogger {
                      )(implicit instanceRegistry: InstanceRegistry, context: ActionPipelineContext): HiveTableDataObject = {
     val table = Table(db=db,name=tableName,primaryKey=primaryKeyColumns)
     val path = dirPath+s"$tableName"
-    val hTabDo = HiveTableDataObject(id=s"${tableName}DO",path=Some(path),schemaMin=schemaMin.map(SparkSchema),table=table)
+    val hTabDo = HiveTableDataObject(id=s"${tableName}DO",path=Some(path),schemaMin=schemaMin.map(SparkSchema.apply),table=table)
```
Geheiner, line 123: in case your wondering why I am adding SparkSchema.apply: it makes it much clearer that actually a function, the "constructor" is passed as an argument and not the class SparkSchema itself. During Refactoring I had created a companion object SparkSchema and encountered an error as it tried to pass the companion object here. Took me a while to understand that you actually want to pass the constructor function. So let's just do that, less confusing and if we ever create a companion object it will save somebody some time

pgruetter (approved):

PR #861, opened 2024-07-25
Update SmartDataLakeBuilder.scala

Geheiner:
Correct Doc of startSimulation method


sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -398,7 +398,7 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
   /**
    * Start a simulation run.
    * This executes the DAG and returns all SubFeeds including the transformed DataFrames.
-   * Only prepare and init phase are executed.
+   * Only the exec phase is executed.
```
zzeekk: I dont think this is correct. According to SmartDataLakeBuilder:515 only init Phase is executed. Can you please check?

Geheiner: you are right. is it better now?

zzeekk (approved):

Issue #873, opened 2024-08-02
metricsFailCondition is not executed if Action is skipped


zzeekk:
**Describe the bug**
metricsFailCondition of CopyAction or CustomDataFrameAction is not executed if Action is skipped.

**To Reproduce**
See description.

**Expected behavior**
metricsFailCondition should also be evaluated if Action is skipped (but not if it fails).

zzeekk:
postExecFailed is executed if there was an Exception, this includes NoDataToProcessWarning.

I'll add evaluating metricsFailCondition in postExecFailed if exception is NoDataToProcessWarning. Like that metricsFailCondition can override NoDataToProcessWarning with an Exception to fail the Action, instead of skipping it.

PR #877, opened 2024-08-06
Fix #872, #873

### LINKED ISSUE #872: SubFeed metrics keep input values if skipped
**Describe the bug**
Output SubFeeds keep metrics from Input SubFeeds when action is Skipped

**To Reproduce**
2 Actions, second is skipped because there is no-data.
When debugging it seems that SubFeed metrics are not cleared, but keep the value from the Input SubFeed, e.g. from the previous Actions output SubFeed.

When an Action is executed successfully or fails because of expectation validation failure, each output SubFeeds metrics are overwritten. But if the Action is skipped this seems not to be the case.

**Expected behavior**
Output SubFeeds metrics are cleared before execution is started.

zzeekk:
see #872, #873

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
@@ -77,7 +77,7 @@ abstract class DataFrameActionImpl extends ActionSubFeedsImpl[DataFrameSubFeed] 
   override def isAsynchronousProcessStarted: Boolean = isAsynchronous && sparkStreamingQuery.nonEmpty
 
   /**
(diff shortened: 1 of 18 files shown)


Geheiner (approved):
Looks good!

sdl-core/src/main/scala/io/smartdatalake/workflow/action/DataFrameActionImpl.scala
```diff
@@ -553,8 +553,4 @@ abstract class DataFrameActionImpl extends ActionSubFeedsImpl[DataFrameSubFeed]
         }
     }
   }
-
-  override def postExecFailed(implicit context: ActionPipelineContext): Unit = {
-    super.postExecFailed
```
Geheiner, line 558: thanks for removing this, now there is less risk that someone overrides it and then wonders why the metrics are not executed.

PR #884, opened 2024-08-15
Feature/502 implement odata dataobject

### LINKED ISSUE #502: Implement OData DataObject
I would like to read data from OData REST Webservices with SDLB, as it's a standard interface for D365 CE (CRM) & FO (ERP) and SAP.

Implement an ODataWebserviceEntityDataObject to read data from OData REST Webservices with Spark.
An OData endpoint presents a list of entities it provides. The data object should have an attribute to select the entity it will extract.
Potential Java Libraries: https://www.odata.org/libraries/
Also OData provides metadata about the data structures: https://www.odata.org/blog/queryable-odata-metadata/. It would be nice to convert this to a Spark Schema in Init-Phase, as we do it already with XSD Schema, Json Schema ...

**Additional context**
https://www.odata.org/
D365 OData: https://docs.microsoft.com/en-us/dynamics365/fin-ops-core/dev-itpro/data-entities/odata
Extracting data from SAP BW with OData ODP: https://nttdata-solutions.com/ch/local-blog/sap-bw-4hana-datenextraktion-mit-odata-odp/

PMuellerELCA:
New DataObject ODataDataObject


zzeekk:
Open Points:
- consolidate local / hdfs File Buffer implementation
- rename incremental Output column expression (changeDataColumn -> incrementalOutputExpr), see e.g. JdbcTableDataObject

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/ODataDataObject.scala
```diff
@@ -0,0 +1,644 @@
...
+import org.apache.spark.sql.custom.ExpressionEvaluator
+
+import scala.reflect.runtime.universe.typeOf
+
+/**
+ * InversionOfControl container class
+ * This class serves as a hub or a relay for all calls to class creators and to static functions outside of this module.
+ * It allows test code to override these calls and to inject Mock- and Spy objects into the running code.
+ */
+class ODataIOC {
```
zzeekk, line 37: As this is state less, it could be an "object" instead of "class".

zzeekk (approved):
Nice implementation, only one comment about a small improvement from my side.
Is the PR still a draft, or should it actionally be "ready for review"?
If we can merge soon, it might be included in release 2.7.1

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/ODataDataObject.scala
PMuellerELCA: Converting to an object may lead the problems during the UnitTests. The purpose of this class is to use the Mockito Spy feature to override methods and inject test objects like Mocks into the execution.

Issue #880, opened 2024-08-13
Typo in CompletenessExpectation


pgruetter:
**Describe the bug**
As mentioned in PR Review #876, there's a typo in Compl**e**tenessExpectation (missing "e").

**Additional context**
The typo is already in release 2.7.0.

PR #888, opened 2024-08-22
fix typo in CompletenessExpectation class name (#880)

zzeekk:
See #880

Note that this is a breaking change, but as CompletenessExpectation was introduced recently in Release 2.7.0, we can assume that its not yet largely in use.


pgruetter (approved):
:+1:

Issue #890, opened 2024-09-03
Snowflake listPartitions returns uppercase column names


zzeekk:
**Describe the bug**
SnowflakeTableDataObject.listPartitions returns PartitionValues with uppercase column names. In consequence partitions are not found, as comparision is not done case-sensitive.

**To Reproduce**
Create a SnowflakeTableDataObject with partitions=[test].
Write some data to the table.
Execute listPartitions on the DataObject. It will return PartitionValues("TEST" -> "...").

**Expected behavior**
If not Environment.caseSensitive, convert columns to lowerCase when instantiating PartitionValues.

PR #891, opened 2024-09-03
Snowflake listPartitions lower case

zzeekk:
See #890

The PR includes some other cleanups and improvements
- removal of dependency com.github.stefanbirkner:system-rules from sdl-azure as it's pom somehow refers to a Snapshot Version of junit-dep, which can create build problems on some environments. We can get rid of it as it is only used in one not so important Java unit test, which we can remove as well.
- upgrade of snowpark to version 3.0.0, which is no longer dependent on a Spark minor version :-)
- implementation of SnowflakeConnection.proxy authentication
- improvements of IcebergTableDataObject when using Hadoop Catalog


sdl-iceberg/src/main/scala/io/smartdatalake/workflow/dataobject/IcebergTableDataObject.scala
```diff
@@ -75,25 +78,15 @@ import scala.util.Try
  * - [[CanEvolveSchema]] by using internal Iceberg API.
  * - Overwriting partitions is implemented by using DataFrameWriterV2.overwrite(condition) API in one transaction.
  *
- * @param id unique name of this data object
  * @param path hadoop directory for this table. If it doesn't contain scheme and authority, the connections pathPrefix is applied.
  *             If pathPrefix is not defined or doesn't define scheme and authority, default schema and authority is applied.
- * @param partitions partition columns for this data object
+ *             If Iceberg table is defined on aa hadoop catalog, path must be None as it is defined through the catalog directory structure.
```
mand35: typo aa

mand35 (approved):
looks good

sdl-iceberg/src/main/scala/io/smartdatalake/workflow/dataobject/IcebergTableDataObject.scala
zzeekk: Ok, fixed

zzeekk:
Thanks @mand35.
I cleaned up git history and will merge when build succeeds.

Issue #900, opened 2024-10-06
HistorizeAction needs input schema in prepare phase


zzeekk:
**Describe the bug**
HistorizeAction with mergeModeEnable = true needs input schema in prepare phase, see the following stack trace:
	at io.smartdatalake.workflow.dataobject.CsvFileDataObject.getDataFrame(CsvFileDataObject.scala:68)
	at io.smartdatalake.workflow.action.HistorizeAction.initSaveModeOptions(HistorizeAction.scala:155)
	at io.smartdatalake.workflow.action.HistorizeAction.prepare(HistorizeAction.scala:189)

This violates the principal that schemas are only available in init phase and results in broken pipelines on fresh environments where inferred schemas are not yet available.

**To Reproduce**
Getting-started `Action~historize-airports`, when executing whole pipeline in fresh environment.
Exception:
2024-10-06 12:41:52 ERROR ActionDAGRun - Prepare: TaskFailedException: Task historize-airports failed. Root cause is 'IllegalArgumentException: requirement failed: (DataObject~stg-airports) DataObject schema is undefined. A schema must be defined if there are no existing files.' [main]

**Expected behavior**
HistorizeAction.initSaveModeOptions (HistorizeAction.scala:155) should not use a schema in prepare phase. 
Getting-started should run successfull in a fresh environment.

Issue #896, opened 2024-10-01
Count and TransferRate expectations broken


PMuellerELCA:
When adding the CountExpectation to a DeltaLakeTableDataObject or the new TransferRateExpectation to an action the program aborts with the following error message:

SDLB Version 2.7.1

```
TaskFailedException: Task slv-dim-dates failed. Root cause is 'IllegalStateException: (DataObject~slv-dim-dates) Metric 'countXXX' for expectation 'countXXX' not found for validation'
Caused by: IllegalStateException: (DataObject~slv-dim-dates) Metric 'countXXX' for expectation 'countXXX' not found for validation
Caused by: java.lang.IllegalStateException: (DataObject~slv-dim-dates) Metric 'countXXX' for expectation 'countXXX' not found for validation
	at io.smartdatalake.workflow.dataobject.expectation.BaseExpectation.$anonfun$getMetric$1(Expectation.scala:89)
	at scala.collection.immutable.Map$Map2.getOrElse(Map.scala:236)
	at io.smartdatalake.workflow.dataobject.expectation.BaseExpectation.getMetric(Expectation.scala:89)
	at io.smartdatalake.workflow.dataobject.expectation.BaseExpectation.getMetric$(Expectation.scala:87)
	at io.smartdatalake.workflow.dataobject.expectation.CountExpectation.getMetric(CountExpectation.scala:44)
	at io.smartdatalake.workflow.dataobject.expectation.CountExpectation.getValidationErrorColumn(CountExpectation.scala:81)
	at io.smartdatalake.workflow.dataobject.ExpectationValidation.$anonfun$validateExpectations$3(ExpectationValidation.scala:168)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:196)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:194)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:199)
[...]

Issue #889, opened 2024-08-27
CsvFileDataObject does not use schemaMin as its schema if there is no other definition


zzeekk:
**Describe the bug**
If only a schemaMin is defined for a CsvFileDataObject, and not schema, no sample file and there is no real data files, then CsvFileDataObject will throw a schema undefined error on read in init phase.

**To Reproduce**
Check stg-airport in getting-started

**Expected behavior**
The schemaMin should be taken as last resort if no other schema definition exists.

PR #906, opened 2024-10-29
Bugfix/896 count expectation

zzeekk:
see #896 
+ additional improvements of IcebergTableDataObject
+ HistorizeAction: fix not using schema in prepare phase (#900)
+ SparkFileDataObject: use schemaMin as schema if no other definition exists (#889)


PMuellerELCA (approved):
Approved

Issue #915, opened 2024-12-02
Add filter to ignore invalid Kafka messages


AdiWehrli:
**Is your feature request related to a problem? Please describe.**
Messages in a Kafka topic can be null or an array of byte with null elements.

**Describe the solution you'd like**
Those messages should be filtered out.

AdiWehrli:
Maybe we could accomplish this with adjustment of method `convertToReadDataFrame()` in `KafkaTopicDataObject` like this:
```scala
  private def convertToReadDataFrame(dfRaw: DataFrame): DataFrame = {
    import DataFrameUtil._

    // convert key & value
    val colsToSelect = ((if (selectCols.nonEmpty) selectCols else Seq("kafka.*")) ++ partitions).distinct.map(col)
    logger.info(s"Filtering dataFrame 'dfRaw' for incorrect values")
    val dfWithoutIncorrectValues = dfRaw
      .withColumn("key", col("key"))
      .withColumn("value", col("value"))
      .where(col("value").isNotNull && len(col("value")) > 0)
      .as("kafka-filter-incorrect-values")
      .withOptionalColumn(datePartitionCol.map(_.colName), udfFormatPartition(col("timestamp")))
    logger.info(s"Converting filtered dataFrame keys/values from Kafka")
    val df = dfWithoutIncorrectValues
      .withColumn("key", convertFromKafka(keyType, col("key"), SubjectType.key, keySchema))
      .withColumn("value", convertFromKafka(valueType, col("value"), SubjectType.value, valueSchema))
      .as("kafka")
      .withOptionalColumn(datePartitionCol.map(_.colName), udfFormatPartition(col("timestamp")))
      .select(colsToSelect:_*)
    // return
    df
  }
```

What do you think, @zzeekk?

zzeekk:
Hi @AdiWehrli, these are probably tombstone messages, see also https://medium.com/@damienthomlutz/deleting-records-in-kafka-aka-tombstones-651114655a16

I suggest to *not* apply the convertFromKafka conversion if a value is null (or empty array), but i would not filter them out by default. This should be done in an action/transformer later. Otherwise tombstone messages are lost for cases where they would be relevant.

AdiWehrli:
Hello @zzeekk 

Thanks for your response. I want to mention, that the topic's cleanup policy is not 'compact'.

Kind regards,
Adi

zzeekk:
[...]

PR #921, opened 2024-12-13
KafkaTopicDataObject: fix passing through tombstone messages (#915)

zzeekk:
see #915

### DIFF
--- a/sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
+++ b/sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
@@ -229,9 +229,10 @@ case class KafkaTopicDataObject(override val id: DataObjectId,
   }
 
   private def decodeKeyValue(dfRaw: DataFrame): DataFrame = {
+    assert(dfRaw.schema("value").dataType == BinaryType)
     dfRaw
       .withColumn("key", convertFromKafka(keyType, col("key"), SubjectType.key, keySchema))
-      .withColumn("value", when(col("value").isNotNull or length(col("value")) > 0, convertFromKafka(valueType, col("value"), SubjectType.value, valueSchema)))
+      .withColumn("value", when(col("value").isNotNull and length(col("value")) > 0, convertFromKafka(valueType, col("value"), SubjectType.value, valueSchema)))
   }
 
   override def getSparkDataFrame(partitionValues: Seq[PartitionValues] = Seq())(implicit context: ActionPipelineContext): DataFrame = {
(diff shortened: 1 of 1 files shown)


sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
```diff
@@ -349,11 +349,11 @@ case class KafkaTopicDataObject(override val id: DataObjectId,
 
   private def convertToWriteDataFrame(df: DataFrame): DataFrame = {
     require(df.columns.toSet == Set("key","value"), s"($id) Expects columns key, value in DataFrame for writing to Kafka. Given: ${df.columns.mkString(", ")}")
-    keySchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("key").dataType.asInstanceOf[StructType]), "read (keySchema)"))
-    valueSchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("value").dataType.asInstanceOf[StructType]), "read (valueSchema)"))
+    keySchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("key").dataType.asInstanceOf[StructType]), "write (keySchema)"))
```
zzeekk, line 355: note: this is just a cosmetic change for a log statement...

sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
```diff
@@ -347,21 +349,21 @@ case class KafkaTopicDataObject(override val id: DataObjectId,
...
+  private def encodeKeyValue(df: DataFrame): DataFrame = {
     require(df.columns.toSet == Set("key","value"), s"($id) Expects columns key, value in DataFrame for writing to Kafka. Given: ${df.columns.mkString(", ")}")
-    keySchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("key").dataType.asInstanceOf[StructType]), "read (keySchema)"))
-    valueSchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("value").dataType.asInstanceOf[StructType]), "read (valueSchema)"))
+    keySchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("key").dataType.asInstanceOf[StructType]), "write (keySchema)"))
+    valueSchema.foreach(schema => validateSchema(schema, SparkSchema(df.schema("value").dataType.asInstanceOf[StructType]), "write (valueSchema)"))
     df.select(
       convertToKafka(keyType, col("key"), SubjectType.key, keySchema).as("key"),
-      convertToKafka(valueType, col("value"), SubjectType.value, valueSchema).as("value")
+      when(col("value").isNotNull, convertToKafka(valueType, col("value"), SubjectType.value, valueSchema)).as("value")
```
iliassoto: When decoding the value in *decodeKeyValue*, we exclude values that are NULL **and** the ones that have length == 0. Here, we only convert nonNull values to Kafka. Why the difference?

iliassoto (commented):
See one comment. LGTM otherwise

sdl-kafka/src/main/scala/io/smartdatalake/workflow/dataobject/KafkaTopicDataObject.scala
zzeekk: Good question.
In decodeKeyValue a Binary column has to be decoded. If its value is an empty binary array, decoding probably fails.
In encodeKeyValue it is a column of arbitrary type. Checking for length is possible for Array, String and Binary, but is not defined for Struct or Numeric types.
To have the same definition we could check length=0 if it is a binary type. I have already adapted the code.

zzeekk:
@iliassoto: comment resolved, please approve again.
Note: we have changed a setting recently so that PRs can not be merge anymore if there are open conversations. So next time you can make comments to resolve, and approve the PR already, if you agree that we merge after all comments have been resolved...

iliassoto (approved):

PR #930, opened 2024-12-20
Update ODataDataObject doc

### LINKED ISSUE #929: Cleaning of ODataDataObject
The schema-inference of the ODataDataObject has been generalized to a GenericSchema, which is why one method is obsolote and the documentation can be generalized.

iliassoto:
See https://github.com/smart-data-lake/smart-data-lake/issues/929


PMuellerELCA (approved):
Approved

Issue #871, opened 2024-08-02
Implement ColumnTransformer


zzeekk:
**Is your feature request related to a problem? Please describe.**
The existing AdditionalColumnTransformer can be used to add new Columns to a DataFrame.
Often also dropping or renaming columns is needed, using a specific order of these operations.

This is not possible with AdditionalColumnTransformer. Currently this needs creating SQL Transformer, which can not be used without enumuerating all other columns of the DataFrame, or creating a ScalaCodeSparkDfTransformer, which is complicated to use.

**Describe the solution you'd like**
Create a ColumnTransformer which is configured with a list of Operations (op: String, column: String, expr: String), where `op` can be 'add', 'drop', 'rename'.

iliassoto:
@zzeekk can we close this issue thanks to https://github.com/smart-data-lake/smart-data-lake/pull/937?

PR #937, opened 2024-12-31
#871 implement generic column transformer

Geheiner:

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/ColumnsTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/ColumnsTransformer.scala
@@ -30,17 +30,28 @@ import io.smartdatalake.workflow.{ActionPipelineContext, DataFrameSubFeed}
 /**
  * Add, Rename or Drop columns from the Input DataFrame. The order of execution of the operations is the same as the parameters of the transformer: (1. add context info, 2. add custom sql, 3. rename, 4. drop).
  * Note that you can mix and match these operations to your liking and reference the same column from a previous operation if needed.
-  * @param name         name of the transformer
+ *
+ * @param name         name of the transformer
  * @param description  Optional description of the transformer
  * @param additionalColumns optional tuples of [column name, spark sql expression] to be added as additional columns to the dataframe. This allows you to include commonly used context information available at runtime.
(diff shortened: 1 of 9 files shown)


Geheiner:
Hi @zzeekk 
Instead of adding different modes, I just propose to give extra parameters to the existing tranformer. What do you think?
If it looks right to you as well I would rename the transformer and then add an Alias for backwardscompatibility as well as a test with an actual HOCON Config

zzeekk:
Yes, it's a good and simple proposal. I like it.
Please add also Scaladoc for the new attributes, and document the order in which the operations are applied (1. additional, 2. rename, 3. drop).

sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/ColumnsTransformer.scala
```diff
@@ -28,16 +28,30 @@ import io.smartdatalake.workflow.dataframe.GenericDataFrame
...
  *                          [[DefaultExpressionData]] contains informations from the context of the SDLB job, like runId or feed name.
- * @param additionalDerivedColumns optional tuples of [column name, spark sql expression] to be added as additional columns to the dataframe.
- *                                 The spark sql expressions are evaluated against the input DataFrame and added to the DataFrame as derived columns.
+ * @param additionalDerivedColumns optional tuples of [column name, spark sql expression] to be added as additional columns to the dataframe. This allows you to run custom sql code on the input DataFrame and save the result into a new column for which you define the name.
+ * @param renamedColumns optional tuples of [old column name, new column name]. For each tuple, a column is renamed. A RuntimeError will occur if a column to be renamed does not exist after applying all previous operations.
+ * @param droppedColumns optional list of column names to be dropped. A RuntimeError will occur if a column does not exist after applying all previous operations.
+ *
+ * Example Config:
```
Geheiner, line 43: @zzeekk What do you think about this? We need to have examples somewhere. I noticed it myself when I tried out my transformer how long it takes to get the config right when you don t do HOCON every day.
Before, I did big Unit-Tests that test a HOCON Config, but this has a big overhead. The advantage is of course that it never gets outdated What is better?

zzeekk (approved):
Ready to merge.
I needed to pull latest changes from develop-spark3 to fix Unit Tests.
Attention: i did a "force push" after rebasing onto develop-spark3.

PR #942, opened 2025-01-19
HadoopFileDataObject: use listFiles instead of globFiles

zzeekk:
Use listFiles instead of globFiles in HadoopFileDataObject.
As listFiles returns an iterator instead of an array as globFiles, it is much more memory efficient. Checking if a directory is empty is much faster.
This avoids driver out of memory exceptions to check if the DataObjects Hadoop directory has data files in environments with many files.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
@@ -21,11 +21,12 @@ package io.smartdatalake.workflow.dataobject
 import io.smartdatalake.config.InstanceRegistry
 import io.smartdatalake.config.SdlConfigObject.ConnectionId
 import io.smartdatalake.definitions.{Environment, SDLSaveMode, TableStatsType}
+import io.smartdatalake.util.hdfs.HdfsUtil.RemoteIteratorWrapper
 import io.smartdatalake.util.hdfs.{HdfsUtil, PartitionLayout, PartitionValues}
 import io.smartdatalake.util.misc.{AclDef, AclUtil, SmartDataLakeLogger}
 import io.smartdatalake.workflow.ActionPipelineContext
 import io.smartdatalake.workflow.connection.HadoopFileConnection
-import org.apache.hadoop.fs.{FileAlreadyExistsException, FileStatus, FileSystem, Path}
+import org.apache.hadoop.fs._
 
 import java.io.{FileNotFoundException, InputStream, OutputStream}
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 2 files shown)


sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
```diff
@@ -81,16 +82,32 @@ private[smartdatalake] trait HadoopFileDataObject extends FileRefDataObject with
...
-      val globPath = if (partitions.nonEmpty) new Path(hadoopPath, PartitionValues(Map()).getPartitionString(partitionLayout().get)) else hadoopPath
-      status.isDirectory && filesystem.globStatus(new Path(globPath, fileName)).exists(_.isFile)
-    }
+    status.isFile || (status.isDirectory && listDataFiles().nonEmpty)
+  }
+
+  def listDataFiles(pv: PartitionValues = PartitionValues(Map()))(implicit context: ActionPipelineContext): Iterator[Path] = {
+    val pathPattern = if (partitions.nonEmpty) new GlobPattern(new Path(new Path(hadoopPath, pv.getPartitionString(partitionLayout().get)), fileName).toString)
+    else new GlobPattern(new Path(hadoopPath, fileName).toString)
+    RemoteIteratorWrapper(filesystem.listFiles(hadoopPath, partitions.nonEmpty /*recursive*/))
```
Geheiner: I cannot see where the recursion is by looking at the code without IntelliJ... Will try in an IDE tomorrow

Geheiner: This is not recursive, right? Can you remove the /*recursive*/ comment?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
```diff
@@ -81,16 +82,32 @@ private[smartdatalake] trait HadoopFileDataObject extends FileRefDataObject with
...
+      .filter(_.isFile)
+      .map(_.getPath)
+      .filter(p => pathPattern.matches(p.toString))
+  }
+
+  def listPartitionPathsStatus(pv: PartitionValues = PartitionValues(Map()), partitionLayoutParam: String = partitionLayout().get)(implicit context: ActionPipelineContext): Seq[FileStatus] = {
+    assert(partitions.nonEmpty)
+    val pathPattern = new Path(hadoopPath, pv.getPartitionString(partitionLayoutParam))
+    filesystem.globStatus(pathPattern).filter(_.isDirectory)
```
Geheiner: So you just wanted to use globFiles a bit less? It's still used

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
```diff
@@ -81,16 +82,32 @@ private[smartdatalake] trait HadoopFileDataObject extends FileRefDataObject with
...
+  def listDataFiles(pv: PartitionValues = PartitionValues(Map()))(implicit context: ActionPipelineContext): Iterator[Path] = {
+    val pathPattern = if (partitions.nonEmpty) new GlobPattern(new Path(new Path(hadoopPath, pv.getPartitionString(partitionLayout().get)), fileName).toString)
+    else new GlobPattern(new Path(hadoopPath, fileName).toString)
+    RemoteIteratorWrapper(filesystem.listFiles(hadoopPath, partitions.nonEmpty /*recursive*/))
+      .filter(_.isFile)
+      .map(_.getPath)
+      .filter(p => pathPattern.matches(p.toString))
+  }
+
+  def listPartitionPathsStatus(pv: PartitionValues = PartitionValues(Map()), partitionLayoutParam: String = partitionLayout().get)(implicit context: ActionPipelineContext): Seq[FileStatus] = {
```
Geheiner, line 103: we could make 3 methods private. Do we keep them public on purpose to be able to extend them?

Geheiner (approved):
Looks good, I just have a few questions in the comments for clarification

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
zzeekk: Yes, but here we filter for a directories already with the glob pattern, so the list is not large.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataobject/HadoopFileDataObject.scala
zzeekk: True. On the other side i often struggle with methods not being accessible when debugging or creating custom implementations. I'm no longer a fan of making everything private/protected where possible.
In Python real private methods dont exist, they are just marked by a prefix "__". We could learn from them ;-)

PR #944, opened 2025-01-28
Upload started event

zzeekk:
- implement uploading PREPARING/INITIALIZING/RUNNING events to UI. This allows to show running actions in the UI. If the UI only receives an event when an Action phase is finished, no timeline of running actions can be shown.
- BuildVersionInfo: fix catching fatal exception if git command fails
- Store LocalDateTime as UTC datetime strings in state files


iliassoto (approved):
Looks good to me. We can clean up the UI source code to not expect the stage metrics anymore (e.g. https://github.com/smart-data-lake/sdl-visualization/blob/develop/src/util/WorkflowsExplorer/randomStateFile.ts or https://github.com/smart-data-lake/sdl-visualization/blob/develop/src/types.ts)

Issue #947, opened 2025-02-11
Missing dev.conf file in envConfig for getting-started setup


JessicaBL:
When running buildJob.sh I experienced an error stating that the dev.conf file could not be found. When looking into the folder, there is only a file called dev.conf.part-2-solution.
The dev.conf file is also referenced in the pom.xml leading to the above mentioned error. So either the pom.xml needs to be adapted or the dev.conf is really missing?

PR #948, opened 2025-02-11
Enhance documentation for technical setup

JessicaBL:
- Adapt the documentation for the technical setup in the getting started part
- Add hint to copy dev.conf.part-2-solution as dev.conf

See #947


iliassoto:
Looks good. As discussed, as of now, everyone will encounter this error, so we should fix the issue directly in the getting-started repo (see https://github.com/smart-data-lake/getting-started/issues/36)

PR #952, opened 2025-03-10
#939: Remove TickTockHiveTableDataObject and all references, improve handling for JdbcTableDataObject

### LINKED ISSUE #939: Remove TickTockHiveTableDataObject
TickTockHiveTableDataObject is no longer needed as all known users have migrated away from Hadoop.
It was used to have a TransactionalTableDataObject on Hive/Hadoop. We have other TransactionTableDataObject (DeltaLake, Iceberg, Jdbc) to be used in current technologies.

Remove TickTockHiveTableDataObject from version 2.8.0.
Enable merge mode for Historize- and DeduplicateAction and remove old code implementing "full" historize and deduplicate.

JessicaBL:
- Remove all references to TickTockHiveTableDataObject
- Replace TickTockHiveTableDataObject by JdbcTableDataObject where TransactionalTableDataObject is needed
- Improve handling of JdbcTableDataObject with case insensitivity checks where applicable to fix issues in tests
- Add deprecated message for enableMergeMode to give users the chance to make necessary changes before removing the parameter completely

see #939


sdl-core/src/main/scala/io/smartdatalake/util/hdfs/Partition.scala
```diff
@@ -209,6 +209,9 @@ object PartitionLayout {
   }
 
   def extractPartitionValues(partitionLayout: String, path: String): PartitionValues = {
+    val cleanPath = if (Environment.caseSensitive) path
+    else "^(.*?)=".r.replaceAllIn(path, m => m.group(0).toLowerCase)
```
zzeekk: This only matches the beginning of the path, why? Shouldn't we make the whole path lowercase?

sdl-core/src/main/scala/io/smartdatalake/util/misc/SchemaUtil.scala
```diff
@@ -217,6 +218,24 @@ object SchemaUtil {
...
+    if (caseSensitive) colsLeft.diff(colsRight)
+    else colsLeft.map(_.toLowerCase).diff(colsRight.map(_.toLowerCase))
+  }
+
+  def checkPartitionMatch(configuredPartitions: Seq[String], existingPartitions: Seq[String], caseSensitive: Boolean): (Boolean, Set[String], Set[String]) = {
+    val (confPartitions, existPartitions) = if (caseSensitive) {
+      val conf = configuredPartitions.toSet
+      val existing = existingPartitions.toSet
+      (conf, existing)
+    } else {
```
zzeekk: Why not a little bit shorter?
```suggestion
    val (confPartitions, existPartitions) = if (caseSensitive) (configuredPartitions.toSet, existingPartitions.toSet)
    else {
```

sdl-core/src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
```diff
@@ -225,9 +224,10 @@ private[smartdatalake] object HiveUtil extends SmartDataLakeLogger {
 
     // check if partitionsOpt match with existing table definition
     if (tableExists) {
-      val configuredCols = partitions.toSet
-      val existingCols = getTablePartitionCols(table).getOrElse(Seq()).toSet
-      require( configuredCols==existingCols, s"(${table.fullName}) writeDfToHive: configured vs existing partition columns are different: configured=$configuredCols, existing=$existingCols" )
+      val existingPartitions = getTablePartitionCols(table).getOrElse(Seq())
+      val (partitionsMatch, configuredCols, existingCols) = SchemaUtil.checkPartitionMatch(partitions, existingPartitions, Environment.caseSensitive)
+
+      require( partitionsMatch, s"(${table.fullName}) writeDfToHive: configured vs existing partition columns are different: configured=$configuredCols, existing=$existingCols" )
```
zzeekk, line 230: It would also be ok to use partitions and existingPartitions directly, maybe even clearer about the origin of different columns. WDYT?

zzeekk (commented):
Looks good to me.
Some minor comments to improve before merge.

For merging: please use "Squash and merge" and add the issue number in the new commit message as follows: "<msg> (#<issuenb>)"

sdl-core/src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
JessicaBL: What do you mean? To put e.g. getTablePartitionCols(table)... directly as parameter into the checkPartitionMatch function? Or to remove the function as it's only used in this case here and do the logic directly here?

sdl-core/src/main/scala/io/smartdatalake/util/hdfs/Partition.scala
JessicaBL: You mean also the part after = ? I think that would mess up the data. Or are there also possibilities to have a path like partition_X/partition_Y=anything?

zzeekk: A path can have multiple partition columns, and i think its not sure that it is relative. Examples would be:
- partition_X=aBC/partition_Y=Def/test.json
- /some/root/path/partition_X=aBC/partition_Y=Def/test.json
I suggest to adapt the Regex as follows:
```suggestion
    else "(?:^|/)([^/=]*)=".r.replaceAllIn(path, m => m.group(0).toLowerCase)
```

sdl-core/src/main/scala/io/smartdatalake/util/hive/HiveUtil.scala
zzeekk: I mean to simplify `SchemaUtil.checkPartitionMatch` so that it returns only partitionsMatch, and use $partitions and $existingPartitions in require text:
```suggestion
      require( partitionsMatch, s"(${table.fullName}) writeDfToHive: configured vs existing partition columns are different: configured=$partitions, existing=$existingPartitions" )
```

zzeekk: As discussed, lets keep it like it is.

zzeekk (approved):
LGTM

PR #961, opened 2025-04-21
Implement time interval between vacuum operations

### LINKED ISSUE #951: DeltaLakeTableDataObject: More control for vacuum execution
Currently if the field retentionPeriod is set for DeltaLakeTableDataObject, vacuum will be executed on every write operation to satisfay the retentionPeriod.

More fine-grained control would be nice. For example, being able to set that vacuum should run every 7 days.

**Additional context**
To implement this, Metadata information from the Tablehistory can be used to check for the last write date of the table.

iliassoto:
see https://github.com/smart-data-lake/smart-data-lake/issues/951

The implementation uses ISO 8610 Duration for time interval definition.


Geheiner (approved):
Nice! LGTM

Issue #804, opened 2024-03-13
New DataObject for change data capture (CDC) with Debezium


zzeekk:
**Is your feature request related to a problem? Please describe.**
To synchronize database tables incrementally to an analytics platform, we can use DataFrameIncrementalMode, which needs a strictly increasing column as "high watermark" + a primary key for merging changes into the target. This approach can not synchronize physically deleted records, and it does not work if there is no strictly increasing column for a given table.

This can be solved with change data capture, which reads database logs to get the change information.

**Describe the solution you'd like**
Debezium is a Java project that implements change data capture for different databases, and abstracts away the differences by providing a generic interface. The Debezium engine is a java library that could be well integrated with SDLB. 

Debezium engine can create Avro Records, that we could easily decode with Sparks from_avro function. Additionally there is state to be managed, for which we could use DataObjectStateIncrementalMode.

I therefore propose to implement a DebeziumJdbcTableDataObject that implements DataObjectStateIncrementalMode. Further we should create another trait that marks DataObjects that can deliver CDC data, so that HistorizeAction can switch into CDC mode automatically.

**Additional context**

Calling Debezium Engine in Airbyte: https://github.com/airbytehq/airbyte/blob/0f2685be4f2ce2705a384a1091b77743638001eb/airbyte-cdk/java/airbyte-cdk/db-sources/src/main/java/io/airbyte/cdk/integrations/debezium/internals/DebeziumRecordPublisher.java#L49

Calling Debezium Engine in Kestra:  https://github.com/kestra-io/plugin-debezium/blob/master/plugin-debezium/src/main/java/io/kestra/plugin/debezium/AbstractDebeziumTask.java#L150
https://medium.com/@kestra-io/debezium-change-data-capture-without-kafka-connect-18f43cf095d2

PR #965, opened 2025-05-04
804 new dataobject for change data capture cdc with debezium

ThierryGirod:
 
Add new connection for Debezium.
Add new data object for Debezium.
Add shaded modules for mariadb and mysql connector.

Limitations:
-> Mariadb and Mysql connector had to be shaded because of conflicting ANTLR Libraries. This means if interacting with these connectors and f.ex. the S3 Schema History, the prefix "shaded." needs to be added to all class package names.

To fullfill the needs of Issue #804


sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+                                 table: Table,
+                                 debeziumProperties: Option[Map[String, String]] = None,
+                                 maxWaitTimeInSeconds: Int = 10,
+                                 override val metadata: Option[DataObjectMetadata] = None)
+                                (@transient implicit val instanceRegistry: InstanceRegistry)
+  extends DataObject with CanCreateDataFrame with CanCreateSparkDataFrame with CanCreateIncrementalOutput {
+
+  val connection: DebeziumConnection = getConnection[DebeziumConnection](connectionId)
+
+  private def debeziumPropertiesForEngine: Properties = {
```
Geheiner, line 54: Is there a particular reason why this is a def and not a val?

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+
+      val engine = DebeziumEngine.create(classOf[Connect])
+        .using(properties)
+        .notifying(changeConsumer)
+        .using(completionCallback)
+        .build()
+
+      executorService.execute(engine)
+
+      Thread.sleep(timeoutMilliSeconds)
```
Geheiner: this is the first thing I find, after some searching, that looks not so clean. Why do you need to launch a thread in the background and then wait 10 seconds but still return the records without ever waiting for the Thread to complete? I'm wondering how can you know when "everything" has been fetched?

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+
+/**
+ * Helper object to convert from debezium events in SourceRecord format to sdlb compatible spark dataframe
+ */
+private object DebeziumEventConverter {
+
+  def convert(records: Seq[SourceRecord])(implicit spark: SparkSession): DataFrame = {
+
+    val sparkSchema = inferSparkSchema(records.head.valueSchema())
+    val rows = records.map {
```
Geheiner: records.map(recordToRow)

would also be possible and more concise, just so you know

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+        case s: Struct => structToRow(s) // Recursively handle nested structs
+        case _ => fieldValue // Primitive types
+      }
+    }.toSeq
+
+    Row(values: _*)
+
+  }
+
+  private def flattenDf(df: DataFrame): DataFrame = {
```
Geheiner: Did you derive this from io.smartdatalake.workflow.action.spark.transformer.SparkFlattenDfTransformer.flattenDf ?
Is your method Debezium specific? If yes then maybe call it flattenDebeziumDF?

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+      .withColumn("event_data", coalesce(col("after"), col("before")))
+      .withColumn(COMMIT_TYPE_COLUMN_NAME,
+        when(col("op") === "c", lit("create"))
+          .when(col("op") === "d", lit("delete"))
+          .otherwise(lit("read")))
+
+    val unionDf = updateBeforeDf.union(updateAfterDf).union(otherOperationsDf)
+      .withColumn(COMMIT_TIMESTAMP_COLUMN_NAME, from_unixtime(col("source.ts_ms") / 1000).cast(TimestampType))
+      .drop("before", "after", "source", "op", "ts_ms", "ts_us", "ts_ns", "transaction")
+      .orderBy(col(COMMIT_TIMESTAMP_COLUMN_NAME).desc)
```
Geheiner: do you need to always to this sorting? I can imagine that it s costly and as soon as you do any other DF Operation after that your ordering is again not guaranteed anymore

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+  }
+
+  private[smartdatalake] var incrementalState: mutable.Map[String, String] = mutable.Map()
+
+  /**
+   * To implement incremental processing this function is called to initialize the DataObject with its state from the last increment.
+   * The state is just a string. It's semantics is internal to the DataObject.
+   * Note that this method is called on initializiation of the SmartDataLakeBuilder job (init Phase) and for streaming execution after every execution of an Action involving this DataObject (postExec).
```
Geheiner: with streaming do you mean actual spark streaming here? Probably not the sdlb in "streaming mode"

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
+
+import java.util
+import java.util.Properties
+import java.util.concurrent.{ExecutorService, Executors}
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+
+case class DebeziumCdcDataObject(override val id: DataObjectId,
```
Geheiner, line 73: Overall I find it very easy to follow. Kudos to Thierry and also to those that designed the SDLB Architecture patterns, I find that each feature ends up with a similar structure, thank god for that.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+        .build()
+
+      executorService.execute(engine)
+
+      Thread.sleep(timeoutMilliSeconds)
+
+      engine.close()
+      executorService.shutdown()
+
+      completionCallback.error.foreach(err => throw new Exception(err))
```
Geheiner, line 162: I did not go through the tests yet, maybe this case is treated there, but:
How do we handle failures ? In the Doc it often talks about at least once delivery and that duplicates are possible. Did you experience issues like this in your tests?

I guess this exception is for a generic failure of Debezium before anything gets consumed?

I'm just wondering how we are addressing this part of the doc:

"Note that your application’s handler function should not throw any exceptions; if it does, the engine will log any exception thrown by the method and will continue to operate on the next source record, but your application will not have another chance to handle the particular source record that caused the exception, meaning your application might become inconsistent with the database."

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
```diff
@@ -0,0 +1,132 @@
+# Setup of databases for integration tests
+
+This document describes how the databases were setup to run the integration tests. 
+While following the official [debezium (3.1.0) documentation](https://debezium.io/documentation/reference/3.1/connectors/index.html) for all the supported connectors, some additions are made, f.ex. for Oracle database there exists no out-of-the-box image on Dockerhub. 
+You have to build it first.
+
+
+**Important Note:** Not all databases were explicitly tested as part of this integration test setup. 
+The tests primarily focused on databases that include `org.antlr` as a transitive dependency.
```
Geheiner: Can you elaborate on that? I assume you mean the dependency conflicts that you fixed with the shading?

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectPostgresqlIT.scala
```diff
@@ -0,0 +1,188 @@
...
+import io.smartdatalake.workflow.action.{ActionMetadata, CopyAction}
+import io.smartdatalake.workflow.connection.DebeziumConnection
+import io.smartdatalake.workflow.connection.authMode.BasicAuthMode
+import io.smartdatalake.workflow.connection.jdbc.JdbcTableConnection
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.sql.functions.{col, lit}
+
+import java.nio.file.Files
+
+object DebeziumCdcDataObjectPostgresqlIT extends App with SmartDataLakeLogger {
```
Geheiner, line 39: Please provide your config to start the test so that I don't have to manually discover the env variables. You could put the IntelliJ Runtime Config into the folder .idea folder that is located on the root level of the sdl project.
I assume the correct env variables are:
PSQL_HOSTNAME=localhost;PSQL_DB=test;PSQL_PORT=5432;PSQL_USER=postgres;PSQL_PASSWORD=psql_demo

But I still get error message
Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
```diff
@@ -0,0 +1,132 @@
...
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
+
+```
+
+## PostgreSQL
+Server can be setup with the following docker / podman command: 
+````shell
+podman run -d --name postgres -p 5432:5432 -e POSTGRESQL_WAL_LEVEL=logical -e POSTGRESQL_DATABASE=test -e POSTGRESQL_POSTGRES_PASSWORD=psql_demo -e POSTGRESQL_PASSWORD=psql_demo -e POSTGRESQL_SHARED_PRELOAD_LIBRARIES=pgaudit,pgoutput bitnami/postgresql:17.4.0
+````
```
Geheiner: Also add the docker commands to connect to postgres and execute the sql scripts.
In my case:
````Run 
podman ps 
to discover container id of postgres.
Then with your container id (46ea388a9592 in this example), 
Run:
podman exec -it 46ea388a9592 bash

Then you are located inside the pod and you may connect to postgres by using this command (enter password shown above):
psql -U postgres
````

**Comment by Geheiner** [comment]
When merged, shall we add ourselves here ?
https://debezium.io/community/users/
I think at some point it's good to get pointed onto by the community

File: sdl-debezium/src/main/scala/io/smartdatalake/workflow/connection/DebeziumConnection.scala [hunk 12]
```diff
@@ -0,0 +1,110 @@
...
+import io.debezium.connector.oracle.OracleConnector
+import io.debezium.connector.postgresql.PostgresConnector
+import io.debezium.connector.spanner.SpannerConnector
+import io.debezium.connector.sqlserver.SqlServerConnector
+import io.debezium.connector.vitess.VitessConnector
+import io.smartdatalake.config.SdlConfigObject.ConnectionId
+import io.smartdatalake.config.{FromConfigFactory, InstanceRegistry}
+import io.smartdatalake.workflow.connection.authMode.{AuthMode, BasicAuthMode}
+
+case class DebeziumConnection(override val id: ConnectionId,
```
zzeekk, line 53: Please add some scaladoc describing the parameters, it will be displayed in sdl-visualizer.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,329 @@
...
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
+
+import java.util
+import java.util.Properties
+import java.util.concurrent.{ExecutorService, Executors}
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+
```
zzeekk, line 46: Please add some scaladoc describing the parameters, it will be displayed in sdl-visualizer.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/connection/DebeziumConnection.scala
```diff
@@ -0,0 +1,110 @@
...
+object DebeziumConnection extends FromConfigFactory[Connection] {
+  override def fromConfig(config: Config)(implicit instanceRegistry: InstanceRegistry): DebeziumConnection = {
+    extract[DebeziumConnection](config)
+  }
+}
+
+private object DbEngineHelper {
+
+  private val dbEngineConnectorClassNames: Map[String, String] = Map(
+    "mysql" -> classOf[MySqlConnector].getName,
```
zzeekk: classOf[xyz] needs the corresponding class to be available in the classpath... i would like to avoid that a client needs to have all connectors in the classpath. I think normally he just needs one connector.

We could change this to a match / case structure, so it is just executed when a specific connector is looked up. And then catch and wrap the exception with a hint for the dependency to be added...

Exception is:
java.lang.NoClassDefFoundError: io/debezium/connector/postgresql/PostgresConnector
	at io.smartdatalake.workflow.connection.DbEngineHelper$.<init>(DebeziumConnection.scala:94)

ThierryGirod: Do you mean something like this?


private val dbEngineConnectorClassName: String => String = { 

case "mysql" => try {
 classOf[MySqlConnector].getName
} catch {
   case e: NoClassDefFoundError => throw new RuntimeException("MySQL Connector not found on the classpath... Add dependency)
}
...
}

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
ThierryGirod: Can you connect to the database successfully with e.g. DBeaver?

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
ThierryGirod: No implemented by myself. Will rename it to be more specific.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
ThierryGirod: I copied from another class from zach and me where we use this. The "streaming" is actually the sdlb streaming mode. Not spark streaming.

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
ThierryGirod: No not only the shaded ones. I also excluded `org.antlr` e.g. from the transitive dependencies in the Oracle connector. But it seems only Mysql and Mariadb actually use some functionality from `org.antlr`. The other connectors were not testet because they don't have `org.antlr` as a transitive dependency. I think to test them properly is the job of the Debezium guys 😄

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectPostgresqlIT.scala
ThierryGirod: Yes, the env variables are: 

PSQL_DB=test;PSQL_HOSTNAME=localhost;PSQL_PASSWORD=psql_demo;PSQL_PORT=5432;PSQL_USER=postgres

Can you connect to the database successfully with e.g. DBeaver?

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
ThierryGirod: No, I did't experienced any issues with duplicates. I had several failures sometimes but sdlb still managed to handle it correctly.  I could imagine that a problem only occures with Debezium when the SDLB data object state is not persistet, but the data already written to target.  Because then the Debezium offset is in a wrong state and all the records would be read again for consumtion.

@zzeekk What do you think? Could this be an issue somehow?

And yes you guessed right, this is to check if Debezium crashed somehow. As long you set the offset correctly, Debezium reads from that position. And since we are in charge of how we store the the offset I don't see a problem here.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
ThierryGirod: Nope, thats something I had once for debugging. Good spot, totally agree with costly operation!

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
Geheiner: Then I suggest this:
```suggestion
   * Note that this method is called on initializiation of the SmartDataLakeBuilder job (init Phase). When starting SDLB with the --streaming option, if will be called after every execution of an Action involving this DataObject (postExec).
```
Actually, the streaming Option should be renamed to micro batching cos that's what this is, but this is for another time...

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
Geheiner: I'm running podman on Ubuntu WSL, but I'm starting the Test in IntelliJ. Probably your setup is different. For me I had to specifiy --network host
```suggestion
````podman run -d --name postgres --network host -e POSTGRESQL_WAL_LEVEL=logical -e POSTGRESQL_DATABASE=test -e POSTGRESQL_POSTGRES_PASSWORD=psql_demo -e POSTGRESQL_PASSWORD=psql_demo -e POSTGRESQL_SHARED_PRELOAD_LIBRARIES=pgaudit,pgoutput bitnami/postgresql:17.4.0

```
Further, I was able to run the tests successfully when connecting to the DB with DBeaver and creating the schema there instead of using psql.
When using psql the schema is somehow not visible to anyone outside podman, it s very strang...

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
Geheiner: Ok, then I suggest something like this (please improve if you can :D)
```suggestion
Because of incompatibilities of the versions of the library org.antlr between spark and debezium we had to resort to shading for the debezium connectors of mariadb and mysql. Therefore, the tests primarily focused on databases that include `org.antlr` as a transitive dependency to test that everything still works after doing the shading
```
Further, I suggest to put Postgresql as the first paragraph since it's the one that 90% of people will be using

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectPostgresqlIT.scala
Geheiner: works now. please click this button in IntelliJ and put the file into a new folder sdl-debezium/.idea
image

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectPostgresqlIT.scala
Geheiner, line 1: Nice Test structure! Again, easy to follow and consistent. Thanks !

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectPostgresqlIT.scala
Geheiner, line 38: Can you add an example config in the doc? Something like this?
https://github.com/smart-data-lake/smart-data-lake/blob/master-spark3/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/ColumnsTransformer.scala#L43

sdl-debezium/src/main/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObject.scala
```diff
@@ -0,0 +1,337 @@
...
+/**
+ * [[DataObject]] of type DebeziumCdcDataObject.
+ * Provides details to access Change data over Debezium Engine.
+ *
+ * @param id unique name of this data object
+ * @param connectionId optional id of [[io.smartdatalake.workflow.connection.DebeziumConnection]]
+ * @param table Source table to get change data from
+ * @param debeziumProperties Properties for the specific Debezium connector
+ * @param metadata optional data object metadata
```
Geheiner: Can you add an example config in the doc? Something like this?
https://github.com/smart-data-lake/smart-data-lake/blob/master-spark3/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/ColumnsTransformer.scala#L43

sdl-debezium/src/main/scala/io/smartdatalake/workflow/connection/DebeziumConnection.scala
```diff
@@ -0,0 +1,121 @@
...
+ * Connection information for debezium connection
+ *
+ * @param id unique id of this connection
+ * @param dbEngine what database engine to use, currently supported engines: mysql, postgresql, oracle, mariadb, mongodb, sqlserver, db2, vitess, spanner
+ * @param hostname database server
+ * @param db database to read data from
+ * @param port
+ * @param authMode authentication information: for now BasicAuthMode is supported.
+ * @param metadata optional connection metadata
+ */
```
Geheiner, line 52: Can you add an example config in the doc? Something like this?
https://github.com/smart-data-lake/smart-data-lake/blob/master-spark3/sdl-core/src/main/scala/io/smartdatalake/workflow/action/generic/transformer/ColumnsTransformer.scala#L43

Geheiner, line 52: Add your Limitations Paragraph Also as a doc here?
Mariadb and Mysql connector had to be shaded because of conflicting ANTLR Libraries. This means if interacting with these connectors and f.ex. the S3 Schema History, the prefix "shaded." needs to be added to all class package names.

Geheiner (approved):
Thanks for giving me something fun to look at. Looks cool. Let's discuss the remaining open points together, but I see no risk of merging this (if it builds :D )

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
```diff
@@ -0,0 +1,132 @@
...
+
+## MariaDB
+Server can easily be setup with an oneliner as docker / podman command:
+```shell
+ podman run -d --name mariadb -p 3306:3306 -e MARIADB_ROOT_PASSWORD=debezium -e MARIADB_REPLICATION_MODE=master -e MARIADB_REPLICATION_USER=debezium -e MARIADB_REPLICATION_PASSWORD=debezium -e MARIADB_EXTRA_FLAGS="--log-bin=mysql-bin --binlog-format=ROW" bitnami/mariadb:11.4.5
+```
+
+Then run following ddl (one after another):
+```sql
+CREATE DATABASE `demo` /*!40100 DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci */;
```
zzeekk:
```suggestion
CREATE DATABASE demo /*!40100 DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci */;
```

sdl-debezium/src/test/resources/TEST_DB_SETUP.md
```diff
@@ -0,0 +1,132 @@
...
+  `decimalCol` decimal(6,3) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
+```
+```sql
+CREATE TABLE `test2` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `value` varchar(100) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```
zzeekk:
```suggestion
CREATE TABLE test (
  id int NOT NULL AUTO_INCREMENT,
  value varchar(100) DEFAULT NULL,
  timestampCol timestamp NULL DEFAULT NULL,
  decimalCol decimal(6,3) DEFAULT NULL,
  PRIMARY KEY (id)
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;

CREATE TABLE test2 (
  id int NOT NULL AUTO_INCREMENT,
  value varchar(100) DEFAULT NULL,
  PRIMARY KEY (id)
) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectMySqlIT.scala
```diff
@@ -0,0 +1,177 @@
...
+import io.smartdatalake.workflow.action.{ActionMetadata, CopyAction}
+import io.smartdatalake.workflow.connection.authMode.BasicAuthMode
+import io.smartdatalake.workflow.connection.jdbc.JdbcTableConnection
+import io.smartdatalake.workflow.connection.DebeziumConnection
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.sql.functions.{col, lit}
+
+import java.nio.file.Files
+
+object DebeziumCdcDataObjectMySqlIT extends App with SmartDataLakeLogger {
```
zzeekk, line 39: I get the following exception when executing this on demo Database with empty tables:
```
Exception in thread "main" io.smartdatalake.util.dag.TaskFailedException: Task copyAction1 failed. Root cause is 'NoSuchElementException: head of empty list'
Caused by: java.util.NoSuchElementException: head of empty list
        at scala.collection.immutable.Nil$.head(List.scala:469)
        at scala.collection.immutable.Nil$.head(List.scala:466)
        at io.smartdatalake.workflow.dataobject.DebeziumEventConverter$.convert(DebeziumCdcDataObject.scala:206)
```

zzeekk (approved):
Nice work!
Will push the improved shading right afterwards.
Some comments to solve before merging. Looks fine otherwise.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/connection/DebeziumConnection.scala
zzeekk: Yes, correct. Maybe it's overkill to write the dependency to add for every connector. We can also have just have one try/catch saying that you should add the debezium connector dependency for $dbName.

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectPostgresqlIT.scala
ThierryGirod: Done

sdl-debezium/src/test/scala/io/smartdatalake/workflow/dataobject/DebeziumCdcDataObjectMySqlIT.scala
ThierryGirod: This is fixed now. As discussed if there is an empty table with no events, the schemaMin is taken to generate the schema in the init and exec phase. If no schemaMin is available, an error message is displayed with a hint to set the schemaMin.

sdl-debezium/src/main/scala/io/smartdatalake/workflow/connection/DebeziumConnection.scala
ThierryGirod: @zzeekk  i corrected it. Please have a look at the final result.

Postgres test and mysql test run successfull

sdl-debezium/src/main/scala/io/smartdatalake/workflow/connection/DebeziumConnection.scala
ThierryGirod: After the rework of @zzeekk  I did not have to prefix the class anymore. Seems working now.

zzeekk (approved):
Retested and works :-)
Lets go.

PR #968, opened 2025-05-08
823 replace scala4jhttp with sttp

### LINKED ISSUE #823: ScalaJ Http Library is deprecated -> replace with sttp
We use ScalaJ Http Library to query Webservices in different places.
It is deprected and no longer maintained, see https://github.com/scalaj/scalaj-http/blob/master/README.md

ScalaJ supports sttp or http4s as replacement. It seems to me that sttp is a good replacement with similar API.
We should replace ScalaJ everywhere through sttp.

**Additional context**
sttp could also help on #815, as there is a specific library to parse api specification: https://github.com/softwaremill/sttp-apispec

iliassoto:
- Replaced ScalaJ-functionality with sttp functionality using existing code. The new class SttpWebserviceClient is written as close as possible regarding method signatures as its ScalaJ predecessor to avoid more refactoring.
- Marked ScalaJ-Classes as deprecated --> To be revmoved in a future release
- Moved some sttp-functionality to a central place for readability reasons.

see #823


zzeekk (commented):
Is there a reason that ScalaJWebserviceClient is just annotated as deprecated, or can we also remove ScalaJWebserviceClient and the dependency in the pom?

iliassoto:
@zzeekk just as a safety measure in case we encounter unexpected behaviour. Other than that there is no reason to keep the ScalaJ Client and the dependency.

zzeekk:
> @zzeekk just as a safety measure in case we encounter unexpected behaviour. Other than that there is no reason to keep the ScalaJ Client and the dependency.

Ok, i removed it.

zzeekk (approved):
Lets merge :-)

PR #969, opened 2025-05-09
Configure mutliple plugins option

### LINKED ISSUE #959: Plugin Class variable takes a list of classes
Using the SDLB plugin environment variable for enabling plugins, e.g. for redirecting log4j to logAnalytics or to the console. I want to be able to enable multiple plugins.

The environment variable support taking a list of classes.

iliassoto:
Include a new option / env variable SDL-Plugins to allow more than one plugin to be configured.
The configuration is done with a List of configuration options. These options have to have a key "className"

See #959

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
@@ -22,7 +22,7 @@ package io.smartdatalake.app
 import com.typesafe.config.Config
 import configs.ConfigReader
 import configs.syntax._
-import io.smartdatalake.config.ConfigImplicits
+import io.smartdatalake.config.{ConfigImplicits, ConfigurationException}
 import io.smartdatalake.config.SdlConfigObject.DataObjectId
 import io.smartdatalake.definitions.Environment
 import io.smartdatalake.util.misc.{LogUtil, MemoryUtils, SmartDataLakeLogger}
@@ -70,6 +70,8 @@ import org.apache.spark.util.PrivateAccessor
  * @param pluginOptions  Options for SDLPlugin initialization.
  *                       Note that SDLPlugin.startup is executed before SDLB parses the config, and pluginOptions are only available later when calling SDLPlugin.configure method.
  *                       An SDLPlugin is set through Environment.plugin, normally this is configured through the java system property "sdl.pluginClassName".
+ * @param pluginsOptions  List of options for SDLPlugin initialization in case there are more than one plugin defined. Note that for each options block (Map of configuration KV-pairs)
+ *                        a key 'className' must be provided in order to assign the configuration to the relevant plugin.
  * @param uiBackend      Configuration of the UI backend to upload state updates of the Job runs.
  */
 case class GlobalConfig(kryoClasses: Option[Seq[String]] = None
@@ -87,6 +89,7 @@ case class GlobalConfig(kryoClasses: Option[Seq[String]] = None
                         , synchronousStreamingTriggerIntervalSec: Int = 60
                         , environment: Map[String, String] = Map()
                         , pluginOptions: Map[String, StringOrSecret] = Map()
+                        , pluginsOptions: Seq[Map[String, StringOrSecret]] = Seq()
                         , uiBackend: Option[UIBackendConfig] = None
                        )
 extends SmartDataLakeLogger {
@@ -105,6 +108,21 @@ extends SmartDataLakeLogger {
     SecretsUtil.registerProvider(id, providerConfig.provider)
   }
 
+  /**
+   * If more than one plugin is defined
+   * 1. check that all options have a className key.
+   * 2. Check that all definfed plugins have been configured
+   * 3. configure
+   */
+  if (Environment.sdlPlugins.isDefined) {
+    val pluginList = Environment.sdlPlugins.get
+    val pluginsOptionsClassKey = "className"
+    if (!pluginsOptions.forall(_.get(pluginsOptionsClassKey).isDefined)) throw ConfigurationException(s"Wrong configuration of pluginsOptions: Each plugin-option block must have a '$pluginsOptionsClassKey' key to match the corresponding plugin")
+    val missingOptions = (pluginList.map(_.className).toSet) diff (pluginsOptions.map(_(pluginsOptionsClassKey).resolve()).toSet)
+    missingOptions.foreach(missingClass => throw ConfigurationException(f"Missing configuration block for plugin class $missingClass"))
+    val classToOptionsMap = pluginsOptions.map(m => (m(pluginsOptionsClassKey).resolve() -> m.-(pluginsOptionsClassKey))).toMap
+    pluginList.foreach(p => p.plugin.configure(classToOptionsMap(p.className)))
+  }
   // configure SDLPlugin
   Environment.sdlPlugin.foreach(_.configure(pluginOptions))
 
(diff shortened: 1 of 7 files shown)


sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
```diff
@@ -615,6 +615,18 @@ object Environment extends SmartDataLakeLogger {
   }
   private[smartdatalake] var _sdlPlugin: Option[Option[SDLPlugin]] = None
 
+
+  // instantiate sdl plugins if configured. The class names must be separated by a comma. The classname is helpful to map against a configuration list.
+  private[smartdatalake] def sdlPlugins: Option[Seq[SDLPluginWithClassName]] = {
```
zzeekk: Why not just a list, which is empty if there are no plugins?
```suggestion
  private[smartdatalake] def sdlPlugins: Seq[SDLPluginWithClassName] = {
```

sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
```diff
@@ -283,8 +283,10 @@ abstract class SmartDataLakeBuilder extends SmartDataLakeLogger {
   def run(appConfig: SmartDataLakeBuilderConfig): Map[RuntimeEventState, Int] = {
     AppUtil.setSdlbRunLoggerContext(appConfig)
     val stats = try {
-      // invoke SDLPlugin if configured
+      // invoke SDLPlugin if configured. Either sdlPlugin or sdlPlugins is allowed!
+      assert(Environment.sdlPlugin.isEmpty || Environment.sdlPlugins.isEmpty, "Either sdlPlugin or sdlPlugins is allowed, but not both!")
```
zzeekk: Can we combine Environment.sdlPlugin with sdlPlugins list when preparing Environment.sdlPlugins, so we need to handle only one input here, and only one foreach in Line 287 and 333?

sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
```diff
@@ -87,6 +89,7 @@ case class GlobalConfig(kryoClasses: Option[Seq[String]] = None
                         , synchronousStreamingTriggerIntervalSec: Int = 60
                         , environment: Map[String, String] = Map()
                         , pluginOptions: Map[String, StringOrSecret] = Map()
+                        , pluginsOptions: Seq[Map[String, StringOrSecret]] = Seq()
```
zzeekk: I would not add a separate entry pluginsOptions here, but just pass all pluginOptions to all plugins. Keep it simple and stupid.
WDYT?

sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
```diff
@@ -615,6 +615,18 @@ object Environment extends SmartDataLakeLogger {
...
+    if (_sdlPlugins.isEmpty) {
+      _sdlPlugins = Some(EnvironmentUtil.getSdlParameter("pluginClassNames")
+        .map(listString => listString.split(",").map(className => SDLPluginWithClassName(CustomCodeUtil.getClassInstanceByName[SDLPlugin](className.trim), className.trim))))
```
zzeekk:
```suggestion
        .map(listString => listString.split(",").map(_.trim).map(className => SDLPluginWithClassName(CustomCodeUtil.getClassInstanceByName[SDLPlugin](className), className))))
```

zzeekk (approved):

sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
iliassoto: This was to keep the same pattern as sdlPlugin :)

sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
iliassoto: I don't know how much potential we see in the concept of plugins. Are we envisioning many plugins? If so, then I would prefer to have them separated. If we think that this is a narrow feature, I would agree with you. 

Plan B: Marking the pluginOptions as deprecated and in release 3.0.0. only have pluginsOptions. WDYT?

sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
iliassoto: Will merge them depending on the comment about having two options

sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
zzeekk: I think that a Seq[Map[]] is a complicated structure to configure, and as plugins are rather an exception for stuff you really need to inject yearly in an SDLB run, e.g. logging, i vote for keeping it simple, and having a single Map[] with options that gets distributed to all plugins.

sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
iliassoto: Please see https://github.com/smart-data-lake/smart-data-lake/pull/969/commits/7b7cf60085a0e0a0037741ff248f1af75aeb048b

sdl-core/src/main/scala/io/smartdatalake/app/SmartDataLakeBuilder.scala
iliassoto: Please see https://github.com/smart-data-lake/smart-data-lake/pull/969/commits/7b7cf60085a0e0a0037741ff248f1af75aeb048b

sdl-core/src/main/scala/io/smartdatalake/app/GlobalConfig.scala
iliassoto: Done: Please see https://github.com/smart-data-lake/smart-data-lake/pull/969/commits/7b7cf60085a0e0a0037741ff248f1af75aeb048b

sdl-core/src/main/scala/io/smartdatalake/definitions/Environment.scala
```diff
@@ -604,16 +604,24 @@ object Environment extends SmartDataLakeLogger {
...
 
-  // instantiate sdl plugin if configured
-  private[smartdatalake] def sdlPlugin: Option[SDLPlugin] = {
-    if (_sdlPlugin.isEmpty) {
-      _sdlPlugin = Some(EnvironmentUtil.getSdlParameter("pluginClassName")
-        .map(CustomCodeUtil.getClassInstanceByName[SDLPlugin]))
 
+  // instantiate sdl plugins if configured. The class names must be separated by a comma.
+  // initializes with envVars pluginClassName (deprecated) and pluginClassNames
+  private[smartdatalake] def sdlPlugins: Option[Seq[SDLPlugin]] = {
```
zzeekk:
```suggestion
  private[smartdatalake] def sdlPlugins: Seq[SDLPlugin] = {
```
Hier würde ich einfach eine Seq[SDLPlugin] zurückgeben. Dann brauchst du SmartDataLakeBuilder.scala nicht doppelte foreach.
In Zeile 614 solltest du dann _sdlPlugin.get lassen. Das ist an dieser Stelle ja dann immer definiert.

iliassoto: removed the Option Type (see https://github.com/smart-data-lake/smart-data-lake/pull/969/commits/27f10e18549991333a969bfde71c6339a22c18ec) but sdlPlugin doesn't exist anymore, only sdlPlugins (which accepts two env variables).

Issue #976, opened 2025-05-20
Schema is undefined because files are not found


ib31iat:
**Describe the bug**
The result file of the previous step cannot be found and thus the compute feed fails as neither a schema is defined nor are there existing files.

**To Reproduce**
Use the solution files of the Getting started guide, Step 1 (`airports.conf.part-1c-solution`, `btl.conf.part-1b-solution`, `departures.conf.part-1-solution`) and run the `download` and `compute` feed separately.

**Expected behavior**
The compute phase completes without any errors.

**Additional context**
For more details see https://github.com/smart-data-lake/getting-started/issues/38.

ib31iat:
PR #977 is merged.

PR #977, opened 2025-05-21
Fix path filter in `listDataFiles` (#976)

ib31iat:
Use the relative path of the file instead of the absolute path as the Regex pattern for filtering only contains the relative path. 

**Note**: The `hadoopPath` has been removed from the Regex pattern as the method `relativizePath` removes this part of the path. Because only files in the `hadoopPath` are listed, this should not be a problem in my opinion.

To fix the problem in issue #976.


iliassoto (approved):
LGTM after discussion

zzeekk (approved):

PR #981, opened 2025-05-26
test case to illustrate how to use SparkExpressionUtil.evaluate on a …

kaelte:


zzeekk (approved):
Great, thanks @kaelte,

You find more unit tests for the ExpressionEvaluator in spark-extensions repo at https://github.com/smart-data-lake/spark-extensions/blob/develop-spark3.5/src/test/scala/org/apache/spark/sql/custom/ExpressionEvaluatorTest.scala

PR #988, opened 2025-06-05
Add indentation handling for Python blocks (#938)

### LINKED ISSUE #938: python inline indentation
When defining a *PythonCodeDfTransformer* with code directly in HOCON, the multiline inline code currently needs to start right at the beginning of the line. Otherwise we get an indentation error. As a user I want to be able to move the whole python code block right fitting the HOCON transformer indentation. 

We could gather the indentation of the first line and substract this amount from all lines in that code block. This should leave the relative indentation within the code block unchanged, just removing the overall offset.

**Additional context**
example:
now we need to have:

```
    transformers = [{
      type = PythonCodeDfTransformer
      code = """
from pyspark.sql.functions import from_unixtime, date_format, col
transformedDf = inputDf.withColumn("dt",
  date_format(from_unixtime(col("first_seen")), "yyyyMMdd"))
setOutputDf(transformedDf)
      """
    }]
```

but as a user I want to be able to specify as:

```
    transformers = [{
      type = PythonCodeDfTransformer
      code = """
          from pyspark.sql.functions import from_unixtime, date_format, col
          transformedDf = inputDf.withColumn("dt",
            date_format(from_unixtime(col("first_seen")), "yyyyMMdd"))
          setOutputDf(transformedDf)
      """
    }]
```

juliam6:
A function is defined which aims to translate Python's in-built [textwrap.dedent()](https://github.com/python/cpython/blob/main/Lib/textwrap.py) function into Scala. The function allows for indentations in python codeblocks by finding the largest common whitespace, handling both tabs and spaces.

Allows for better indented config files.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/PythonCodeSparkDfTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/PythonCodeSparkDfTransformer.scala
@@ -87,8 +87,8 @@ object PythonCodeDfTransformer extends FromConfigFactory[GenericDfTransformer] {
   def dedent(code: String): String = {
     val lines = code.stripMargin.linesIterator.toList
     val nonEmptyLines = lines.filter(line => line.trim.nonEmpty)
-    val commonIndent = if (nonEmptyLines.isEmpty) "" else nonEmptyLines.map(_.takeWhile(c => c == ' ' || c == '\t')).min
-    val dedentedLines = lines.map(_.drop(commonIndent.length))
+    val minIndentLength = if (nonEmptyLines.isEmpty) 0 else nonEmptyLines.map(_.prefixLength(c => c == ' ' || c == '\t')).min
+    val dedentedLines = lines.map(_.drop(minIndentLength))
     dedentedLines.mkString(System.lineSeparator())
   }
 }
--- a/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/PythonCodeSparkDfTransformer.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/workflow/action/spark/transformer/PythonCodeSparkDfTransformer.scala
@@ -87,22 +87,8 @@ object PythonCodeDfTransformer extends FromConfigFactory[GenericDfTransformer] {
   def dedent(code: String): String = {
     val lines = code.stripMargin.linesIterator.toList
     val nonEmptyLines = lines.filter(line => line.trim.nonEmpty)
-    val indents = nonEmptyLines.map(_.takeWhile(c => c == ' ' || c == '\t'))
-    val commonIndent: String = if (indents.isEmpty) {
-      ""
-    } else {
-      val firstIndent = indents.head
-      indents.tail.foldLeft(firstIndent) { (common, next) =>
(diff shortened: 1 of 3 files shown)


iliassoto:
Nice @juliam6 . I simplified the dedent function a bit: instead of comparing each indent per line, we just take the smallest one. This way we can also identify the common indent even if it's in a different order but with same characters. 

We still must check how indents are read. If they are interpreted as two-spaces we can leave it as-is. Otherwise, neither of our functions handles the case where two lines have the same intentation but using different chars: (e.g. ("\t\t") vs. ("  \t").

Once we checked this we can write some tests.

juliam6:
@iliassoto Thank you for your code review. I investigated the space/tab thing. In Intellij by default, tabs are converted to two spaces. You can change this setting at the bottom right > "Tab" > "configure Indents for HOCON" > "Use tab characters". Also apply the changes by clicking "Tab" > "Reindent File with tabs".

Python's dedent function considers both tabs and spaces as whitespace, but they are not equal: the lines "  hello" and "\\thello" are considered to have no common leading whitespace. So in your example (e.g. ("\t\t") vs. (" \t")) should have no common whitespace. That means the dedent function should do nothing, and an indentation error should be thrown.

The way our simplified function is designed currently, it will work fine as long as there is the same number of whitespace characters per line. As soon as there are uneven characters per line, for example something like this:

	val input = """
 	line B // space + tab
 	line C // space + tab
	line A // tab
    """
"\r\n \tline A\r\n \tline B\r\n\tline C\r\n"

It will take the lexigraphical minimum (/t) as "common indent", and remove the length of it (1), which results in an indentation error.

imo we can't get around character by character comparison (like in python's function).

I see some other minor issue, but If we dont care about getting this level of detail then I think this function is fine, and I would just change these two lines to simplify further:
`val minIndentLength = nonEmptyLines.map(_.prefixLength(c => c == ' ' || c == '\t')).min`
`val dedentedLines = lines.map(_.drop(minIndentLength))`

iliassoto:
@juliam6 As discussed internally, we have two options: 
1. We assume that the user uses only tabs or only spaces and not a weird mix in their code. In this case we can leave our function as is.
2. Parting from the first Python line, we see which combination of spaces and tabs is a common indent. We then remove this common indent from each line by keeping a list of removed spaces and removed tabs. This ensures that indents with different orderings are considered equal ("<space>\t\t" == "\t<space>\t" == "\t\t<space>")

iliassoto:
As discussed: 
`val minIndentLength = nonEmptyLines.map(_.prefixLength(c => c == ' ' || c == '\t')).min` 
 requires an 'isEmpty' check, as `List().min` will throw an error. 
 
 `val minIndentLength = if (nonEmptyLines.isEmpty) 0 else nonEmptyLines.map(_.prefixLength(c => c == ' ' || c == '\t')).min`

juliam6:
Hi @iliassoto 

I think we should go with assumption 1) that users use only spaces or tabs (or a mix of spaces and tabs consistently). This keeps the code simple. The worst case scenario results in our function breaking alignment, which would typically surfaces as `TabError: inconsistent use of tabs and spaces in indentation` or as a general IndentationError.

I have implemented further simplification as discussed. Let's do the test cases next!

iliassoto (approved):
@juliam6 let's merge

PR #990, opened 2025-06-16
Improve code documentation parsing (#989)

### LINKED ISSUE #989: Improve SchemaViewer Readability
The tab on the right-hand side of the SchemaViewer containing the definition is hard to read due to unnecessary line breaks and bad formatting stemming from the documentation in the code.

Improve the parsing from the code documentation. 

**Additional context**
https://github.com/smart-data-lake/schema-viewer/issues/27

ib31iat:
- Remove double square brackets around Link elements
  - Retain hyperrefs to keep them as an <a> Tag in the SchemaViewer
  - Fix spaces around the Link elements
  - Handle the plural `s` correctly
- Filter out `OtherTag` instead of just `ParamTag`
- Add Github URL to scaladoc import
- Replace `->` with `\u2192`
- Improve conversion from wiki code block (`{{{...}}}`) to Markdown code block

It is hard to read the documentation in the SchemaViewer.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/misc/ScaladocUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/misc/ScaladocUtil.scala
@@ -59,30 +59,61 @@ private[smartdatalake] object ScaladocUtil {
     }
   }
 
-  def formatScaladocStringLinkTag(capture_group1: String, capture_group2: String): String = {
+  // Remove leading spaces in code blocks
+  def dedentCodeBlock(code: String): String = {
+    val lines = code.stripMargin.linesIterator.toList
+    val nonEmptyLines = lines.filter(line => line.trim.nonEmpty && ! List("{{{", "}}}").contains(line.trim))
+    val firstLineIndentation = if (nonEmptyLines.isEmpty) (0, 0) else spacesAndTabs(nonEmptyLines.head) // Assume nicely formatted code
+    val dedentedLines = lines.map(removeSpacesAndTabs(_, firstLineIndentation))
+    dedentedLines.mkString("\n")
+  }
+
+  // Leading number of (spaces, tabs)
+  def spacesAndTabs(line: String): (Int, Int) = {
+    line.takeWhile(Seq(' ', '\t').contains(_)).foldLeft((0, 0))((spacesTabs, char) => {
+      if (char == ' ') (spacesTabs._1 + 1, spacesTabs._2) else (spacesTabs._1, spacesTabs._2 + 1)
+    })
+  }
+
+  def removeSpacesAndTabs(line: String, spacesTabs: (Int, Int)): String = {
+    require(spacesTabs._1 >= 0 && spacesTabs._2 >= 0, "Indentation error. The line has either too many spaces or too many tabs")
+    if (line.isEmpty) ""
+    else if (List("{{{","}}}").contains(line.trim)) line.trim
+    else line.head match {
+      case c if ((0, 0) == spacesTabs) => line
+      case ' ' => removeSpacesAndTabs(line.tail, (spacesTabs._1 - 1, spacesTabs._2))
+      case '\t' => removeSpacesAndTabs(line.tail, (spacesTabs._1, spacesTabs._2 - 1))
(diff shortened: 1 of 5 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/misc/ScaladocUtil.scala
```diff
@@ -59,9 +59,33 @@ private[smartdatalake] object ScaladocUtil {
...
   }
 
+  def formatScaladocStringLinkTag(capture_group1: String, capture_group2: String): String = {
+    // Do not wrap urls in inline code blocks
+    if (capture_group1.contains("https://")){
+      val split_hyperref = capture_group1.split(" ")
+      if (split_hyperref.length > 1){
+        // If the Url contained an alias (pretty name), preserve it
+        s" [${split_hyperref.drop(1).mkString(" ")}](${split_hyperref(0)}) "
+      }else{
```
zzeekk, line 99:
```suggestion
      } else {
```

sdl-core/src/main/scala/io/smartdatalake/util/misc/ScaladocUtil.scala
```diff
@@ -59,9 +59,33 @@ private[smartdatalake] object ScaladocUtil {
     }
   }
 
+  def formatScaladocStringLinkTag(capture_group1: String, capture_group2: String): String = {
+    // Do not wrap urls in inline code blocks
```
zzeekk, line 93: Can we make a unit test for this?

zzeekk (commented):
Works fine for me, looks already much better in schema viewer!

Some more formatting improvements, maybe easy to implement:

If "->" is at the beginning of a line, create a line break before it (Example: SFtpFileRefDataObject)

How to create more line-breaks for text without beeing to excessiv... e.g. SFtpFileRefDataObject.partitionLayout, or SqlDfTransformer
-> make a simple linebreak if there is a Dot at the end of a line?
-> make an additional line break (e.g. new paragraph) when the line is smaller than 30 lines and there is a Dot at the end of a line?

Keep indendation for code examples, e.g. global.uiBackend?

sdl-core/src/main/scala/io/smartdatalake/util/misc/ScaladocUtil.scala
ib31iat: Concerning the line breaks, I think that is not that easy to make a case distinction based on the line length. Considering your example of the `SFtpFileRefDataObject.partitionLayout`:
```
 * @param partitionLayout partition layout defines how partition values can be extracted from the path.
 *                        Use "%<colname>%" as token to extract the value for a partition column.
 *                        As partition layout extracts partition from the path of individual files, it can also be used to extract partitions from the file name.
 *                        With "%<colname:regex>%" a regex can be given to limit search. This is especially useful
 *                        if there is no char to delimit the last token from the rest of the path or also between
 *                        two tokens.
 *                        Be careful that for directory based partition values extraction, the final path separator must be part
 *                        of the partition layout to extract the last token correctly, e.g. "%year%/" for partitioning with yearly directories.
```

In this case there would be a new paragraph after `two tokens` but it belongs to a longer sentence, so there shouldn't be one if I understand your proposal correctly. Splitting the stringinto sentences using full stops is also not trivial (due to e.g. and so on).
My proposal is to have carriage returns (.  \n) whenever there is a full stop at then end of a line and have no distinction based on line length for paragraphs. Instead there will be a new paragraph if there is one anyways caused by the formatting of the documentation or the parsing.

zzeekk: Yes, your proposal makes sense as.

ib31iat:
As discussed, here a list of the things that should be looked at in the documentation:
- Replace lists that are created with `->`, with `-`
- End lines with a full stop (`.`) for correct and pretty rendering in SchemaViewer (markdown)

It is possible that we find other things along the way. But those are the ones that I noticed and we should maybe add to a new issue?

zzeekk (approved):
Very nice, thanks!

PR #1013, opened 2025-10-01
Refactor to use scala types instead of ScalaDataTypeEnum

zzeekk:
Hi @iliassoto,
I did some refactoring to use scala types directly instead of ScalaDataTypeEnum, and some more cleanups and improvements. Please have e look.

Next step is probably to have column references and literals supported, e.g. col("abc") and lit(1). In my opinion, for this we would need a super class to ScalaColumn, e.g. ScalaExpression, which has no data as it is not yet bound to DataFrame when instantiated.
When added to a DataFrame, e.g. by select or withColumn, it gets calculated and becomes a normal ScalaColumn.

This would also allow us to implement aggregate expressions later.

WDYT?


zzeekk:
Note: as i renamed ScalaDataframe to ScalaDataFrame, there is no diff for this file :-(

zzeekk:
Added a draft for expressions in a 2nd commit.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/plainScala/ScalaGroupedDataframe.scala
```diff
@@ -19,26 +19,19 @@
...
 import scala.reflect.runtime.universe
 
 
-case class ScalaGroupedDataframe(keyColNames: Seq[String], groups: Map[Seq[Any], Seq[Int]], df: ScalaDataframe) extends GenericGroupedDataFrame {
+case class ScalaGroupedDataframe(keyColNames: Seq[String], df: ScalaDataFrame) extends GenericGroupedDataFrame {
 
-  override def agg(columns: Seq[GenericColumn]): GenericDataFrame = throw new NotImplementedError("For aggregations please use another signature providing an aggregation expression")
+  override def agg(columns: Seq[GenericColumn]): GenericDataFrame = {
+    DataFrameSubFeed.assertCorrectSubFeedType(subFeedType, columns)
+    val scalaCols = columns.map(_.asInstanceOf[ScalaColumn[_]])
```
iliassoto, line 32: @zzeekk Why did you add these two lines if we're throwing an Error anyway?

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/plainScala/ScalaSchema.scala
```diff
@@ -36,55 +36,42 @@ case class ScalaSchema(_fields: List[ScalaColumnDefinition], isInferred: Boolean
...
-  override def toLowerCase: ScalaSchema = ScalaSchema(_fields.map(_.makeNullable))
+  override def toLowerCase: ScalaSchema = copy(fields = fields.map(_.makeNullable))
 
-  override def removeMetadata: ScalaSchema = ScalaSchema(_fields.map(_.removeMetadata))
+  override def removeMetadata: ScalaSchema = copy(fields = fields.map(_.removeMetadata))
 
-  override def getEmptyDataFrame(dataObjectId: SdlConfigObject.DataObjectId)(implicit context: ActionPipelineContext): GenericDataFrame = ???
+  override def getEmptyDataFrame(dataObjectId: SdlConfigObject.DataObjectId)(implicit context: ActionPipelineContext): GenericDataFrame = {
+    // TODO
+    null
```
iliassoto, line 70: Better null than error?

iliassoto (approved):
Hi @zzeekk , thank you for the changes. The architecture with the ScalaTypes is way more elegant of course!

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/plainScala/ScalaGroupedDataframe.scala
zzeekk: we will need these once when we implement it.

sdl-core/src/main/scala/io/smartdatalake/workflow/dataframe/plainScala/ScalaSchema.scala
zzeekk: We need to add an implementation soon for the unit tests i think. Lets continue this on your branch.

Issue #1020, opened 2025-10-26
Unchecked access to optional attribute  ODataDataObject#responseBufferSetup leads to Exception


qnob:
**Describe the bug**
An NoSuchElementException is thrown when responseBufferSetup isn't declared in Hocon config. However, the attribute is declared as optional.

**To Reproduce**
Run  SDLB with follow config:

dataObjects {

  ext-users {
    type = ODataDataObject
    schema = "ddl#Id String, Name String, Upn String, TenantId String, OrganizationUnitId String"
    tableName = "Users"
    baseUrl = "https://api.ch-02.luware.cloud/v1/odata/"
    authorization {
      type = TokenAuthMode
      token = "###ENV#ACCESS_TOKEN###"
    }
    //incrementalOutputExpr = "modifiedon"
    nRetry = 0
//    responseBufferSetup {
//      tempFileDirectoryPath = "tmp/"
//      memoryToFileSwitchThresholdNumOfChars = 0
//    }
    timeouts {
      connectionTimeoutMs = 200000
      readTimeoutMs = 200000
    }
  }

**Expected behavior**
No exception is thrown. InMemory Buffer is used only

**Additional context**
2025-10-26 18:46:31 INFO  ActionDAGRun$ActionEventListener - Action~nimbus-transfer-users[CopyAction]: Init succeeded [dag-1-66]
2025-10-26 18:46:31 INFO  ActionDAGRun$ActionEventListener - Action~nimbus-transfer-users[CopyAction]: Exec started [dag-1-67]
2025-10-26 18:46:31 INFO  CopyAction - (Action~nimbus-transfer-users) enrichSubFeedDataFrame: getting DataFrame for DataObject~ext-users [exec-nimbus-transfer-users]
2025-10-26 18:46:31 WARN  ActionDAGRun$ActionEventListener - Action~nimbus-transfer-users[CopyAction]: Exec failed with: Exception: NoSuchElementException: None.get [dag-1-67]
2025-10-26 18:46:31 ERROR ActionDAGRun - Exec: TaskFailedException: Task nimbus-transfer-users failed. Root cause is 'NoSuchElementException: None.get' [main]
2025-10-26 18:46:31 INFO  ActionDAGRun$ - Exec FAILED for transfer-nimbus runId=1 attemptId=1:
                  ┌─────┐                   
                  │start│                   
                  └───┬─┘                   
                      │                     
                      v                     
 ┌─────────────────────────────────────────┐
[...]

PR #1021, opened 2025-10-27
#1020: handle case where bufferSetup isn't defined correctly

qnob:
#1020 :handle case where bufferSetup isn't defined correctly

Documentation and attribute declaration isn't align with the actual behavior of the component.


PMuellerELCA (approved):
Reviewed

Issue #1047, opened 2026-02-08
Map input/output DataFrame names of ScalaClassSparkDfsTransformer


zzeekk:
**Is your feature request related to a problem? Please describe.**
Transformers with multiple Input/Output DataFrames expect specific names for the input DataFrames, respectively give specific names to the output DataFrames. This makes it difficult to create generic Transformers with multiple Input/Output DataFrames.

This is no problem for single DataFrame Transformers, as there is only one unnamed input/output.

**Describe the solution you'd like**
Add additional parameters to ScalaClassSparkDfsTransformer to map generic input/outputIds to other names.
Add additional parameter to override outputId for ScalaClassSparkDfsTransformer returns only one DataFrame (currently it is just named after the Actions output DataObject, if the Actions has only one output DataObject).

PR #1048, opened 2026-02-09
configure input/output DataFrame mapping for ScalaClassSparkDfsTransformer (#1047)

zzeekk:
see #1047


kaelte (changes requested):
Changes look good but branch cannot be build.
Even after merging fixed develop-spark3 -> feature/1047-input-output-dataframe-names

kaelte (approved):

PR #1050, opened 2026-02-17
DsComment.transformCommentCols to add comments to transformed columns

kaelte:
Method to comment transformed columns of a DataSet

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/CommentedColumn.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/CommentedColumn.scala
@@ -22,6 +22,13 @@ package io.smartdatalake.util.spark.dataset
 import org.apache.spark.sql.Column
 
 /**
+ * CommentedColumn is part of return type of DsComment.transformCommentCols
+ * That method allows to not only transform columns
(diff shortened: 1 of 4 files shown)


sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/CommentedColumn.scala
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.smartdatalake.util.spark.dataset
+
+import org.apache.spark.sql.Column
+
+/**
+ *
```
zzeekk, line 32: Can you add a short description, and what it can be used for?

zzeekk (approved):

sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/CommentedColumn.scala
kaelte: Done: 5c9b0f5f4ab118a3d2e943051e95f79ba1f561d5

Issue #972, opened 2025-05-09
Rewrite core tests to use ScalaMockDataObject


iliassoto:
E2E tests should use the new ScalaMockDataObject instead of the (Spark-based) MockDataObject.

iliassoto:
The following test files need to be updated for the core module:

- DebugTransformerTest
- SmartDataLakeBuilderTest
- UniqueKeyExpectationTest
- ActionDAGTest
- CopyActionTest
- CopyActionScalaSubFeedTest
- CustomDataFrameActionTest
- DeduplicateWithMergeActionTest
- OfflineCopyActionTest
- CustomDfsTransformerTest
- PKViolatorsDataObjectTest
- UniqueKeyExpectationTest
- ValidateOnReadTest

The most critical one is *SmartDataLakeBuilderTest*, as it validates the base abstract class *SmartDataLakeBuilder* containing the main app logic. *SmartDataLakeBuilder* itself has several Spark dependencies, but creating another parent class is not an option as most of the logic is already baked into the methods *run*, *exec*, etc. The same goes for the overarching *ActionPipelineContext*. 

A way of handling this would be to create a *SubFeedProvider* as well as an *ActionPipelineContextProvider* that define which engine is being used for the pipeline. 

WDYT @zzeekk ?

iliassoto:
@pgruetter what do you think? :)

zzeekk:
Hi @iliassoto,
Interesting question, especially how we handle SparkSession in GlobalConfig and SmartDataLakeBuilder.startSimulation.

The fundamental question might be if we want to have just one Engine/Session per SDLB job/process, e.g. Spark, or DuckDB, or Spark Connect, or if we allow running pipelines using different engines in the same SDLB job/process.

I would try if we can go for the second, running multiple engines in the same job.
One idea could be to move getting the Session to the Connection. We have that already for Snowflake, e.g. SnowflakeConnection.getSnowparkSession. We could have a trait "SparkSessionProvider", that is implemented by all "Connections" that work with Spark.

Removing it from SmartDataLakeBuilder.startSimulation is not so difficult, as we pass the SparkSession mainly to get the Hadoop Configuration. We cannot get rid of Hadoop in the core, it is SDLBs interface to the filesystem. But instead of passing the SparkSession to startSimulation, we can pass directly the Hadoop Configuration.

[...]

Issue #1042, opened 2026-01-28
Support nullable PKs in HistorizeAction


qnob:
## Is your feature request related to a problem? Please describe.

When running a `HistoryAction` in SDLB, every load results in `INSERT`-only behavior, even when nullable primary key fields remain unchanged and the `dl_hash` value is identical.  This causes duplicate history records to be written during every run.

In my example. the target table does not have a technical primary key, so all business-relevant columns were defined as primary key fields for historization. And some of them contain `null`!

The Databricks history log shows the following generated merge predicate:

`hst_table_cd#2790 = hst_table_cd#3066 AND
hst_pk_value#2791 = hst_pk_value#3067 AND
hst_pk_value_2#2792 = hst_pk_value_2#3068 AND
hst_pk_value_3#2793 = hst_pk_value_3#3069 AND
hst_pk_value_4#2794 = hst_pk_value_4#3070 AND
hst_field_name#2795 = hst_field_name#3071 AND
hst_mut_dat#2796 = hst_mut_dat#3072 AND
hst_old_value#2797 = hst_old_value#3074 AND
hst_new_value#2798 = hst_new_value#3075 AND
dl_ts_captured#2876 = dl_ts_captured#3077`

Following issue arise here:

Several primary key fields (e.g., `hst_pk_value_2`, `hst_pk_value_3`, `hst_pk_value_4`) are nullable  
Especially `hst_pk_value_4`, which is *always* NULL.  
A comparison using `=` results in `NULL = NULL` → `NULL`, thus the merge condition does not match, even when the records are logically identical.

Example of repeated records:

| hst_table_cd | hst_pk_value | hst_field_name | hst_mut_dat | hst_mut_uid | hst_old_value | hst_new_value | dl_hash     | dl_ts_captured                |
|--------------|---------------|----------------|-------------|-------------|---------------|---------------|-------------|-------------------------------|
| DFK          | 23            | DFK_FKT_CD     | 2020-05-13T13:26:28.773 | 57 | 0 | 12 | 1177443906 | 2026-01-19T14:52:37.9        |
| DFK          | 23            | DFK_FKT_CD     | 2020-05-13T13:26:28.773 | 57 | 0 | 12 | 1177443906 | 2026-01-26T10:58:36.776718   |
| DFK          | 23            | DFK_FKT_CD     | 2020-05-13T13:26:28.773 | 57 | 0 | 12 | 1177443906 | 2026-01-27T08:33:58.51713    |

[...]

PR #1062, opened 2026-03-22
fix incremental historize null aware hashing

zzeekk:
Implement incremental historize with null aware hashing, and fix condition to add dl_hash column when starting with incremental historize on a traditionally historized table.

This didn't popup in unit tests up to now because initial dl_hash column was "accidentially" overwritten, and so the test succeeded, but was actionally not testing properly what happens with "null" values. Spark hash function is ignoring null values. If a column with null values is added to a DataFrame, these records have the same hash value as the record without null value. This is intended by spark as it somehow is the ternary logic of SQL. See also https://stackoverflow.com/questions/72026028/how-to-create-unique-hash-for-each-row-having-null-value-in-column-in-spark-scal for an documented example

When implementing #972 i came across some strange behavior of unit tests, investigated it and found that the current behaviour is not correct...


qnob:
Hi @zzeekk 

Is this issue related to https://github.com/smart-data-lake/smart-data-lake/issues/1042

Issue #1086, opened 2026-06-11
Substitution in Hocon replacer


mand35:
**Is your feature request related to a problem? Please describe.**
As a user I want to specify my dataObjects with - seperator. I also want to use the replacer like ~{id} for table names. 
Table names typically need to be specified without -, better with _.

**Describe the solution you'd like**
A replacer with substitution.

mand35:
merged, works as expected

PR #1088, opened 2026-06-14
config parser replacer for correct table names

mand35:
Allow transforming a resolved ~{path} value with an optional pipe modifier, e.g. ~{id|snake}. The snake modifier replaces dashes with underscores, so an element id like 'int-airports' can be reused as a table name 'int_airports' where dashes are not permitted.

Unknown modifiers fail with a ConfigurationException; the plain ~{path} form is unchanged.

What changes are included in the pull request?
Solves https://github.com/smart-data-lake/smart-data-lake/issues/1086
For the path replacer a substitution can be specified. So far only snake, but more can be added. Any further ideas?

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/config/ConfigParser.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/config/ConfigParser.scala
@@ -271,8 +271,10 @@ private[smartdatalake] object ConfigParser extends SmartDataLakeLogger {
   }
 
   /**
-   * Substitutes parts inside values by other paths of the configuration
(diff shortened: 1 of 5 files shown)


zzeekk:
Can we also convert camelCase to snake_case?
I already had some discussions that using camelCase for IDs would be more natural, so we should support both for substitution to e.g. table names.

zzeekk (approved):
Good improvement!

mand35:
@zzeekk I modified. So now the substitution does not only replace - to _, but also camelCase to camel_case. Any further substitution you have in mind, we could add directly?

PR #1090, opened 2026-06-16
cofigParser: replacer now with chained modifier, and noPrefix operator

mand35:
Generalize the local config substitution (~{path}) to support chaining multiple modifiers applied left-to-right, e.g. ~{id|noPrefix|snake}. Previously only a single modifier (snake) was supported.

Adds a new noPrefix modifier that removes the first part of a value, separated by a dash (-), underscore (_) or camelCase boundary:
  - int-airports → airports
  - int_airports → airports
  - intAirports → Airports
  
Includes tests for noPrefix (each separator type, earliest-separator selection, no-separator passthrough), for both chaining orders, and for unknown modifiers in a chain. Scaladoc and the "unknown modifier" error message updated accordingly.

Why are the changes needed?

The snake modifier (#1086) only handled one transform per token. Reusing an element id as a downstream name often needs more than one step — e.g. stripping a stage/layer prefix and converting to snake_case. noPrefix covers the common case of dropping a
  prefix (like int-/stage-) from an id, and chaining lets these be combined in a single substitution.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/config/ConfigParser.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/config/ConfigParser.scala
@@ -284,12 +284,36 @@ private[smartdatalake] object ConfigParser extends SmartDataLakeLogger {
   }
 
   /**
+   * Removes the first part of a value, where parts are separated by a dash ('-'), an underscore ('_')
+   * or a camelCase boundary (a lowercase letter or digit followed by an uppercase letter).
+   * The earliest separator in the value determines the prefix to remove; for dash/underscore the separator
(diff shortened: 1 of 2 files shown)


zzeekk (approved):

pgruetter:
@mand35 @zzeekk 
Looks good, nice addition 👍 

Only thing I'm wondering is with camelCase if the first letter should be converted to lowercase. 
so for intAirport -> airport instead of Airport. 
WDYT ?

mand35:
@pgruetter 
let's answer the different cases for intAirports:
~{id|noPrefix} → Airports 
~{id|noPrefix|snake} → airports
~{id|snake} → int_airports

PR #1094, opened 2026-06-29
fix snake_case normalization for column names with mixed separators

mand35:
## Problem

`strCamelCase2LowerCaseWithUnderscores` (used by `StandardizeColNamesTransformer`) produced
double underscores and trailing underscores when column names contained special chars alongside
uppercase word boundaries. Three failure modes, all with `replaceNonStandardSQLNameCharsWithUnderscores=true`:

| Input | Before | Expected |
|---|---|---|
| `Invoice-No` | `invoice__no` | `invoice_no` |
| `InvoiceName.ID` | `invoice_name__id` | `invoice_name_id` |
| `SO item` | `so__item` | `so_item` |
| `value of property!in$$` | `value_of_property_in_` | `value_of_property_in` |

## Root cause

The regex tail `[^A-Z_]*` greedily consumed any non-uppercase, non-underscore character
(hyphens, dots, spaces) into the camelCase token. Those chars then survived into the joined
result, where `replaceNonSqlWithUnderscores` turned them into a second `_` adjacent to the
mkString separator. Trailing special chars in the last token produced a trailing `_` for the
same reason.

## Fix

Two changes to `strCamelCase2LowerCaseWithUnderscores` (`StringUtil.scala`):

1. **Narrow the regex tail** from `[^A-Z_]*` to `[a-z0-9]*` — special chars between camelCase
   words are no longer absorbed into the token; they become their own single-char matches under
   the second alternative `[^A-Z_]+`.

2. **Strip boundary special chars from each token** via
   `.replaceAll("^[^a-z0-9]+|[^a-z0-9]+$", "")` — pure-delimiter tokens (lone space, hyphen,
   dot) collapse to empty and are dropped; mixed tokens like `" item"` or `"value...$$"` have
   their leading/trailing noise removed before joining.

All-lowercase tokens such as `"one dot"` or `"two-do-ts"` match as a single chunk under the
second alternative and keep their internal special chars intact for downstream processing by
`replaceNonSqlWithUnderscores`.

## Tests

Extended `StringUtilTest` and `StandardizeColNamesTransformerTest` to cover:
- special chars between camelCase words (`Invoice-No`, `Invoice.No`, `Invoice!No`)
- lowercase word after acronym separated by space (`SO item`, `Sales Order SO item`)
- trailing special chars (`Invoice No `, `SO item `)
- updated `value_of_property_in_` expectation to `value_of_property_in` (trailing `_` removed)

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/misc/StringUtil.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/misc/StringUtil.scala
@@ -33,7 +33,7 @@ object StringUtil {
    * @return transformed [[String]]
    */
   def strCamelCase2LowerCaseWithUnderscores(x: String): String = {
-    val normalized = "([A-Z]+[a-z0-9]*)|[^A-Z_]+".r.findAllMatchIn(x)
+    val normalized = "([A-Z]+[^A-Z_]*)|[^A-Z_]+".r.findAllMatchIn(x)
       .map(_.group(0).toLowerCase.replaceAll("^[^a-z0-9]+|[^a-z0-9]+$", ""))
       .filter(_.nonEmpty).mkString("_")
     // preserve leading underscores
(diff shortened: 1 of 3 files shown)


kaelte (approved):

kaelte (changes requested):
test case fails

mand35:
@kaelte Now tests run thru successful

kaelte (approved):
Now you may merge :)

zzeekk (approved):

PR #1098, opened 2026-07-01
provision referential keys (PK/FK) to catalog on write (#1097)

### LINKED ISSUE #1097: primary and foreign keys into unity catalog
  `DeltaLakeTableDataObject` accepts `table.primaryKey` and `table.foreignKeys` in its HOCON configuration, but these constraints are never written to the underlying Databricks Unity Catalog table. As a result, the table relationships defined in the SDLB      
  config are invisible in the Databricks UI and to any Unity Catalog governance tooling — they exist only in SDLB's own metadata store and lineage graph.                                                                                                           
                                                                                                                                                                                                                                                                    
  When SDLB creates or updates a Delta table registered in Unity Catalog, it should also apply the declared primary key and foreign key constraints using Unity Catalog's informational constraint DDL:                                                             
                                                                                                                                                                                                                                                                    
  ```sql                                                                                                                                                                                                                                                            
  ALTER TABLE <catalog>.<schema>.<table>                                                                                                                                                                                                                            
    ADD CONSTRAINT <name> PRIMARY KEY (<cols>);                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                    
  ALTER TABLE <catalog>.<schema>.<table>                                                                                                                                                                                                                            
    ADD CONSTRAINT <name>                                                                                                                                                                                                                                           
    FOREIGN KEY (<local_cols>)                                                                                                                                                                                                                                      
    REFERENCES <ref_schema>.<ref_table> (<ref_cols>);                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                    
  Unity Catalog constraints are informational (not enforced at write time), so applying them carries no performance cost and introduces no risk of breaking existing pipelines. The constraints should be reconciled on each run (drop + re-add if the definition   
  changed, no-op if unchanged) to stay in sync with the SDLB config.                                                                                                                                                                                                
                                                                                                                                                                                                                                                                    
**Additional context**
  - Unity Catalog primary/foreign key constraint support: https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-alter-table.html                                                                                                                    
  - The ForeignKey case class currently holds db, table, columns (Map), and an optional name — all fields needed to generate the ALTER TABLE … ADD CONSTRAINT DDL are already present.                                                                              
  - This feature is particularly valuable for star-schema gold layers (fact + dimension tables) where the FK graph represents the intended join topology and should be surfaced in Unity Catalog's data lineage and Explorer views.

mand35:
Extends SDLB's catalog metadata provisioning (which already pushes table/column descriptions) to also provision primary key and foreign key constraints to the target catalog on write.                                                                           
                                                                                                                                                                                                                                                                    
  Table model changes                                                                                                                                                                                                                                               
  - createAndReplacePrimaryKey renamed to createAndReplaceReferentialKeys — one flag now controls both PK and FK provisioning                                                                                                                                       
  - ForeignKey gains a catalog field for Unity Catalog 3-part names (catalog.schema.table)                                                                                                                                                                          
                                                                                                                                                                                                                                                                    
  Core trait (CanHandleReferentialKeys, replaces CanHandleConstraints (constrains is an overloaded term referring to data quality))                                                                                                                                                                                              
  - Adds ForeignKeyDefinition case class                                                                                                                                                                                                                            
  - Adds ensureColumnsNotNull — sets PK columns NOT NULL before applying the constraint                                                                                                                                                                             
  - Adds three FK abstract methods (getExistingFKConstraints, dropForeignKeyConstraint, createForeignKeyConstraint)                                                                                                                                                 
  - Adds unified createOrReplaceReferentialKeys entry-point that handles both PK and FK in one call; FK orchestration compares defined vs. existing constraints and drops/recreates on change                                                                       
                                                                                                                                                                                                                                                                    
  New SparkCatalogReferentialKeys trait                                                                                                                                                                                                                             
  Shared Spark SQL implementation (Unity Catalog INFORMATION_SCHEMA queries + ALTER TABLE DDL with RELY) used by both DeltaLakeTableDataObject and IcebergTableDataObject, avoiding code duplication.                                                               
                                                                                                                                                                                                                                                                    
  DataObject changes                                                                                                                                                                                                                                                
  - DeltaLakeTableDataObject — replaces inline PK methods with the shared mixin; single postWrite call                                                                                                                                                              
  - IcebergTableDataObject — new UC-gated PK/FK support; warns and skips on non-Databricks catalogs                                                                                                                                                                 
  - JdbcTableDataObject — PK provisioning moved from prepare() to postWrite() to align with FK (referenced tables may not exist at prepare time); full FK support via JdbcCatalog / JdbcTableConnection                                                             
  - SnowflakeTableDataObject — FK support added via SnowflakeConnection                                                                                                                                                                                             
                                                                                                                                                                                                                                                                    
  Why are the changes needed?                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                    
  The ForeignKey case class and table.foreignKeys config have existed since early versions but were never consumed at runtime — FK definitions were purely informational. Users with Unity Catalog on Databricks or Snowflake expect FK relationships to be visible 
  in the catalog for lineage, documentation, and query optimisation hints, just as table descriptions already are.                                                                                                                                                  
                                                                                                                                                                                                                                                                    
  The PK behaviour is also improved: previously a PK constraint could fail silently if the target columns were still marked nullable in the catalog schema. The new ensureColumnsNotNull step corrects that before applying the constraint.                                         

Closes #1097


zzeekk (approved):
Great.

kaelte (approved):

PR #1099, opened 2026-07-01
quality: Add fluent Column extensions for comments and non-nullability

mand35:
  Adds a new `DsColComment` implicit class to `Quality` (`sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/Quality.scala`) that extends `Column` with two fluent methods:                                                                                
  - `col("abc").withComment("my description")` — shorthand for the existing `Quality.withComment(column, commentText)`.                                                                                                                                             
  - `col("abc").makeNotNullable` — marks a column as not-nullable. Implemented via Spark's `AssertNotNull` expression so that the schema's `nullable` flag is actually set to `false` and a runtime exception is thrown if a null value is encountered.             
                                                                                                                                                                                                                                                                    
  Added tests in `DsCommentTest.scala` covering the fluent `withComment` call, the schema nullability change, and the runtime null-check.                                                                                                                           
                                                                                                                                                                                                                                                                    
  `Quality` already supports setting column comments, but only via standalone functions (e.g. `withComment("abc", "desc")`) or dataset-level helpers. This adds a more ergonomic, chainable API directly on `Column`, and extends it to also cover changing column  
  nullability — which wasn't previously supported at all.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/Quality.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/Quality.scala
@@ -22,6 +22,7 @@ package io.smartdatalake.util.spark.dataset
 import io.smartdatalake.util.LogUtils.{debLogFun, debugLog}
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.expressions.NamedExpression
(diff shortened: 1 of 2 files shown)


zzeekk (approved):
Very nice :-)

kaelte (approved):

Issue #973, opened 2025-05-09
Separation of sdl-core and sdl-spark Modules


iliassoto:
The core module should not include the Spark-dependencies anymore, but only include app-functionality (engine agnostic). 

- Some core functionality (such as SQL-Expressions) still depend on Spark --> Find an alternative?
- Create new sdl-spark module that includes Spark components
- In the first step, all DataObjects that only work with Spark shold be part of the module as well. A further separation can be discussed in the future.

zzeekk:
Notes for refactoring:
- ExpectationValidation:177 should not be tied to SparkColumn (col.inner.expr.sql is not nice), e.g. expectation.getValidationErrorColumn in Line 170 should not return a Spark Column, but a validator function that directly returns a errorMsg String?

zzeekk:
misc.SchemaUtil still contains many methods that use Spark Schema types, e.g. StructType.

PR #1102, opened 2026-07-05
Move Spark-specific code from sdl-core to sdl-spark module (issue #973)

zzeekk:
Moving all Spark-specific code from sdl-core to sdl-spark (issue https://github.com/smart-data-lake/smart-data-lake/issues/973) so sdl-core compiles without Spark

see #973


kaelte (approved):
We should ease the quality gate if 50% code coverage suffices for us.

PR #1103, opened 2026-07-08
Fix Quality.withComment discarding column expression for renamed/computed columns

mand35:
## Problem                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                    
`Quality.withComment(column, commentText)` — the overload reachable via the fluent `col(...).withComment(...)` syntax from `DsColComment` — throws away the `Column` it was given and re-resolves a plain `col(colName)`  against whatever DataFrame the surrounding `.select(...)` compiles against. This happens because it delegates to the two-string overload  instead of the three-argument overload that actually aliases the passed column.                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                    
Effect on any column built from a new expression (window aggregate, coalesce, alias not already present upstream):                                                                                                                                                                                                                    
  - if the alias name doesn't exist in the source DataFrame, Spark analysis throws `UNRESOLVED_COLUMN.WITH_SUGGESTION`                                                                                                                                                                                                                      
  - if the alias name collides with a different pre-existing column, the computed value is silently discarded and the wrong column is written instead — no error, no warning                                                                                                                                                                                                                                  

Plain passthrough columns (same name before and after) happened to work by coincidence.                                                                                                                                                                                                                                                   

## Fix                                                                                                                                                                                                                                                            

One-line change: the single-Column overload now calls `withComment(colName, column, commentText)` (three-arg overload) instead of `withComment(colName, commentText)` (two-string overload), so the original column expression is preserved through `.as(colName, metadata)`.

### DIFF
--- a/sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/Quality.scala
+++ b/sdl-core/src/main/scala/io/smartdatalake/util/spark/dataset/Quality.scala
@@ -22,6 +22,7 @@ package io.smartdatalake.util.spark.dataset
 import io.smartdatalake.util.LogUtils.{debLogFun, debugLog}
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.expressions.NamedExpression
+import org.apache.spark.sql.catalyst.expressions.objects.AssertNotNull
 import org.apache.spark.sql.expressions.Window
 import org.apache.spark.sql.functions._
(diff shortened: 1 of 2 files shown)


kaelte (approved):

zzeekk (approved):

kaelte:
@mand35 , do you want to sqash and merge?

Issue #940, opened 2025-01-17
Rename DeduplicateAction as UpsertAction


zzeekk:
**Is your feature request related to a problem? Please describe.**
The name DeduplicateAction is confusing as it is actionally not really deduplicating incoming data. In fact it is a SCDT1 (slowly changing dimension type 1) implementation. But SCD is also not a self explanatory name.
We should name it more according to what it does, i.e. MergeAction?

**Describe the solution you'd like**
Rename DeduplicateAction as MergeAction, and create a type alias DeduplicateAction which is marked as deprecated in schema viewer.

zzeekk:
A better name is UpsertAction; merge is an SQL statement. Upsert is what is done (update or insert).

zzeekk:
Implemented on branch `feature/940-rename-deduplicate-as-upsert`.

Renamed to `UpsertAction` rather than `MergeAction`: "merge" describes the SQL statement, which `HistorizeAction` and `CopyAction` with `saveMode = Merge` use as well, while "upsert" says what happens to the data. The docs now name the pattern explicitly - `UpsertAction` is SCD1, `HistorizeAction` is SCD2.

A Scala type alias does not work for the deprecated name: both `ConfigParser.className` and the json schema exporter resolve types by reflection over class files, and a type alias emits none. So `DeduplicateAction` stays a real case class, marked `@Deprecated`, sharing all logic with `UpsertAction` through a new `UpsertActionImpl` base class. Old configurations keep working and the type stays listed in the schema viewer.

Class level deprecation was not exported to the json schema so far, only attribute level. `GenericTypeDef` now carries `isDeprecated`, read from the Java `@Deprecated` annotation and emitted as `"deprecated": true` on the definition.

PR #1135, opened 2026-09-06
Rename DeduplicateAction as UpsertAction (#940)

zzeekk:
see #940


Geheiner (approved):
Thanks for the better Action Name, future users will no longer be confused by DeduplicateAction :)

PR #1144, opened 2026-09-13
Switch to the actively maintained Jackcess fork

spannm:
`com.healthmarketscience.jackcess` is maintained at a much slower pace than [`io.github.spannm/jackcess`](https://github.com/spannm/jackcess), a fork released regularly on Maven Central (currently 5.1.7, up from 4.0.5 here). The fork also has zero required runtime dependencies, unlike `jackcess:4.0.5`, which pulls in `commons-lang3` and `commons-logging` directly.

`AccessTableDataObject`'s usage (`DatabaseBuilder`, `Table`, `Column`) is unchanged, so this is a coordinate swap plus an import reorder and a stale comment fix -the `ignoredUnusedDeclaredDependency` entry claimed jackcess was "loaded via reflection". It's actually used directly, just missed by the bytecode analyzer like the other Scala call sites listed there.

Verified that `mvn compile -pl sdl-spark -am` still succeeds.


zzeekk:
Hi @spannm, thanks for the hint. As reading access data is not a key feature of our framework, we ok with the current situation. We might consider you're fork in the future.

spannm:
Hi @zzeekk,
thanks for taking a look.
`jackcess:4.0.5` pulls in extra dependencies, while the modern fork has zero and is CVE-free.

